Issue #187, opened 2014-12-04
Modal : open() and not close on overlay


baurez:
Hi,

Very nice framework !!
But I have two question : 
Could we open a modal without having a trigger in a button, directly in javascript ? 
like : $('#modal_ID').openModal()

Could we chose if a click in overlay will close or not the modal ?

:)

Thx !!

davilocho:
hello
I share a few lines that provide solution to anadi para Various modal windows open and close properly
```javascript

$.fn.extend({
    closeModal: function(options) {
      var defaults = {
        out_duration: 250,
        complete: undefined
      }
      var options = $.extend(defaults, options);
        if($("div#lean-overlay").size()==1){
            $('.modal-close').off();
            $(document).off('keyup.leanModal');
        }

       $("div#lean-overlay:last").velocity( { opacity: 0}, {duration: options.out_duration, queue: false, ease: "easeOutQuart"});

      // Define Bottom Sheet animation
      if ($(this).hasClass('bottom-sheet')) {
        $(this).velocity({bottom: "-100%", opacity: 0}, {
          duration: options.out_duration,
          queue: false,
          ease: "easeOutCubic",
          // Handle modal ready callback
          complete: function() {
            
            $("div#lean-overlay:last").css({display:"none"});

            // Call complete callback
            if (typeof(options.complete) === "function") {
              options.complete();
            }
            
            $('div#lean-overlay:last').remove();
          }
        });
      }
      else {
        $(this).fadeOut(options.out_duration, function() {
          $(this).css({ top: 0});
          
          $("div#lean-overlay:last").css({display:"none"});

          // Call complete callback
          if (typeof(options.complete) === "function") {
            options.complete();
          }
          $('div#lean-overlay:last').remove();
        });
      }

    }
  })

PR #245, opened 2014-12-13
Open modal without trigger of a button

afifsohaili:
This completes Dogfalo/materialize#187

I extracted the process of opening a new modal to a new function called `openModal()`
Now developers can do something like:

``` javascript
$(document).ready(function() {
  $("#modalToOpen").openModal();
});
```

to open the modal within Javascript.

Binding to click event is still as is.

Updated the documentation as well.


andresams:
Amazing, exactly what I was looking for

PR #286, opened 2014-12-17
example rewrite, custom namespace, AMD/CommonJS support

m3nt0r:
Here’s my „forms.js“ draft, which is AMD/CommonJS compatible and uses a custom namespace that could be build upon. Works with the "components.html" demo and could replace the current forms.js without change. But this is a mere proof of concept. I recommend further abstraction. What you see here would usually be the body of the main dist-file, as the environment preamble and intialization at the bottom only needs to take place once. 

I've prototyped two _main_ sub-namespaces called "plugins" and "behaviours". Plugins are essentially functions that are then wrapped into actual jQuery functions for a single element/selection, while Behaviours are only executed once and add events for the entire document to a _certain type_ of element. 

The third sub-namespace is called "selectors". I am not sure if that's a keeper, but it certainly helped getting a better picture.

The file should be separated further into smaller modular files like "select.js", "autoresize.js", "range.js", etc... so that in the end we have one file per component. Those files could then be bundled with something like [browserify](http://browserify.org/) and would become easily extendable and testable.

The initialization code at the bottom is not something i'd usually put in there. Fully fleshed out i'd throw all that stuff either inside the modules or create a single "start" (and "stop") method that could be called on .ready() and "as needed". 

`console.log(Materalize);`
object

``` js
  // new jquery methods
  $().materialAutoResize();
  $().materialRangeInput();
  $().materialSelect();
```

_PS: i've added some comments regarding the range input behavior for consideration_

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -1,86 +1,72 @@
-(function ($) {
-
-  // Text based inputs
-  var text_inputs = $('input[type=text], input[type=password], input[type=email], textarea');  
+(function (factory) {
+    if (typeof define === 'function' && define.amd) {
+        // AMD. Register as an anonymous module.
+        define('materialize-forms', ['jquery'], factory);
+    } else {
+        // Browser globals
+        factory(jQuery);
+    }
+}(function ($) {
+  var Materialize, scope = this;
+  scope.Materialize = Materialize = (scope.Materialize || {});
   
-  text_inputs.each(function(){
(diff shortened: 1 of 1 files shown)


smileytechguy:
I agree with these concepts, and think they should definitely be worked towards implementation.

Dogfalo:
We've added commonJS and amd support in 1.0.0 alpha

PR #294, opened 2014-12-18
More W3C Valid

coltonfitzgerald:
Fixed the Jade files instead of outputted HTML


Dogfalo:
It did not work, the branch may have been deleted
```
fatal: Couldn't find remote ref patch-4
Unexpected end of command stream
```

PR #369, opened 2014-12-31
Change the <select> callback to support jQuery (or any third party lib) onchange events

zegerhoogeboom:
Currently there is no way to attach event handlers to select elements because of the way you dynamically create your own structure to display the select. 

This callback adds this functionality through an alternative way, you can do:
$('#mySelect').material_select(myCallBackFunction);

Ofcourse, $('select').material_select(someFunction) wouldn't make sense since you would have a single function handling every select's change event, so you should just use this on a single element.

Accepting this pull request will not break any current code. Not passing in a callback simply doesn't fire it.


zegerhoogeboom:
Please see c775467

acburst:
Thanks for the help!

zegerhoogeboom:
Glad i could be of help!

PR #385, opened 2015-01-02
Template Footer Test Color Update

ConsoleTVs:


Dogfalo:
The problem was the `lighten` class

PR #405, opened 2015-01-06
composer.json

RaSeR:


Dogfalo:
Can you explain a bit on what this is for?

PR #416, opened 2015-01-07
Update tooltip.js

jesperbjerke:
Enhanced so that .tooltip(); can be called multiple times without breaking. It now clears variables, removes junk html and unbinds upon calling it.


Dogfalo:
Thanks for this!

jesperbjerke:
No problem, although it seems like the fix isnt in the compiled materialize.js files in /dist för v.0.95.0 ?

Issue #443, opened 2015-01-12
Select's dropdown misplaced when using in a modal or fixed parent


255kb:
The select's dropdown which is created when calling `material_select()` is placed relatively to the document and not the `<select>` element, when the select is in a modal or a fixed parent. 

See in these screen captures: 

Before clicking on the Select:
![image](https://cloud.githubusercontent.com/assets/7489814/5703968/464b1f84-9a6d-11e4-876c-5f317bc7cf86.png)
After clicking on the select: 
![image](https://cloud.githubusercontent.com/assets/7489814/5703985/6c167f74-9a6d-11e4-96bd-485afc4ceb0a.png)

I also tested with a simple `<div>` containing a select and placed with a fixed position (for example at `left:50px and top:50px`), and the select's dropdown is placed at 0,0 relatively to the document which is also not correct and should be 50,50.

The issue seems to be related with the dropdown JS code, I think that the use of `position()` and `offset()` is inverted. 
So this part (starting at line 34): 

```javascript
if (elementOrParentIsFixed(origin[0])) {
    activates.css({
        display: 'block',
        position: 'fixed',
        height: 0,
        top: origin.position().top,
        left:origin.position().left
    });
}
else {
    activates.css({
        display: 'block',
        top: origin.offset().top,
        left: origin.offset().left,
        height: 0
    });
}
```
should be rewritten like this: 

```javascript
if (elementOrParentIsFixed(origin[0])) {
    activates.css({
        display: 'block',
        position: 'fixed',
        height: 0,
        top: origin.offset().top,
        left:origin.offset().left
    });
}
else {
    activates.css({
        display: 'block',
        top: origin.position().top,
        left: origin.position().left,
        height: 0
    });
}
```

I am going to create a pull request for this bug.
Thanks

acburst:
Fixed in 3c501184aed7b5d4c6a7e5bca380fce731cee9ff

PR #444, opened 2015-01-12
Correcting the dropdowns positioning bug

255kb:
Fixes issue #443 when dropdown placed in fixed parent.


acburst:
I looked at this and there was a smaller bug so I just pushed a commit that fixed it using this. Thanks for the code! 

Fixed in 3c501184aed7b5d4c6a7e5bca380fce731cee9ff

PR #451, opened 2015-01-13
Added a Toast dismiss click handler

zegerhoogeboom:
On click of any element with a 'data-dismiss' attribute (which does not need a value) the toast will dismiss.

Example : 
```javascript
toast('Well.. <a class="btn-flat white-text" data-dismiss href="#">Dismiss already!</a>', 50000);
```
In this example the toast will dismiss after 50 seconds, are when the button is clicked.

### DIFF
--- a/js/toasts.js
+++ b/js/toasts.js
@@ -13,9 +13,7 @@ function toast(message, displayLength, className, completeCallback) {
     container.append(newToast);
 
     container.on('click', function(e) {
-        if ($(e.target).data('dismiss') !== 'undefined') {
-            dismiss();
-        }
+        if (typeof $(e.target).attr('data-dismiss') !== 'undefined') dismiss();
     });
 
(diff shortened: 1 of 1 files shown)


owencm:
Nit: line 16 has an if that doesn't use curley braces . It is preferable to write:

if (true) {
 something();
}

Even for single line expressions in a conditional since it makes editing in the future less likely to introduce a bug if someone wants to add to or modify the contained expression.

Otherwise this LGTM (although I don't have merge permission)

jadsonlourenco:
+1

ProfNandaa:
@zegerhoogeboom - The file on `master` has changed a lot ever since, might be good to revisit this again.

zegerhoogeboom:
As @ProfNandaa pointed out, this file has changed a lot ever since almost a year ago. Lets just consider this patch deprecated regardless of whether the functionality has since been added or not.

PR #454, opened 2015-01-13
Add .active to label elements if input has been prepopulated

afifsohaili:
If an input value has been prepopulated (as in `<input type="text" name="etc" value="Prepopulated">`), the label will overlap on top of the field value. Screenshot:

error

This fix checks if the input_selectors have value prepopulated on document ready, and adds class="active" to their labels accordingly.

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -24,6 +24,15 @@
       }
     });
 
+    // Add active if input element has been pre-populated on document ready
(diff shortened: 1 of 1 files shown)


Dogfalo:
Thanks for this, it actually used be in the file, but I removed it by accident.

PR #460, opened 2015-01-14
Parallax fix

jrvboesch:
There was a problem with the demo, that when you resize your window it get's a white padding on the top of the image.


acburst:
Thanks for the code!

I added it myself here, because I made a small tweak and you recompiled the javascript so you don't need to add it to the demo.

bc0d1be4b5b1df39e1c759af08a6e7869b28e2db

PR #461, opened 2015-01-14
Remove jQuery for document ready

luanmuniz:


Dogfalo:
Thanks for removing the jQuery dependency here!

PR #525, opened 2015-01-26
Integrate BrowserSync

drKraken:
Suggest using a browserSync instead of connect. Its more flexible, faster and reusable.
https://github.com/shakyShane/browser-sync

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -51,6 +51,18 @@ module.exports = function(grunt) {
         }
       }
     },
+     
+  // Browser Sync integration
+    browserSync: {
+      bsFiles: {
+          src : '**/**/*.*' // watch in all directories or in special one
+      },
+      options: {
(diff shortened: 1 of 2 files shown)


Dogfalo:
Tried it out, seemed much slower than connect. I may have set it up wrong but it took around 6 seconds for the server to start

drKraken:
It starts a little bit slower but it contain live reloading that work much faster for me than other one and allow to test page in different browsers easily. See it in action https://www.youtube.com/watch?v=Rv5f_v5gqw4

TimonVS:
I agree with @drKraken. Live reloading (CSS injection instead of reloading the whole page) across browsers is a huge plus.

Dogfalo:
Updated to browser sync 2.5. Seems faster now and the UI is nice! Made a few changes, but thanks for the PR!

PR #535, opened 2015-01-27
Mixin keyframe fix

complementstudio:
This fixes the mixin @keyframes issue and allowing grunt to
successfully build this when integrated with a grunt file.


Dogfalo:
Interesting, it was working for us. But thanks for this fix!

PR #549, opened 2015-01-29
Make page-footer use $primary-color

tscolari:
It's currently hard-coded to color("materialize-red", "lighten-2")

### DIFF
--- a/sass/components/_global.scss
+++ b/sass/components/_global.scss
@@ -213,7 +213,7 @@ img.responsive-img {
 footer.page-footer {
   margin-top: 20px;
   padding-top: 20px;
-  background-color: color("materialize-red", "lighten-2");
+  background-color: $primary-color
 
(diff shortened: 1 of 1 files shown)


snow01:
I would say rather than again hard-coding it to use $primary-color - it's better to define another variable for footer color, with a fallback to $primary-color

tscolari:
@snow01 I agree.
I believe the best would probably do like twt-bootstrap does, have variables for all the customizable aspects, not only for colors.

snow01:
True. I suggest if you can re-submit your pull request after changing footer color as I suggested above. I will also submit pull request for a few customization I did. I think slowly and steadily we can achieve a higher customization.

acburst:
Added in 45c7118a25a3cd43a59df88b714154e9fd534284

acburst:
Thanks!

PR #557, opened 2015-01-29
Slider lazy loading of images fix

smotru:
I don't know yet how to build or what are your rules. But I would like to help on future.


Dogfalo:
Hey thanks for pointing out the problem. We actually changed the slider a bit ourselves so we decided not to merge this in. Let us know if you are still interested in helping out with development!

PR #560, opened 2015-01-29
Option to display dropdown below origin

Olivier-Daire:
#533

Usage :
```javascript
$(".dropdown-button").dropdown({
      inDuration: 300,
      outDuration: 225,
      constrain_width: false, // Does not change width of dropdown to that of the activator
      hover: false, // Activate on click
      alignment: "left", // Aligns dropdown to left or right edge (works with constrain_width)
      gutter: 0, // Spacing from edge
      belowOrigin: true,
    }
``` 
or
```html
 data-beloworigin="true"
``` 

Hope this helps !


Dogfalo:
Good stuff, Thanks!

PR #561, opened 2015-01-29
sass: $sidenav-width to _variables and dynamic sideNav width in js

ttemiel:
Shouldn't the variable $sidenav-width be defined in _variables.scss, and preferably as !default?
And: dynamically getting the sideNav width in the JS.


ttemiel:
Please disregard/close this pull request, and look at #569 and #570 instead.

acburst:

PR #567, opened 2015-01-30
Bug fix expandable functionality in collapsible.

snow01:
When mode is expandable click event can directly pass to the open collapsible, rather than entire collapsible. If we do not do that way - for even number of collapsible expandable does not work. Change is small - changed from $this to $(this)

### DIFF
--- a/collapsible.html
+++ b/collapsible.html
@@ -109,7 +109,7 @@
         <p class="caption">
           Collapsibles are accordion elements that expand when clicked on. They allow you to hide content that is not immediately relevant to the user.
         </p>
-        <ul class="collapsible collapsible-accordion" data-collapsible="expandable">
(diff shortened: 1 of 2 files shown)


acburst:
I'm not sure what bug youre fixing. Youre saying that even number of collapsibles don't work? I've tried that and it seems to work for me.

snow01:
Did you try for "expandable" option or "accordion" ? - Please note issue is for "expandable"

I just committed a small test-change in collapsible.html --- you can see preview here:

http://htmlpreview.github.io/?https://github.com/snow01/materialize/blob/master/collapsible.html

Try to expand the expandable there - last one on the page.

Dogfalo:
Looks good. Thanks!

PR #569, opened 2015-01-30
move sidenav-width to variables.scss

ttemiel:
Shouldn't the variable $sidenav-width be defined in _variables.scss, and preferably as !default?


acburst:
We've decided to go with just a jQuery plugin option

PR #570, opened 2015-01-30
JS: make menuWidth an option of sideNav

ttemiel:
Make the menuWidth an option of the initializer of sideNav as suggested here: http://materializecss.com/mobile.html

    $('.button-collapse').sideNav({menuWidth: 240, activationWidth: 70});

Might not be ideal, because two values need to be changed (the sass variable, and the js initializer), but seems a sensible choice.


acburst:

PR #571, opened 2015-01-30
fix : initial dropdown height change when animation is in progress

eskan:
if a dropdown is updated when animation is in progress the dropdownRealHeight var is set to "current height" as dropdownRealHeight was defined in the update function.
This patch set the dropdownRealHeight as object property instead.


acburst:
Thanks! we've been meaning to address that bug

PR #572, opened 2015-01-30
(fix) Disabled buttons should have `cursor: default`

douggr:
Actually `cursor: pointer` means *clickable* elements.

This PR adds `cursor: pointer` to disabled buttons.


Dogfalo:
Looks good thanks!

PR #580, opened 2015-02-01
fix | picker__table  keep a proper distance

Aaaaaashu:


Dogfalo:
Can you explain what this is fixing?

Issue #548, opened 2015-01-28
Materialize Select is not focusable with key tab


kevinliteon:
Tested on this page http://materializecss.com/forms.html

robinbortlik:
:+1:

zen6:
:+1:

acburst:
The following elements support the tabindex attribute: A, AREA, BUTTON, INPUT, OBJECT, SELECT, and TEXTAREA.

I'm not sure how we would solve this issue atm, sorry.

robinbortlik:
Can you please reopen this task. I'm working on this functionality and it is almost done. I will create a pull request in few days.
Thank you.

robinbortlik:
Here is promised pull request https://github.com/Dogfalo/materialize/pull/582

PR #582, opened 2015-02-01
Feature #548 - allow select component to be controlled by keyboard

robinbortlik:
- refactored dropdown component
- add open and close listeners to dropdown component
- refactored select component (replace span by input)
- add keypress binding and logic to select


Dogfalo:
I took a look at this, it looks very good. Just wondering why the animation style has been changed for the select dropdown?

robinbortlik:
Ohh, really? I did not noticed any change in animation. Let me know, if it will need some fixes.

robinbortlik:
Or do you mean the css style changes? That I had to replace arrow with icon? It is because, it is not possible to use :after selector with input.

Dogfalo:
Take a look at http://materializecss.com/forms.html and click on the select dropdown. Then if you click the one after the change it has a different effect.

robinbortlik:
Yes, you are right. I think it is issue with the scrollTo function. hhttps://github.com/Dogfalo/materialize/pull/582/files#diff-338f03716a63d52eee0f79b4510d617cR245

When you open the select, then the dropdown should be on the position of selected option. 
Maybe we can postpone this behaviour with setTimeout. So it will "maybe" stop break the initial animation. I will look on it on tomorrow evening.

Dogfalo:
Great, looking forward to seeing it. Thanks for taking the time to add this
functionality for us!

On Mon, Feb 2, 2015 at 4:15 PM, Robin Bortlík <notifications@github.com>
wrote:

> Yes, you are right. I think it is issue with the scrollTo function.
> hhttps://
> github.com/Dogfalo/materialize/pull/582/files#diff-338f03716a63d52eee0f79b4510d617cR245
>
> When you open the select, then the dropdown should be on the position of
> selected option.
> Maybe we can postpone this behaviour with setTimeout. So it will "maybe"
> stop break the initial animation. I will look on it on tomorrow evening.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Dogfalo/materialize/pull/582#issuecomment-72539389>.
>



-- 
Doggy sends his greetings from Mars.

robinbortlik:
Ok, fixed. Can you test it now? :)

Dogfalo:
Looks like it is working now. I merged it in. Thanks again for your help :)

NewBo1234:
Please....what was the fix?  This is something I desperately need.  Where can I look at the code?

robinbortlik:
Hi @NewBo1234 , 

this is a commit which added a functionality: https://github.com/Dogfalo/materialize/pull/582/commits/ec69063ac60bda0050f88e7c2ed9aea97c56988d
Can you access the page?

Regarding to your second issue, I'm afraid, that you will have to investigate by yourself. I remember that times when I was working on this library, I had to pull it locally, update a code as I needed and then compile it for myself.

PR #597, opened 2015-02-03
Fix bug when use arrow up and down on select

robinbortlik:
- removed setTimeout from activateOption function
- caused problems when user wanted to navigate dropdown by up and down arrows


robinbortlik:
Hi @Dogfalo , I noticed that my yesterday hotfix caused even more problems. So I decided to revert it back. I understand that now, we will have again a small problem with animation but I think it is not big issue like this one. The users was not able to properly navigate by up and down keys through the select options. Which was really annoying.
Can you please merge this "another fix", and I will try to find other solution for this animation issue. 
Thank you.

Dogfalo:
What is the problem with keyboard navigation? I tried using up and down arrows and it worked.

robinbortlik:
I does not work in long list. Maybe we could modify example to use in select box at least 15 items. What do you think?

PR #612, opened 2015-02-04
Fix textarea autoresize issue (undefined variable)

thomasruiz:
Hey!

Since defining a variable in a block make it only available in this block, the variable `hiddenDiv` was not found and the script failed. These simple changes fix it.


acburst:
Not sure where youre seeing this, there is no reference to the variable outside the block

thomasruiz:
Why am I getting this error then? `materialize.js:2258 Uncaught TypeError: Cannot read property 'html' of undefined`

Line 2258 is `hiddenDiv.html(content + '<br>');`

gsilis:
I saw this issue as well. It didn't occur when I was using the non-minified JS file in the dist directory.

thomasruiz:
I don't understand why this PR is not merged. If you are not seeing this bug you should at least know that the variable `hiddenDiv` is defined in a block and used in the parent one. I don't even know why it is working on the non-minified version.

PR #621, opened 2015-02-05
fix events are removed in collapsible

eskan:
I'm not sure why those lines was there but they disallow events inside a collapsible.

### DIFF
--- a/js/collapsible.js
+++ b/js/collapsible.js
@@ -15,10 +15,6 @@
 
       var collapsible_type = $this.data("collapsible");
 
-      // Turn off any existing event handlers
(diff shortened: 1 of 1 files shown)


Dogfalo:
If you don't understand why they are there, Why are you changing it?

eskan:
:D ... sorry for the bad explanation. In fact these lines did not make sense to me and I did not understand their uses in the context.
if you want to put a button in this component, the events attached to it are deleted with these lines

Dogfalo:
Fixed in 11d4a188039ea4f376e7ef986bdf00609eefcafd

PR #623, opened 2015-02-05
fix(_buttons): parent selector on a base-level rule

urecio:
Removed useless parent selector. Also was crashing with compass


Dogfalo:
Thanks!

PR #636, opened 2015-02-07
Update Footer w/ updateyourfooter.com snippet

coltonfitzgerald:
The Materialize site still says 2014, with this javascript snippet it will always display the current year :+1:


Dogfalo:
We don't really want to add a script for this since its just a yearly thing. Thanks though!

PR #639, opened 2015-02-07
Adding viewport meta header to getting started tpl

licryle:
Adding viewport meta header to getting started tpl in order to enable proper mobile support, which materialize does nicely


Dogfalo:
Good idea, thanks!

PR #643, opened 2015-02-08
Fix SideNav button toggling behavior

licryle:
Old behavior being fixed:
1- SideNav is not out
2- Press button to slide out
3- Press button again
=> the pane is going out with a second [third, fourth...] overlay (and with same id)

You can also call "show" several time creating the same behavior.

Note: the code looks for whether "menu_id" has CSS class "active", but class is never set nor removed.

New behavior post fix:
- SideNav is not out
- Press button to slide out
- Press button again
  => the pane is removed and overlay is removed. The button is effectively toggling.

The fix offers to add the CSS class "active" to "menu_id" when the show action is acted, and to remove the class in "removeMenu".

The fix also modifies "show" and "hide" only when the menu is in the desirable state.

Note: you probably need to regenerate your complete file + mimified versionif you accept this change

### DIFF
--- a/js/sideNav.js
+++ b/js/sideNav.js
@@ -109,6 +109,7 @@
           }
 
           // enable_scroll();
+          menu_id.removeClass('active');
         }
 
 
@@ -227,6 +228,7 @@
             }
             else {
               // disable_scroll();
+              menu_id.addClass('active');
 
               if (options.edge === 'left') {
                 $('.drag-target').css({width: '50%', right: 0, left: ''});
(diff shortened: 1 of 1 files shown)


plrthink:
+1 for @licryle

kmmbvnr:
I can't figure out how to do `3- Press button again` The button is under the overlay, and clicking on the button, hides the sidenav.

But two subsequent call for `show`, still creates an additional overlay - http://codepen.io/kmmbvnr/pen/qNLEXg

kmmbvnr:
Looks fixed in 1d4bbe3a56815dcf1ef66162e342c5e4f09cc0e2 by @acburst 

Here is codepen with github master js: http://codepen.io/kmmbvnr/pen/ZOVYao

smileytechguy:
This seems to be fixed.  Using your codepen, if I run `$('.button-collapse').sideNav('show');` twice, however with a 0.1s delay between, it works as expected (rehides, overlay is removed). The issue is how the sideNav function works, as it does its action asynchronously, allowing for 2 calls simultaneously to occur.  This PR should be closed and a separate one made to fix this issue.

PR #658, opened 2015-02-10
added toptoast

IagoLast:
Just another toast class, I think it can be useful.
Also added docs about how to use it. (small changes)

Looks like:
toptoast


acburst:
Hmm Im not sure how you made this but the toasts don't stack like ours do. If you can fix that and update to latest master so its mergeable and we can take another look

carlosperate:
This actually brings an interesting topic, based on the material design guidelines the toasts should be placed in the bottom or bottom left areas http://www.google.co.uk/design/spec/components/snackbars-toasts.html#

> Snackbars provide lightweight feedback about an operation. They show a brief message at the bottom of the screen on mobile and lower left on desktop. Snackbars appear above all other elements on screen. Snackbars can contain an action.

Are you guys going to allow toast to be placed anywhere? (also I believe current desktop toasts appear at the opposite corner where they are meant to).

IagoLast:
@acburst since the toptoast uses display:fixed it cannot be stacked.
A posible solution would be creating a top-toast container and put the toptoast there, this implies some small javascript changes.
Another option is to change the toast container position, fix it to the top and add some styles to the toast class to keep the same apearance. (im gonna try this one)

IagoLast:
ok, fixed by creating a new container.

Dogfalo:
@carlosperate Material Design is made primarily for mobile, so sometimes it might not be well translated to the web. I think more people are accustomed to notifications on the top right for desktop.

Then for tablet/mobile we stay inline with the specifications

carlosperate:
That is fair enough, although I would argue that since Windows has over 80% of desktop presence most people are accustomed to notifications coming from the bottom right corner.

gdoron:
@Dogfalo  It makes sense, but it would be awesome if you could send an (optional) parameter corner parameter (top-left, top-right, bottom-left, bottom-right).

gdoron:
In case someone is interested in a simple and flexible solution for putting the toast on the bottom:

	var duration = 3000;
	Materialize.toast('Subscribed!', duration, 'rounded')
	$('#toast-container').addClass('bottom')
	this.reset();
	setTimeout(() => $('#toast-container').removeClass('bottom'), duration);
	
And the SCSS:

	@media #{$large-and-up} {
	    #toast-container.bottom {
	        bottom: 10%;
	        top: unset;
	    }
	}
	
It's nice because you can control when to show on the bottom.

PR #667, opened 2015-02-11
Fix broken wave effect on disabled <button/>

danilorossi:
In a project of mine, I had to use a button element instead of an anchor in order to have the disabled attribute fully working (e.g. avoid a click to be fired) with no additional code.

I found out that disabled buttons don't fire events, so, when the time comes to remove the wave effect, the mouseup/mouseleave event inside the showEffect function are not fired, and no Effect.hide is called.

Basically, lot of Effect.show are fired, but no Effect.hide are, so what you get is a lot of overlapping layers over the button (many as the number of clicks on the button) that are never removed.

The fix is basically a test to avoid the Effect.show to be fired if the element is a button with a disabled attribute set (is somehow consistent to avoid a UI feedback on a disabled button).

There will be a better solution, but at least you are aware of the issue.

### DIFF
--- a/js/waves.js
+++ b/js/waves.js
@@ -270,12 +270,20 @@
     }
 
     /**
+     * Test if wave effect is applicable on given element
+     */
+    function canWave(element) {
+        return element !== null && 
+            !(element.tagName === 'BUTTON' && element.hasAttribute('disabled'));
+    }
+    
+    /**
      * Bubble the click and show effect if .waves-effect elem was found
      */
     function showEffect(e) {
         var element = getWavesEffectElement(e);
 
-        if (element !== null) {
+        if (canWave(element)) {
             Effect.show(e, element);
 
             if ('ontouchstart' in window) {


Dogfalo:
I am not seeing the problem you are having
`<button disabled class="btn waves-effect">Test</button>`
This works as intended for me.

danilorossi:
Hi Alvin,
Tomorrow I will send you the piece of HTML affected by the problem. Does
that HTML fire the mouse events used to remove the wave effect?

Dogfalo:
Disabled buttons are not clickable, so there is no wave effect that appears.

danilorossi:
In my case every time I click I can see several wave show effects going on,
buy no wave hide is called.. Tomorrow I will send you my piece of HTML, the
disabled attribute is angular driven.. Maybe it's a specific issue of my
environment.

danilorossi:
Hi Alvin,
here's my HTML:

<button ng-click="someFunction()" ng-class="{'disabled': someExpression }"
ng-disabled="someExpression" class="waves-effect waves-light btn">Some
Text</button>

Without my tweak, the disabled button got clearer and clearer as I keep
clicking on it, and the waves are never removed.

Cheers,
Danilo

2015-02-11 19:56 GMT+00:00 Danilo Rossi <contact@danilorossi.me>:

> In my case every time I click I can see several wave show effects going
> on, buy no wave hide is called.. Tomorrow I will send you my piece of HTML,
> the disabled attribute is angular driven.. Maybe it's a specific issue of
> my environment.

enmanuelr:
Similar problem over here. When a button with `waves-effect` class is disabled, it creates a bunch of ripple children whenever clicked, but are never removed. Please reconsidering taking another look at this.

screen shot 2016-07-03 at 11 45 51 pm

tomscholz:
I am not able to reproduce this issue. Can someone send a codepen to illustrate the issue?

fega:
@Dogfalo @tomscholz @acburst This should be closed.

PR #676, opened 2015-02-12
Allow the user to fling the sidenav away

owencm:
The ability to fling the sidenav away regressed. This fixes it again.

To test:
1) open any materialize page with a sidenav on a touch enabled device
2) attempt to quickly 'swipe' the sidenav away

Expected behaviour: the sidenav swipes away with your momentum
Actual behaviour: the sidenav bounces back

This change causes the expected behaviour to work, which seems to have regressed.

### DIFF
--- a/js/sideNav.js
+++ b/js/sideNav.js
@@ -126,7 +126,7 @@
         var menuOut = false;
 
         $('.drag-target').hammer({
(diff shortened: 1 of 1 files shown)


Dogfalo:
Thanks for fixing it!

PR #678, opened 2015-02-12
null src in img tag

leVirve:
empty 'src' attribute will cause an error and display a broken icon on  element in Firefox (but in Chrome, it works just fine)
so replace empty src with an base64 transparent 1x1 pixel GIF image


Dogfalo:
Great fix, thanks!

PR #679, opened 2015-02-12
Duplicate row, remove

gitarno:


acburst:
I fixed this problem in the jade templates. Thanks for pointing this out!

PR #694, opened 2015-02-16
feat(sidenav): Allows percentage width + example

urecio:
chore(materialize.js): generates bin file


Dogfalo:
Useful option, Thanks!

Dogfalo:
Actually upon further testing, I realized that supporting percentage widths is much more complicated than it seems. There are problems with resizing. I've reverted the change for now.

Issue #696, opened 2015-02-16
.modal-close click listener never removed onClose


MarijnMensinga:
When a modal is created this click listener is added:
```javascript
$(modal).find(".modal-close").click(function(e) {
   e.preventDefault();
  $(modal).closeModal(options);
});
```
But it is never removed in the `closeModal`. So upon opening and closing a new modal via the `.close-modal` class the `closeModal` function will be called twice,.. after that three times etc...

PR #697, opened 2015-02-16
Fix for closeModal recursive click event

MarijnMensinga:
More details: https://github.com/Dogfalo/materialize/issues/696


Dogfalo:
Thanks for catching this bug!

Issue #690, opened 2015-02-15
Side menu width doesn't work


ventouris:
On documentation ```$('.button-collapse').sideNav({menuWidth: 240, activationWidth: 70});```

Even if I change the menuWidth to another value, it stays on 240px

yowkah:
Works for me although i have some CSS added.
for me the JS is:
```js

 $(function(){
 	$(".button-collapse").sideNav({menuWidth: 300});
 });

```

Can you place some excerpts of your UL's HTML and the html of your link that activates the sidemenu.
Also ( even though i assume you did) can you check if you put your javascript in a document ready function from jquery and if it is loaded in AFTER all the javascript from materialize?

ventouris:
```
<ul id="slide-out" class="side-nav full">
    <li class="no-padding">
      <ul class="collapsible collapsible-accordion">
        <li>
          <div class="collapsible-header"><span id="manager">{{headerstr["profile"]|safe}}</span><i class="mdi-navigation-arrow-drop-down"></i></div>
          <div class="collapsible-body">
            <ul>
              <li><a href="/">Home</a></li>
              <li id="login"><a href="/login">Login</a></li>
              <li><a href="#" onclick="logout()">Logout</a></li>
            </ul>
          </div>
        </li>
      </ul>
    </li> 
    .......(the same <li class="no-padding"> code for a few times)
</ul>

<div style="width:100%;top:0;left:0;background-color:black;height:45px;" class="valign-wrapper">
  <a href="#" data-activates="slide-out" class="button-collapse" style="color: white;"><i class="mdi-navigation-menu left">  Menu</i></a>
</div>```


```<script type="text/javascript">
    $(document).ready(function(){
      $('.button-collapse').sideNav({menuWidth: 340, activationWidth: 70, edge: 'right'});
    });
</script>```

But on CSS in Firebug:

```ul.side-nav.full, ul.side-nav.fixed {
    background-color: #fff;
    height: calc(100% + 60px);
    left: -250px;
    margin: 0;
    padding-bottom: 60px;
    position: fixed;
    top: 0;
    width: 240px;
    z-index: 999;
}
```

yowkah:
Okey i just can't seem to replicate the bug. When I use your code ( html converted to jade, but still), I can't recreate the problem within my own environment. Does the collapsing of the menu itself actually work? does the menu get placed on the right side?

for reference: my code.

```jade
html
	head
		meta(charset='utf-8')
[...]

PR #699, opened 2015-02-16
fix right edge with full navigation issue

yowkah:
I want to implement a fix for issue #690


Dogfalo:
We recently fixed many bugs in sidenav, this should have been fixed as well. Let me know if it was not.

Issue #681, opened 2015-02-13
Clean console.log from code


rma4ok:
[Code where it left](https://github.com/Dogfalo/materialize/search?utf8=%E2%9C%93&q=console.log)

tjoskar:
:+1:

venil7:
this issue gets my vote, it's ridiculous to see `console.log` in minified build

Dogfalo:
Gone, we we have a grunt script that removes them from any distributed release

PR #700, opened 2015-02-16
Improve/console log cleanup

xamenrax:
Solve #681


Dogfalo:
We're now deleting console.logs through grunt before every release.

Issue #692, opened 2015-02-16
Tooltips remains visible after the element is removed from the DOM


Cerealkillerway:
The tooltips remains visible after the element they are associated to is removed from the DOM by ajax;
this happens for example in a single page application, when clicking on a button with a tooltip that triggers a new template to be rendered in place of the current one (ex changing route);
is there a way to avoid this behaviour without having to remove all the tooltips manually after every template render?

Cerealkillerway:
A possible solution would be also to have tooltips created with an id that references their "tooltipped DOM element"; in that way it would be possible to remove tooltips when the template is destroyed;

right now tooltips only have a "material-tooltip" class, so when a template is destroyed can only remove ALL tooltips (also the ones referencing elements in other templates that remain visible)

Cerealkillerway:
I've made a little variation of the materialize's tooltip function that supports a container argument (to append the tooltips to) and calculates position corrections if the container (or any ancestor) is positioned by CSS "position" rule:

https://gist.github.com/Cerealkillerway/0dece72eece0aec32878

maybe someone can find this useful

emd2009:
Do you have a jfiddle demoing this

Cerealkillerway:
You just need to add the file to your project, and then activate the tooltips:

   $('.tooltipped').ckTooltip({delay: 10, container: this.$('desiredContainerIdOrClass')});

In this way you can try if it works for you without editing the materialize's source

Roundaround:
@Cerealkillerway, I'd encourage you to implement your gist as a proper pull request (I see it has a different changeset than your previous solution), and perhaps use http://jsfiddle.net/ or http://plnkr.co/ to demonstrate your new addition.  That way people would be more willing to accept that your solution is a good one.  Right now, there are too many differences between your version and the head (running through a diff breaks all over in places it really shouldn't).  Resolve these and come back and we can work from there.

Cheers.

Cerealkillerway:
well, I've made that gist just for sharing that possibility, that I use in my project; it's not perfect and 360° tested but it's a starting point; This is why i've not made a pull request yet; for example: it corrects the position when the container where you append the tooltips is inside a css-positioned element; but if between that container and the inner element that will have the tooltip there are other css-positioned elements, I guess it will fail (it will need another correction); I will try to implement it as soon as I will have the time ;)

Roundaround:
So while looking for cleaner solutions to this issue, I came across this:
[...]

PR #704, opened 2015-02-17
Fix for #692

Cerealkillerway:
This fixes <a href="https://github.com/Dogfalo/materialize/issues/692">#692</a>

I suggest if possible to use the same behaviour for all components with shadow markup; appending to the parent conteainer instead of appending to 'body' allow MVC frameworks like Ember, Angular or Meteor to remove the shadow markup when the template is removed from the DOM

### DIFF
--- a/js/tooltip.js
+++ b/js/tooltip.js
@@ -24,7 +24,7 @@
       // Create tooltip
       var newTooltip = $('<div></div>');
       newTooltip.addClass('material-tooltip').append(tooltip_text);
-      newTooltip.appendTo($('body'));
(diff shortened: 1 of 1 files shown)


Dogfalo:
This will mess up the placement of the tooltip for fixed elements

Cerealkillerway:
I see that this happens if the parent container has fixed position, but not if the element itself is fixed;
probably a separate management if the container is fixed or using an id for the tooltips, like you do for the select boxes will make easier to remove the tooltips associated elements within the template that is going to be removed...

Dogfalo:
We'll work on this in later updates.

PR #712, opened 2015-02-18
Update _collapsible.scss

coreybutler:
Display the body of an active collapsible region.

### DIFF
--- a/sass/components/_collapsible.scss
+++ b/sass/components/_collapsible.scss
@@ -43,7 +43,7 @@
 }
 
 li.active .collapsible-body {
-
+  display: flex;
 }
 
 // sideNav collapsible styling


coreybutler:
This appeared to just be missing. Without this, the current code depends on the click handler to expand the collapsible body. 

Setting `<li class="active">` would not expand the body until it is clicked on twice (the first click handler detects the `active` class and removes it. The second detects that there is no `active` class and adds it, but it also adds the additional `display: flex;` CSS directly to the `.collapsible-body` element.

Dogfalo:
You can just add class="active" to the collapsible-header.

coreybutler:
@Dogfalo - No - setting this on the collapsible-_header_ has no effect on the collapsible-**body**. The effect when setting `class="active"` on the parent element of the collapsible-header also doesn't work.

Take the following code:

```html
    <ul class="collapsible" data-collapsible="expandable">
      <li class="active">
        <div class="collapsible-header">title</div>
        <ul class="collapsible-body">content</ul>
      </li>
      <li>
        <div class="collapsible-header">title 2</div>
        <ul class="collapsible-body">content</ul>
      </li>
      <li>
        <div class="collapsible-header">title 3</div>
        <ul class="collapsible-body">content</ul>
      </li>
    </ul>
```

I would expect this to render the accordion with the first region expanded. It does not, because the collapsible-body `display` is still hidden. Furthermore, if you click on the header (which is still collapsed), the jQuery removes the `active` class as  though the panel is being collapsed. You then have to click on it again and it expands by setting the `display:flex` on the collapsible-**body**.

The following code DOES work as expected:

```html
    <ul class="collapsible" data-collapsible="expandable">
      <li class="active"> <!-- ADDED CLASS -->
        <div class="collapsible-header">title</div>
        <ul class="collapsible-body" style="display: flex;">content</ul> <!-- ADDED STYLE TO FORCE DISPLAY -->
      </li>
      ...
    </ul>
```

If you look at the diff of the PR I submitted, you'll see there is just an empty block with no CSS attributes in the original file. It just needs `display: flex` to allow devs to just add `class="active"` to the `<li>`. By merging this, the following code works as expected:

```html
    <ul class="collapsible" data-collapsible="expandable">
      <li class="active"> <!-- ADDED CLASS -->
        <div class="collapsible-header">title</div>
        <ul class="collapsible-body">content</ul> <!-- NOTHING ELSE NEEDED! -->
      </li>
      ...
    </ul>
```

For reference, I am only using the SASS with the following imports:

```css
@import "components/prefixer"
@import "components/mixins"
@import "components/color"
@import "components/variables"
@import "components/global"
@import "components/typography"
@import "components/collapsible"
```

The JS code is as follows:

```js
    $(document).ready(function(){
      $('.collapsible').collapsible({
        accordion : false
      });
    });
```

I can reference this as a bug if you like.

Dogfalo:
There is no need to initialize the collapsible.
```HTML
  <ul class="collapsible" data-collapsible="accordion">
    <li>
      <div class="collapsible-header"><i class="mdi-image-filter-drama"></i>First</div>
      <div class="collapsible-body"><p>Lorem ipsum dolor sit amet.</p></div>
    </li>
    <li>
      <div class="collapsible-header active"><i class="mdi-maps-place"></i>Second</div>
      <div class="collapsible-body"><p>Lorem ipsum dolor sit amet.</p></div>
    </li>
    <li>
      <div class="collapsible-header"><i class="mdi-social-whatshot"></i>Third</div>
      <div class="collapsible-body"><p>Lorem ipsum dolor sit amet.</p></div>
    </li>
  </ul>
```

coreybutler:
I don't think the issue has been explained clearly. The goal is to get one or more of the options to expand by default so the user **doesn't have to click on anything**. The goal is to make the initial screen render as shown below:

image

The code you provided above does not do this. You can see a copy/paste of your code rendered below: 

**Notice the display body is `none`**
image

Using your code snippet, you can click on the "Second" header to make it expand like the first screenshot, but the goal is to do this by default so the user isn't forced to click on anything when the page initially loads.

The **original** code ([lines 45-47](https://github.com/Dogfalo/materialize/blob/master/sass/components/_collapsible.scss#L45-L47)) in `_collapsible.sass` responsible for displaying the body is empty, which is why `display` is defaulting to `none` (from the CSS declaration on [line 35](https://github.com/Dogfalo/materialize/blob/master/sass/components/_collapsible.scss#L35)):

image

My PR simply fills in the missing piece:

image

You can set a collapsible element to `active`, as the original CSS clearly intended, and it will auto-expand. Here's a screenshot with this PR applied:

**Notice the `li.active .collapsible-body` class containing** `display: flex;`

image

I hope that clears things up.

Dogfalo:
By adding active to collapsible-header, this makes the collapsible expand without the user having to do anything. The reason this is not working for you is because you initialized the component twice.

coreybutler:
Sorry to be dense... in reviewing the documentation again I see where you're coming from now.

PR #718, opened 2015-02-19
fixed card reveal issue

mendacitees:
Card reveal panel was not hiding after v0.95.2 changes


Dogfalo:
So this is actually a bit more complicated. This line you added was originally in the file. However, we made some changes to support the use of Materialbox in cards. But we may reverse those changes.

acburst:
We manually reverted this. Thanks anyways

PR #726, opened 2015-02-20
Fixed strict error.

yourivdlans:
This PR removes a duplicate position attribute which causes an error while using strict mode.


acburst:
Thanks!

PR #750, opened 2015-02-25
Fix textarea auto resize

### LINKED ISSUE #684: Height of textarea in modal is wrong
If I have a form with a textarea in a modal popup, the height gets wrongly calculated, because the original element is hidden from the DOM.

textarea-modal

Is there a way I can work around that problem?

SmileyChris:
Ensure that textareas of different widths (including those initially
hidden) are resized correctly.

Ensure textareas containing html within their content are escaped
correctly when calculating the width.

Fixes #684

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -69,8 +69,8 @@
 
 
     // Textarea Auto Resize
-    var hiddenDiv;
-    if ($('.hiddendiv').length === 0) {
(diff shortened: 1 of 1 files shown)


acburst:
Merged in e7d7d150de6b49a045b1d999b1bd16170f56d4aa

Thanks for the help!

PR #752, opened 2015-02-25
dropdown onClick bind fix

MarijnMensinga:
When dom data is loaded async and the dropdown functionality needs to be re-initiated via the same selector (e.g. `$('.dropdown-button').dropdown({})`)  previously bound buttons will fail to work. This fixes the problem.

But maybe a check for pre-existing binds is better?


Dogfalo:
It is fine to have them update all incase you update options.

Issue #764, opened 2015-02-26
Using 0.95.3, I get an error about scrollFire


chrisronline:
```
Uncaught ReferenceError: scrollFire is not defined
```

Using both downloaded 0.95.3 and off [the CDN](https://cdnjs.cloudflare.com/ajax/libs/materialize/0.95.3/js/materialize.js)

Looks like the grunt dependency order for the js files is wrong. Will fix in PR

chrisronline:
False alarm. Something wrong in my env

Dogfalo:
For future reference, what was the problem?

chrisronline:
I'm not too sure - I'm resurrecting a site that I made years ago and I had some lazy loading logic to load my scripts instead of simply including them on the html page. It was something with that because when I removed it and just put <script> tags, it worked fine. If I have time later today, I'll dig more into it and see if I can come up with a better explanation

PR #765, opened 2015-02-26
Transitions depend on scrollFire - load after

chrisronline:
Addresses #764


Dogfalo:
I don't see why transitions would depend on scrollfire. ScrollFire is independent of transitions, it can be used to fire any function. If anything, place your js in a document.ready

chrisronline:
Hmm - maybe something I'm doing wrong in my setup. I PRed because I fixed the issue and used the fixed ```materialize.js``` in my local development and the error went away. I'll get back to you about this. Thanks

chrisronline:
Yup something weird on my end. Disregard

PR #768, opened 2015-02-26
allow tab color override

pfafman:
Allow tab underline color to be over ridden.

```
$tabs-underline-color: $primary-color-light !default;
```


Dogfalo:
There are too many changes bundled in this PR, it should be a one line fix.

pfafman:
Yes I know.  Seems OS X on checkout is converting all the linefeeds.  

```
warning: CRLF will be replaced by LF in ...
```

Not sure why, if you know how to get it to ignore that let me know and yes it is a one line fix in 

```
sass/components/_form.scss
```

Tim

PR #777, opened 2015-02-27
Safari fix

pfafman:
Fix for Safari mobile which does a default border radius for text inputs.

1 line change.


pfafman:
Handled with resent commit.

PR #782, opened 2015-02-28
Destroy cancel modal keypress event

SmileyChris:
Ensure modal "escape key" event handler is destroyed no matter how modal is closed


Dogfalo:
Thanks!

PR #789, opened 2015-03-01
Updating index.html for starter-template

oznazilli:
(My apologies if I have any missteps, this is my first time proposing a change to a file but I really like your work with Materialize so wanted to help)

Currently, the index.html in the starter-template project is missing a class for the first <ul> in nav, which is leading to the nav with class="right" not being hidden on mobile. I propose adding hide-on-med-and-down as recommended in your guide to fix the issue that drove me crazy for a while. Namely, having two menus show up on mobile.

Thanks for the framework!


Dogfalo:
Thanks!

PR #794, opened 2015-03-01
fix AMD, permanent solution 

max1011:
the fix was to disable amd by setting define.amd to false and detect that when needed, because we can't define more than one module at the same file the rule says that there should be one module per file, so what happens is that all the plugins work as if they are not required explicitly with require and will be publicly (global var) available to use, and in the jquery hammer js file we detect if we already set that define.amd is false and define also exists, so that we know that amd support exists but we changed it to work the way we want to.
that line gets appended by the append_file grunt task which is for both temp and dist, it will run when js_compile runs and can be run manually if needed, so whenever there is a need to combine all the file into one files we add that line and detect it and do the necessary changes.

another bug was fixed when using the waves effect, it was fixed in the jquery hammer js file as well, the problem is that waves uses the on load event and that some times fires before require loads, so the event will be missed when used with require.js, so we initialize waves manually when require is actually ready.

**Note**: i didn't fully test this, i mean i didn't test all the different cases, rather i tested the usual usage of amd (require.js specifically) and materialize and seems to go well with no performance drawback at all.
if you have any question, i am here! don't hesitate to ask about any problem that you are having with the solution!.

**i would love to hear from anyone about this solution if it works please tell me !!**


timothynode:
i used your fork.But the button waves still not show  and pickadate is undefined

max1011:
Make sure to include the file from the bin directory not from any other

timothynode:
i did include materialize.js from bin directory. there was no error. but the button wave effect never show

max1011:
You might need to initialize it your self inside of the amd require research that function call i think it is something like this:
Waves.displayEffect()
But i am not sure just research a bit and you will know the right one

timothynode:
I tried add ```Waves.displayEffect()``` and it worked. thx~

frankleng:
@max1011 mind resolving the conflicts?

max1011:
Yes but the problem is whenever i resolve they keep updating it if ibresolve they should merge into master if not then there is no need to do it unless they want to merge and they are confident about it

patriciapineda:
This solution works fine.. should be merged into master!

frankleng:
@Dogfalo can haz a merge?

max1011:
for further updates follow me @max1011

dist/js/materialize.js
```diff
@@ -220,8 +217,15 @@ jQuery.extend( jQuery.easing,
...
-    if (typeof define === 'function' && define.amd) {
-        define(['jquery', 'hammerjs'], factory);
+    if (typeof define === 'function' && define.amd == false) {
+        // hammerjs will already be publicly available
+        define(['jquery'], function($, Hammer) {
```
dak0rn: `define()` is missing second dependency, isn't it?
Something like `define('jquery','hammer', function($, Hammer) {` otherwise `Hammer` will always be `undefined`.

max1011: well, what happens is i change define.amd to false so when it comes to hammer it will skip the if statement that checks for define.amd because it's false and it will be used as if it was public (normal), so hammer will be publicly available, and the argument Hammer that is passed to the function probably i forgot to remove it because we can use hammer normally without requiring it, so amd might allow us to do something like that without giving an error and works normally but i am not sure so i will check and i will let you know, when trying to log Hammer inside of the function it will probably give us the Hammer function but like i said i am not sure and i will test it, and if there is a need for change i will change!

dak0rn: I ran into that require.js error, too, so I edited `dist/materialize.js` and added only line 1 and 220-228 you added.
The reason for that is that I did not want to use a version shipped with materialize but a self-managed using bower.
After installing it and adding it to my require.js configuration everything works fine. However, I had to change line 222 to what I wrote in my comment.

So, even though Hammer might be available publicly I would suggest relying on the AMD dependency management to be more stable in custom environments. However, my idea might be totally wrong since I had not time to check how Hammer is integrated internally.

max1011: the problem here with this file is that they are using multiple define statements and that's not allowed so to over come that i did what i did (it's all explained) so the thing is here you can't rely on amd because you can't define a module then require it at some point later in that file so the only solution is to make hammer publicly availabe (from that file) so when hammer runs it's code it thinks it's being used normally but it's actually used with amd, i haven't tested what i told you i wanted to test, i am busy today so maybe tommorow or after it (i don't know when i will have free time) so yeah , just wait until i fully test and i will tell you the whole story ok ?

dak0rn: Interesting, although I use require.js every day I never ran into that issue.
However, I'm looking forward to seeing your solution improving materialize.
Thanks in advance!

max1011: and also by the way your suggestion is amazing but like i said it is not possible that's what the guys that created Require did and it's the AMD spec and i didn't wrote it so i don't know why they don't allow you to do such a thing, anyways i think now you know why i rely on hammer being publicly available vs. amd version of it, and note that you don't have to include hammerjs file in require config all you need is the materialize file because it makes no sense to have the same code twice which is the hammer library, this solution works and tested but what i haven't tested is what i talked about so there is no problem in using it i guess, only if you have any problems just include materialize from the bin directory and it will work fine (of course using require not the normal way)

dak0rn: Btw, the `Hammer` argument would overwrite the global `Hammer` variable, wouldn't it?
Thus, even though you are relying on the built-in Hammer implementation you will get an error in this function.

max1011: so, it's been a while since the last time i worked with this repo so i investigated a bit and just remembered what i did exactly, turns out this file is kind of outdated, because this file is compiled with frunt but i only ran the dist js compile once on the first stages of the solution, this section where talking about is the jquery.hammer.js so you can see what i really did and what worked and what i tested , i was compiling the temp grunt task rather than dest so the most updated version is in the bin directory, that file is not readable so you can take a look at the real code in js/jquery.hammer.js there you can see what i did exactly, if you have any problem just LMK ok ?

Dogfalo:
@max1011 Any chance you know how to make it work with the code we have now? We would be happy to merge it in.

max1011:
yes, i will pull fix conflicts and then re push , the changes i add are very simple so it shouldn't affect your code, do you want me to recompile this file using dest specific commands,  with all your updates and then my own, i will just tell you some other things you have to keep in mind and that's all it works and there is no problem merging after conflicts are resolved, at the moment i am busy i am working on a project so just wait a few days, i will give you a heads up when it's all ready!

max1011:
right now i am like half way to fully fix the conflicts, what's left is the gruntfile and the dist materialize js file, i will try to finish those soon because these conflicts are really annoying and hard to figure out as well  so it might take me a while so that i can understand and fix the conflicts properly not just throwing code to fix.
just be patient !, and hope it all works!

PR #795, opened 2015-03-01
Use dev dependencies

piuccio:
Move all `grunt-` from `dependencies` to `devDependencies` so applications depending on `materialize` don't have to load them as well.


Dogfalo:
Thanks!

PR #796, opened 2015-03-01
Some files are in CRLF

piuccio:
I've seen that some files are in `CRLF`, make it explicit so it's easier to contribute for people on linux/mac


Dogfalo:
Files are now LF

PR #814, opened 2015-03-03
Custom parent container for pushpin

rnzo:
Added ability to specify the parent container for the pushpin plugin (window by default)

### DIFF
--- a/js/pushpin.js
+++ b/js/pushpin.js
@@ -19,7 +19,8 @@
       var defaults = {
         top: 0,
         bottom: Infinity,
-        offset: 0
+        offset: 0,
+        parent: window
       }
       options = $.extend(defaults, options);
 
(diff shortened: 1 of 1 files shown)


Dogfalo:
I tested this and the parent option didn't seem to work, maybe you have a example of this you've created where it works?

sly7-7:
I wish this behavior too, I was trying to use pushpin, but can't make work, I've made this naive implementation in my code:

```js
let scrollingTreshold = this.$('.element-to-pin').position().top;
  this.$('.scrollable-container').on('scroll', (e) => {
    if(e.target.scrollTop >= scrollingTreshold) {
      this.$('.element-to-pinr').addClass('pinned');
    } else {
      this.$('.element-to-pin').removeClass('pinned');
    }
});
``` 

Of course, the `scrollable-container` should have the `overflow-y`property set to `auto` (or `scroll`) I guess.

I think it would be great if pushpin could support that 😄

PR #818, opened 2015-03-03
Add NuGet package option.

ziyasal:
Add NuGet package option to 3.Party options


Dogfalo:
I think we'd rather keep the third party list short and just have the popular options.

PR #830, opened 2015-03-04
Remove default margin-top in footer

tibastral:
This default margin-top in the footer is not working well. I personally prefer to handle vertical margins in other terms (than just a 20px). It should be at least configurable

### DIFF
--- a/sass/components/_global.scss
+++ b/sass/components/_global.scss
@@ -272,7 +272,6 @@ ul.staggered-list li {
 
 // Footer
 footer.page-footer {
-  margin-top: 20px;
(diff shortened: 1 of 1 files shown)


Dogfalo:
Why can't you just write CSS to override it? It is better than having all the content push down on it by default.

tibastral:
I totally can do it, but... By default, Il like margins to go away when I'm in 'mobile' mode

Dogfalo:
I think this would be up to the developer. Some may want it there, some may not.

tibastral:
And what do you think about a constant ?

PR #833, opened 2015-03-04
Add switchModal

luanmuniz:
I want to close a Modal and open a new one. For this i made this function and this modification.


Dogfalo:
In #1161 you can have multiple modals open now. So you can just open a modal in a callback for a switch functionality.

PR #838, opened 2015-03-04
update dropdown.js to improve hide dropdown handling when belowOrigin se...

### LINKED ISSUE #837: Dropdown not hiding properly with hover option
If you have a dropdown with `belowOrigin: true` and `hover: true`, and you mouseout of the activation element without touching the dropdown-content, the dropdown does not hide.

i.e. If you have a navbar with multiple dropdowns, and you drag your mouse along it, you end up with multiple visible dropdowns.
screen shot 2015-03-04 at 3 08 25 pm

apolzon:
...tting is true

This resolves #837


Dogfalo:
This has been fixed in one of the recent commits on master

PR #844, opened 2015-03-05
Restructure the scss to save 1464 lines

Antoine-360pixel:
Refactor scss to save 1464 lines, juste a little bug sass 3.4.x. outputs "\f02b" to "" ,
to remedy this I add a "@function unicode" available on https://github.com/sass/sass/issues/1395
it will just remove it after the bug is fixed and add quotes around unicode in _ionicons-variables.scss.
Enjoy !


Dogfalo:
Looks good, thanks! I'll keep an eye on the Sass issue

PR #846, opened 2015-03-05
scrollSpy bugfix

jakewhiteley:
Updated scrollSpy to use the correct selector when applying scrollTop animation.

This fixes the issue where scrolling to an element while using a fixed navbar, results in the element being obscured behind the nav.

*edited from a previous request*


Dogfalo:
We made some changes so that the scroll moves it more to the middle of the page instead.

PR #871, opened 2015-03-09
Avoid fakepath in file fields

kmmbvnr:
Chrome browser substitutes filepath for fake value. So, let's show only filename.

forms - materialize


Dogfalo:
Looks much better/cleaner. Thanks!

PR #874, opened 2015-03-09
Update waves.js

pomahtuk:
Fixing waves effect so it will be able not to throw errors when user taps/clicks on SVG elements.

Original issue - I am using C3 graphs and when I click on legend to switch graph series errors are thorn into console. Errors are present due to className check performed in code. There are no 'indexOf' method for SVG className property.

So I am basically checking if we could use this property.

### DIFF
--- a/js/waves.js
+++ b/js/waves.js
@@ -259,7 +259,7 @@
         var target = e.target || e.srcElement;
 
         while (target.parentElement !== null) {
-            if (target.className.indexOf('waves-effect') !== -1) {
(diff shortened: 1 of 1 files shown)


coliff:
I'd recommend that Materialize updates the Waves.JS library to the latest version: Waves v0.6.3
http://fian.my.id/Waves/static/waves.min.js

pomahtuk:
You are right, it's better to upgrade to new version go Waves.js, I'd rather propose changes introducing newer version.

Dogfalo:
Thanks @pomahtuk

PR #881, opened 2015-03-10
Waves update

pomahtuk:
Better approach to https://github.com/Dogfalo/materialize/pull/874

Updating a Waves.js library and keeping existing changes in place.

Original issue - I am using C3 graphs and when I click on legend to switch graph series errors are thorn into console. Errors are present due to className check performed in code. There are no 'indexOf' method for SVG className property.


Dogfalo:
Merged.

PR #885, opened 2015-03-10
port parallax to vanillaJS to use without jquery

digitalkaoz:
this PR is port of this jquery plugin to vanillaJS.

the jquery plugin is still registered if possible. additionally we allow it to be used with commonjs (browserify/requirejs environments)


auvipy:
any problem wid jquery?

PR #887, opened 2015-03-11
Modified the responsive table to retain header column height on empty header cells.

snappr:
When the responsive table hits the medium-and-down breakpoint, the
header cells stack vertically. This works well, except when a header
cell is empty, the stack of cells shift up, causing the header and body
to misalign.

image

This fix adds a no-break space character to td::before within thead,
keeping the table sections properly aligned.

image


acburst:
Thanks for the catch!

PR #896, opened 2015-03-11
Generalized button seletors

Stemlet:
`a.btn-floating` is too specific. It doesn't allow the use of `button` elements for example.


Stemlet:
Oops, sorry I'm still learning git. `_buttons.scss` would need the `a.btn-floating` changed as well.

Dogfalo:
I just merged it, although it doesn't detect it in git because I rebased to add the compiled changes.

Issue #895, opened 2015-03-11
Disabled Style for Text Inputs


Dogfalo:

Dogfalo:
Added in ef11bde322bdd2aef348d09c46523d7bab26d02a

PR #908, opened 2015-03-13
Updated to handle disabled input elements.

bradcypert:
Tackling card #895 !


Dogfalo:
I'm actually working on this right now. I do like the look of the dotted line, I also restructured some of the scss so I won't be merging this in however.

Dogfalo:
Added in ef11bde322bdd2aef348d09c46523d7bab26d02a

PR #909, opened 2015-03-13
datetime-local and time input fields get the standard form field styling

tristanseifert:
Even though materialize doesn't present the fancy popup with datetime-local and time fields, they should still get the nice styling in forms.


Dogfalo:
I agree

PR #913, opened 2015-03-14
Destroy select

### LINKED ISSUE #819: Destroy/Remove selects
While developing a Single page application, I noticed there is no way to remove, or destroy a material select. I can remove the `select` tag, but the `ul` created for styling is not removed, and I couldn't find any way to remove it with the jQuery library provided. This can become a problem as the user navigates through the application and the DOM elements don't get removed, resulting in a memory leak.

The best way to do it is adding a overload to `$.fn.material_select()`, passing an argument to destroy the select. So it would be something like
```javascript
$('#myselect').material_select('destroy');
```
Another way would be with an jQuery event, kind of like angularjs does:
```javascript
$('#myselect').trigger('$destroy'); //angularjs uses the $destroy event before removing directives.
```

After a quick look at your code, I think I can implement either for you. Just tell me what of the options you want and I'll fork it and implement it.

viniciusmelquiades:
When calling `material_select('destroy')` will remove the `ul` created for the itens, and revert the select back to its uninitialized state.

I didn't know if I had to commit the `bin/materialize.js` file or not, so I didn't.

fixes #819

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -227,6 +227,12 @@
         $('ul#select-options-'+lastID).remove();
       }
 
+      // If destroying the select, remove the selelct-id and reset it to it's uninitialized state.
+      if(callback === 'destroy') {
+          $select.data('select-id', null).removeClass('initialized');
+          return;
+      }
+
       var uniqueID = Materialize.guid();
       $select.data('select-id', uniqueID);
       var wrapper = $('<div class="select-wrapper"></div>');


Dogfalo:
Thanks! Documentation also added.

carlosperate:
Thanks Vinicius! This has definitely helped me fix my issue was well :)

PR #922, opened 2015-03-15
toast.js modification : Manage multiple classes passed as parameter.

TyrionGraphiste:
File modification to manage multiple classes passed as parameter.


Dogfalo:
Thanks for fixing this!

PR #924, opened 2015-03-15
Fix undesired color in buttons or collections inside cards

r-darwish:
Placing a button or a collection with links in a card cause it to inherit the card link color, which is undesired. Also, card link color was exported to a variable.

### DIFF
--- a/sass/components/_cards.scss
+++ b/sass/components/_cards.scss
@@ -26,13 +26,13 @@
     }
   }
 
-  a {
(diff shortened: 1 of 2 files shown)


acburst:
Thanks for bringing this issue up. Instead of hardcoding exceptions, I decided to instead move the anchor styles to only the card action section

Fixed in efb070131d50cc63bde55f2e344634c435ecb49c

PR #932, opened 2015-03-16
Remove left floating of col from small screen

kamalkhan:
The left floating for the .col does not play well in small screens. I guess float:left should be used in .col only when medium and up.

### DIFF
--- a/sass/components/_grid.scss
+++ b/sass/components/_grid.scss
@@ -48,7 +48,6 @@
   }
 
   .col {
-    float: left;
     @include box-sizing(border-box);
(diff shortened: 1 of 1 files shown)


Dogfalo:
What do you mean by ".col does not play well in small screens"? What is the problem?

kamalkhan:
On small screen size, if a row (.row) and columns (.col) within, are used, the row container does not stretch full width, hence, leaving some space left over on the side of the screen and squeezed the columns.

kamalkhan:
Could not replicate the bug, seems to be working now. Closing.

PR #935, opened 2015-03-16
Update forms.html documentation

yowkah:
Include instructions about manipulating the select box after initializing it. Suggested updated of documentation after question in gitter.im chat.


yowkah:
editted wrong file. will create new pull request.

PR #937, opened 2015-03-16
Added small button

pajasevi:
I created a small version of button - very useful in tables, as footer buttons and maybe in cards.


Dogfalo:
The regular buttons are already small. I don't think these extra thin buttons are necessary / look good.

benjamin79:
+1 for small button

jasonok6:
+1

PR #945, opened 2015-03-18
Do not close menu when it should stay open.

yourivdlans:
Right now when the option `closeOnClick` is true for the `sideNav` it will also close when the sidenav is not collapsed. This PR will make sure it stays open when it should.

### DIFF
--- a/js/sideNav.js
+++ b/js/sideNav.js
@@ -80,6 +80,10 @@
             }
           }
 
+        if (window.innerWidth > 992) {
+          menuOut = true;
+        }
+
         // Window resize to reset on large screens fixed
         if (menu_id.hasClass('fixed')) {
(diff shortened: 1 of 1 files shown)


Nick-Riggs:
+1 There is a similar PR here: https://github.com/Dogfalo/materialize/pull/2070

Why are these PRs not being merged? It seems like the obvious way it should work - is this in some way a breaking change?

TheLinx79:
+1

bigggge:
+1, I have the problem too

friek108:
I think you mean `window.innerWidth < 993` not `window.innerWidth > 992` in the commit.

Dogfalo:
Fixed via #4299

PR #951, opened 2015-03-18
Add a .hide class to hide an element

brandonjmckay:
I wanted a class to toggle on, and off to be able to hide elements in css.

I think this is a decent thing for the framework to handle; Materialize already has a lot of `.hide-on-med-and-up` style classes.

I find I use this a lot, and usually get this function from Bootstrap. I use this when I want Ember to render a template, but not display it yet.

### DIFF
--- a/bin/materialize.css
+++ b/bin/materialize.css
@@ -2102,6 +2102,9 @@ ul.staggered-list li {
 /*********************
   Media Query Classes
 **********************/
+.hide {
+  display: none !important; }
+
 @media only screen and (max-width : 600px) {
   .hide-on-small-only, .hide-on-small-and-down {
     display: none !important; } }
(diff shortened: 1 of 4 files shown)


Dogfalo:
Thanks for this, but I had to make the change manually as there were too many other changes in your commit, and the wrong documentation file was edited.

See 296a3263b7c3d40ae221398414e911acd3200eb1

brandonjmckay:
I think the other changes were partly the way that grunt compiled my css. Are there any special configs to be aware of?

Dogfalo:
If you used our grunt script, it should be exactly the same format.

brandonjmckay:
grunt script? Not sure which that was. I followed online docs on how to install grunt, then ran `grunt watcher` or something.

This was the first I've used grunt. May have botched it.

brandonjmckay:
Good to know it should look the same, I'll keep that in mind next time.

PR #955, opened 2015-03-19
feature/character-counter

chukitow:
Character counter feature based on Text fields components
http://www.google.com/design/spec/components/text-fields.html#text-fields-character-counter

Tell me if you liked or not and what can I do to improve it :+1:


chukitow:
@Dogfalo I would be grateful if tell me your opinion about this :D

acburst:
Hey this is looking pretty good. I just have a couple things that would be nice before merging.

1. An important style change: make sure that when you add the counter it does not affect the flow of the document. Right now it pushes the other content down which is undesirable. 

1. First off, this doesn't need to be its own page. If you could remove this and integrate it into the existing forms page somewhere that would be better. 

1. Also if you want, you can add the javascript to the forms.js file as well to reduce the clutter of js files. 

1. Also there is an inconsistency in your demo where the textarea has a length of 20 but the code demo says 120. 

1. If you could also update it so it can be automatically merged, that would make things a lot easier.


Thanks so much for the help. It's looking really good

chukitow:
@acburst , according to your comments I'll apply those changes, Thanks you :+1:

chukitow:
@acburst, can you review it again? I prefer to keep the code in a separate file to follow the single responsibility principle ( one file per module, component, etc)

acburst:
Yes, its looking good but you didn't address the document flow problem. Do you mind looking into that again?

Maybe just make the counter position absolute or something. 

It just doesn't look right when the character counter is added and then the whole page shifts down and up when it disappears.

chukitow:
I added the proper code to fix when the character counter is added.

acburst:
awesome! thanks so much for this. I'm going to merge it in manually.

PR #957, opened 2015-03-19
More usefull to select a class

jonathanpath:
Otherwise, I can't use nav for anther personnel components.

### DIFF
--- a/sass/components/_navbar.scss
+++ b/sass/components/_navbar.scss
@@ -1,4 +1,4 @@
-nav {
+.nav {
   color: #fff;
   @extend .z-depth-1;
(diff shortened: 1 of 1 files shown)


carlosperate:
I'm not sure this is a such a good idea, as it would break pretty much everybody's implementation of materialize.

jonathanpath:
Why not replace nav by .nav everywhere it is used ?

ipalaus:
Because you can't replace it on people's websites. This is a breaking change and it's not backwards-compatible.

carlosperate:
Closing as this is a breaking change.

PR #958, opened 2015-03-19
Put a relative line-height on h*

jonathanpath:
It's not usefull to have a absolute line-height when I need to change the font-size


Dogfalo:
Merged, thanks!

Issue #947, opened 2015-03-18
When button is disabled you can still click on icon


albertb22:
If you have a button tag that is disabled and you have an icon in it. You can still click on the icon to get the waves effect.

TheRinseM:
+ 1

TheRinseM:
http://0.media.dorkly.cvcdn.com/36/35/6603dc5a9292104b44c349b85b5aaf7a-5-crazy-fan-theories-that-make-total-sense.jpg

coltonfitzgerald:
Under what circumstances? I'm trying to recreate the issue but I can't seem to. Can you post the HTML markup that it doesn't work with?

TheRinseM:
@coltonfitzgerald The code we used:
``` 
<div class="input-field right-align right">
      <button id="submitLoginForm" class="btn waves-effect waves-light blue darken-4">Login
              <i class="mdi-content-send right"></i>
    </button>
</div>

lamoni:
I've submitted pull #970 as a fix for this.

TheRinseM:
@lamoni Thank you for the fix!

lamoni:
@TheRinseM gladly!

coltonfitzgerald:
But here's the thing. (Correct me if I'm wrong) I thought you use the class ".disabled" to disable buttons whereas you are setting the attribute "disabled="true"" with that line of jQuery

lamoni:
Adding class of 'disabled' to buttons actually doesn't stop the waves-effect either.  Assuming someone else doesn't do it, I'll add that check in as well.

albertb22:
Thank you for the fix.

Dogfalo:
Fixed in 345af00

PR #970, opened 2015-03-22
Fix for Issue #947.  By checking if any of the parent elements are disab...

lamoni:
...led, we can disable the delegation for the icon click.

The wave-effect delegation wasn't taking into account if any of its parents were disabled.  So even if the button was disabled, a child <i> icon element would still trigger the waves-effect.

Code used to recreate issue:

```html

  <!DOCTYPE html>
  <html>
    <head>
      <!-- Compiled and minified CSS -->
      <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/materialize/0.95.3/css/materialize.min.css">

      <!--Let browser know website is optimized for mobile-->
      <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no"/>
    </head>

    <body>
    <div class="input-field right-align right">
          <button id="submitLoginForm" class="btn waves-effect waves-light blue darken-4">Login
                  <i class="mdi-content-send right"></i>
        </button>
    </div>

      <!--Import jQuery before materialize.js-->
      <script type="text/javascript" src="https://code.jquery.com/jquery-2.1.1.min.js"></script>

      <!-- Compiled and minified JavaScript -->
      <script src="assets/materialize/bin/materialize.js"></script>
          <script>
            $(document).ready(function() {
                $("#submitLoginForm").prop("disabled",true);
            });
          </script>
    </body>
  </html>
        

```

### DIFF
--- a/js/waves.js
+++ b/js/waves.js
@@ -260,7 +260,7 @@
 
         while (target.parentElement !== null) {
             if (!(target instanceof SVGElement) && target.className.indexOf('waves-effect') !== -1) {
-                if (target.attributes.hasOwnProperty('disabled')) {
+                if (target.attributes.hasOwnProperty('disabled') || target.className.indexOf('disabled') !== -1 ) {
                     return null;
                 }
                 element = target;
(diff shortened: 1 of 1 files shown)


Dogfalo:
We fixed this by disabling click on children of these disabled buttons. I think this is better as it is expected that clicking on a disabled button does nothing at all.

PR #976, opened 2015-03-22
Fixed filename in collections.jade

shawntoffel:
The collections_content.html filename was wrong in the jade file causing the grunt task to fail.


Dogfalo:
Thanks!

PR #987, opened 2015-03-24
Update _typography.scss

Lardjo:
Why do we have a style for light (300) font-weight, but there is no style for the thinnest 200? Maybe I don't understand? I think this should be added. One or another way.


Dogfalo:
Looks good, thanks!

PR #994, opened 2015-03-24
Fixed issue with default and media-query nav.brand-logo.center

wenkesj:
Fixed the transform when the logo would jump -50% in the y direction to be stable at 0%. Just fixed the CSS


wenkesj:
Possibly add class for vertical brand-logo?

Dogfalo:
We added this a while back, I think because you did not modify the source files we could not merge your changes.

PR #997, opened 2015-03-25
Added manualHide on dropdown (and docs)

auser:
Added the option `manualHide` to the dropdown plugin.

This is useful in cases where you're integrating with a framework that depends on the element being available (in my case, react).

Usage:

``` javascript
$(element).dropdown({manualHide: function(evt, hideCallback) {
   // evt is the event object
  // hideCallback is the function to hide the element
});
```


acburst:
This is out of scope for our framework

PR #1008, opened 2015-03-25
Fixes #1007

### LINKED ISSUE #1007: Dynamically Add Collapsible Items
Currently, if a new li is added to a collapsible ul, the newly added item will not be expandable. Only the collapsible-header will be shown.

If .collapsible() is called again on the ul after the new li has been added, all items within the collapsible list will collapse.

The collapsible plugin should correctly being called multiple times to support the addition of new items to the list, while ideally maintaining the active status of the current items in the list.

pcorey:
When re-initializing collapsible plugin, expanded items will stay
expanded.


Dogfalo:
Great fix, thanks!

PR #1009, opened 2015-03-25
js/leanModal: remove extranneous line which broke modal display

kabaka:


dhawken:
The "extranneous" spelling is the subtlest joke ever.

PR #1010, opened 2015-03-26
Update materialize.css

ibrainventures:
without this the slide-in sidenav (not fixed) has a href only for the textsize, so its very uncomfortable.


Dogfalo:
You have to edit the Sass files if you want them to be merged in.

PR #1013, opened 2015-03-26
Removing active class from form labels on 'reset' trigger.

johnschult:
When the reset trigger is sent to the form, it should also remove the
“active” class from any form labels.


acburst:
Thanks for catching that

johnschult:
You bet :)

--john

PR #1023, opened 2015-03-28
Wroong padding-left: avatar collection with header

aron-bordin:
I'm sorry, this is my first time with materialize, I just had a problem and posting here how I fixed it. As long as I don't know the project, maybe this PR in not useful. 
When using a collection with header to show collection itens with avatars, there is a problem with the padding-left:

screenshot from 2015-03-28 01 13 54

The problem:
screenshot from 2015-03-28 01 14 54

So I just added !important here to use it with my project.

### DIFF
--- a/css/ghpages-materialize.css
+++ b/css/ghpages-materialize.css
@@ -2256,7 +2256,7 @@ td, th {
     border-bottom: 1px solid #e0e0e0; }
     .collection .collection-item.avatar {
       height: 84px;
-      padding-left: 72px;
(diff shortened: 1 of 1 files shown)


Dogfalo:
You didn't modify the sass file and adding !important is not a good fix

aron-bordin:
Yeah, I know. But as I told you, I'm not experienced with Materialize, so I'm not able to fix it. I just posted this bug here and a simple solution to help someone to fix it.

PR #1025, opened 2015-03-28
Added a Button Container and added the Documentation for it

### LINKED ISSUE #1022: Feature Request: Same Width Buttons
Could you possibly provide the ability to let a stack of Buttons have all the same width?

For example the width of the biggest button in the group or maybe block style buttons that take the full width of their container...

Imagine a screen with 2 buttons in the center, Register & Login, vertically distributed and horizontaly centered... it would look nicer if they have the same width :)

HenriBeck:
This closes #1022


Dogfalo:
I don't understand why packaage.json is deleted. There are also many unrelated changes in here. Please make changes that are only directly related to the pull request.

PR #1039, opened 2015-03-29
Replaced height with min-height

Suero:
This will make it scale properly on small devices, previously text got cut off.


Dogfalo:
You have to change the sass file or we can't merge it

Dogfalo:
You modified a compiled file instead of the source, so we made the change manually.

Issue #1038, opened 2015-03-29
Color Suggestion


MageGenius:
Add the black and white background colors from the stylesheet to the colors page.  Also, add a "color" class for transparent.  I've created a demo for the ideas at the bottom of this page: http://material.magegenius.com/fabric-color-palette

acburst:
If you want you can add this to our documentation files and compile everything into a pull request

MageGenius:
Another idea for the colors would be to include the ability to add hover colors with the Material color palette. 

Meaning for each color on http://materializecss.com/color.html add those same colors to hover states.  For example, here is a current snippet of the Red color:

.red {
  background-color: #F44336 !important; }

.red-text {
  color: #F44336 !important; }

Using a similar naming format as the 'red-text' if developers wanted to use these same colors for hover colors, these could be updated to:

.red, .red-hover:hover {
  background-color: #F44336 !important; }

.red-text, .red-text-hover:hover {
  color: #F44336 !important; }

Could be useful for hovering over elements that already have a background color or colored text links.  If this is something that would be a good addition to the project I can create the CSS and do a pull request if you like.

acburst:
Merged 950aa8cd1064dbcb5b4e1b0ed344c9a1cb0967f6

PR #1041, opened 2015-03-30
Adding colors to documentation

MageGenius:
This adds an example of the black and white color classes to the colors documentation.  Also, this adds a transparent class to the documentation and CSS.  The transparent class can be used in the same way as any other color class.  For consistency with the other color classes the transparent-text class is included but this is far less likely to be used.

This commit is related to comment: https://github.com/Dogfalo/materialize/issues/1038

If the changes need to be added or moved to another file please let me know and I will update them appropriately.

### DIFF
--- a/bin/materialize.css
+++ b/bin/materialize.css
@@ -1606,6 +1606,12 @@
 .white-text {
   color: #FFFFFF !important; }
 
+.transparent {
+  background-color: transparent !important; }
+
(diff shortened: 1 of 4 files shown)


acburst:
You should take a look at how we compile our files and add it to the components. You can adjust the _colors.scss for the css and the color_content.html file for the jade template.

PR #1052, opened 2015-03-31
Fixes issue where comment created invalid top-level expression

rtablada:
When the _normalize.scss file is run through a SASS processor with linting, the statement on line 290 fails because the top level expression is interrupted by a comment block.

```scss
button,
html input[type="button"], /* 1 */ 
input[type="reset"],
input[type="submit"] {
  -webkit-appearance: button; /* 2 */
  cursor: pointer; /* 3 */
}
```

Attempting to retain the comment yet remove linting errors due to invalid syntax, I have opted to move the button input selector for Android 4.0 as the first in the selector list and moved the comment to the start of the line.

### DIFF
--- a/bin/materialize.css
+++ b/bin/materialize.css
@@ -1834,7 +1834,8 @@ button, select {
  * 3. Improve usability and consistency of cursor style between image-type
  *    `input` and others.
  */
-button, html input[type="button"], input[type="reset"], input[type="submit"] {
+/* 1 */
+html input[type="button"], button, input[type="reset"], input[type="submit"] {
   -webkit-appearance: button;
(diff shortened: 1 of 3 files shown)


rtablada:
The error reported was `./bower_components/materialize/sass/components/_normalize.scss, line 290` `invalid top-level expression`

rtablada:
Error found using ember-cli-sass -> broccoli-sass -> node-sass -> lib-sass

truenorth:
@acburst It would be nice if you could tag a new release with this fix. It's currently breaking a common set of dependencies for materializecss Ember.js apps

PR #1063, opened 2015-04-01
Allows dynamic passing of `activates` for dropdown element

rtablada:
When creating drop downs (especially dynamically or in frameworks like Ember) it would be very helpful to be able to pass an element to the dropdown function rather than relying on a `data-activates` and matching id.

This PR has a flexible API, if no `activates` option is passed, original functionality is retained.
If a string is passed in, this will be used as the query selector to find the activates element.
Otherwise, it is assumed that the `activates` option is a jquery element.


rtablada:
This has now been updated to work with the most recent master

rtablada:
Any new thoughts on this?

The data activates makes things a bit awkward to work with when building things from JS since it requires dynamic id creation and populating both the link and `ul`.

js/dropdown.js.orig
```diff
@@ -0,0 +1,221 @@
...
+      if (origin.data('gutter') != undefined)
+        options.gutter = origin.data('gutter');
+      if (origin.data('beloworigin') != undefined)
+        options.belowOrigin = origin.data('beloworigin');
+    }
+
+    updateOptions();
+
+    // Attach dropdown to its activator
+<<<<<<< HEAD
```
dtracers: looks like you still have merge conflicts in the files

dtracers:
this will be very nice for html5 custom elements too because right now jquery can't find any elements inside the shadow dom.

So passing elements in would make a lot of things much easier!

PR #1066, opened 2015-04-01
In some case, the determinate is not aligned at left

jonathanpath:


acburst:
Please describe your problem more.

jonathanpath:
I did a player, and add the progress component for the volume. I used flexbox, and the determinate div was not aligned on the left.

screenshot 2015-04-09 11 11 34

Dogfalo:
Thanks for clarifying with a picture.

PR #1070, opened 2015-04-01
Adding hover colors for background and text

MageGenius:
This PR add hover classes to change both the background color and text color. It works with all base and shade colors.  The documentation has been updated also.


MageGenius:
If it would help, I can make a demo of this pull request.

Dogfalo:
This would be adding lot of page weight for classes that will be used very rarely. I don't think this is would be used enough to warrant adding this many lines.

PR #1071, opened 2015-04-01
disabled pointer events on an inactive form label

jpopesculian:
Clicking on a label wouldn't trigger the input, making it seem disabled. I disabled pointer events on a label to allow the click to go through and activate the input and then reenabled them when the input is active

### DIFF
--- a/bin/materialize.css
+++ b/bin/materialize.css
@@ -6066,6 +6066,7 @@ input[type=text], input[type=password], input[type=email], input[type=url], inpu
     left: 0.75rem;
     font-size: 1rem;
     cursor: text;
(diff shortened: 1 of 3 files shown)


acburst:
Where are you seeing labels not triggering the input?

jpopesculian:
I've implemented it into a web app that I've been making, but the webapp uses react and es6 module loaders, so the implementation has been a little bastardized. I don't know if I'll be able to reproduce it due to the complexity, but I figured it was a harmless fix.

acburst:
if the label for attribute is set correctly, it should refocus onto the input. disabling pointer events would prevent this from happening

PR #1083, opened 2015-04-02
rewrite character counter code

ckihneman:
This is a drop in replacement, all functionality kept intact.

The old character counter code had hoisted variables,
hardcoded values all over the place, and was generally
not well formatted. This update creates more modular code
and gives hook points for updating some of the values
it uses. I think this is step one, and we need to also
remove it's self initialization so options can get passed
across, or global defaults for the counter can be updated
to the developers liking.

---

This is the general direction I'd love to see all of the JS moving towards. I saw this as an easy way to convey I'm down to help, and here is how I would structure things. The project also really needs a light require system so we can build dependencies easier. Right now its really hard to break the JS apart and load only what you need. The self initiation also really gets in the way and would be better left to the developer to handle.

For example: lets say the developer wanted the counter font size to be `16px`. They could globally set that:

```javascript
$.fn.characterCounter.CharacterCounter.defaults.counterCss['font-size'] = '16px';
```

Or pass it in with the options:

```javascript
$('input, textarea').characterCounter({
  counterCss : {
    'font-size' : '16px',
  }
});
```


ckihneman:
Bummer this never made it in, the code still needs this help. Closing the issue. If you are interested, reach out. Cheers.

PR #1096, opened 2015-04-04
Fix #1084 clicking icon doesn't work in collapsible

### LINKED ISSUE #1084: Clicking on icon doesn't work in Collapsible
In http://materializecss.com/collapsible.html
Clicking on the icon doesn't make the item expand.
But it works for turning it back to original state.

deerawan:
add handling for children element of collapsible header


acburst:
Merged manually because of conflicts. Thanks for the help!

6352aa856bbb47fa7a4536d672c129bc76c10afb

PR #1097, opened 2015-04-04
Fix vertical alignment of pagination

TimonVS:
The list items of the pagination already have a defined height, so using the line-height property (equal to the height of the item) you can get a proper vertical alignment.


Dogfalo:
Merged, thanks!

PR #1131, opened 2015-04-08
Added event triggers for tabHide and tabShow

thecodethinker:
I needed to add some animations to the changing divs with the tab component so I added some event triggers with jQuery to the hide() and show() calls in tab.js.


thecodethinker:
Oh just noticed a couple bugs. I'll be back when I fix them.

PR #1142, opened 2015-04-10
Correct example for Pushpin

epiquiem:
The table of contents is on the left, not the right.


Dogfalo:
No the pushpin example is on the right.

PR #1154, opened 2015-04-13
Add option to have a 5 columns grid layout

sKINNEEE:
It's quite common the use of 5 columns and with your way it would only be possible to use  4 columns (s3, m3 and l3) or 6 columns  (s2, m2 and l2). 

So I added s25, m25 and l25 (not sure about the naming) that will result on 5 column layout.

### DIFF
--- a/sass/components/_grid.scss
+++ b/sass/components/_grid.scss
@@ -59,6 +59,13 @@
         width: $perc;
         margin-left: 0;
       }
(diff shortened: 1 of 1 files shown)


Dogfalo:
The naming is quite confusing and I think that this isn't something we want to add. Later on, we may work more on grid enhancements.

PR #1157, opened 2015-04-13
optimize leanModal;

bbflame:


Dogfalo:
There are too many style changes and 4k line additions?

PR #1161, opened 2015-04-14
Allow multiple modals

### LINKED ISSUE #1021: Multiple not dismissible modal
If you try to open a new not-dismissible modal from a previously opened modal you loose the ability to interact with the original modal once the new modal get's dismissed.

ciencia:
This change allows multiple modals to be opened at the same time.

How it has been done:
 - Don't use a static ID for the overlay. Use a class instead
 - Store references to the associated overlay element so each modal
   can reference its own overlay. We're generating an ID to store the reference
   so we only need to have a string as a reference instead of a whole jQuery
   object
 - Populate the z-index CSS property in JavaScript, so we can increase it for
   each new stacked modal

Also, reuse jQuery objects instead of querying the DOM each time.

fixes #1021

### DIFF
--- a/js/leanModal.js
+++ b/js/leanModal.js
@@ -1,10 +1,13 @@
 (function($) {
+    var _stack = 0,
+    _lastID = 0,
+    _generateID = function() {
+      _lastID++;
+      return 'materialize-lean-overlay-' + _lastID;
+    };
+
   $.fn.extend({
     openModal: function(options) {
-      var modal = this;
-      var overlay = $('<div id="lean-overlay"></div>');
-      $("body").append(overlay);
-
       var defaults = {
         opacity: 0.5,
         in_duration: 350,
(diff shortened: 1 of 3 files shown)


danielhickman:
+1
https://www.youtube.com/watch?v=QnXJD-Qtuyw

Dogfalo:
Great work!

viniciusmelquiades:
Is this already in bower? I'm using the latest version available in bower, but the modal is still using the version with the id for the overlay.

ciencia:
@viniciusmelquiades I did this when materialize v0.96.1 was out. I think it will go to bower once they bump the version

amitlzkpa:
@ciencia 
Can someone please show a demo of how to do this?
I am trying to create a modal with a button, which opens another modal.
Help much appreciated.

ciencia:
@amitlzkpa There's nothing special you need to do, open the second modal like if it wasn't inside another modal... That "should" work, but I can't tell you because where I'm using this library I open all my modals with JavaScript

amitlzkpa:
Thank you for the quick reply.

When I do that the overlay covers the upper modal. Please have a look at
the attached image.
The one in in the background - with 'CMU Course Reviews' written is the
main page.
The one behind with header as 'Taken' is a bottom-sheet modal.
When I click a href link on the bottom-sheet modal it should open the notes
modal, but in the front. Currently, the black-overlay covers the notes
modal and I can't interact with it.

Moreover, when I hit 'escape', it exits both the modals. It would be ideal
to have the escape exit only the top-most modal.
Any ideas?

[image: Inline images 1]

ciencia:
@amitlzkpa what you describe seems to be exactly the situation before my fix. What version of materialize are you using? Can you share the code you're using (simplified, no need to be real content in the modals)?

amitlzkpa:
Sure.
I updated with v0.97.4, but am still getting the same problem.
Here's the HTML


<html>
<head>
</head>

<body>

<script>
    $(document).ready(function(){
    $('.modal-trigger').leanModal();
    });
</script>


<div class="section">
    <div class="row">
        <div class="col s12 m12 l12">
        
        {% for row in groupInfo %}
            
            <td class="center">
                <a class="modal-trigger"
href="#wish-modal-for-prog-wishlist-{{ row.id }}"><i class="tiny
material-icons grey-text">add</i></a>
            </td>


            <div id="wish-modal-for-prog-wishlist-{{ row.id }}"
class="modal modal-fixed-footer">
                <div class="modal-content">
                    <h4>Notes</h4>
                    <form method="POST" action="{% url '...' %}">{%
csrf_token %}
                        <textarea class="materialize-textarea"
name="notes">{{ row.usernote }}</textarea>
                </div>
                <div class="modal-footer">
                    <a href="#!" class="modal-action modal-close
waves-effect waves-grey btn-flat left">Close</a>
                    <button class="modal-action modal-close waves-effect
waves-green btn-flat" type="submit" name="save">Save
                    </button>
                    </form>
                </div>
            </div>
            
        {% endfor %}
        
        </div>
   </div>
</div>


</body>
</html>



I am sure rest of the thing works, as I was able to work with properly
before trying to put it in the bottom sheet.
It's served by a Django server.
Thanks.

ciencia:
@amitlzkpa Your code seems to break because of the `{{ row.id }}` stuff inside element IDs: http://codepen.io/anon/pen/oxWpmj

Removing that stuff from the IDs makes it to work. Still, I only see one modal on your example.

PR #1167, opened 2015-04-15
fix format in doc for forms

nanowang:


acburst:
Thanks for pointing this out. I added the change manually because we use Jade templating files.

Fixed in 1d60870

PR #1176, opened 2015-04-16
materialbox: spell check on initialized class

sirhodes:
The class added to the images when materialbox is initialized is "intialized".

Shouldn't this be "initialized"?


Dogfalo:
Thanks for catching this!

PR #1207, opened 2015-04-21
ability to change navbar font color and brand font size

gebidesign:
added to _variables.scss and _navbar.scss:
- $navbar-font-color and 
- $navbar-brand-font-size  

and set them to default materializecss values.


acburst:
Thanks for this

PR #1209, opened 2015-04-21
Fixed Showcase entries

edmael:
Removed unreachable sites, added a site (ragnate.la) that is using Materialize.css.


Dogfalo:
Please do not bundle your own showcase site within other changes. Submit it in an email and we will add it if we wish.

edmael:
I did not know that, I'll provide the change without my entry (those unreachable site are kind of useless).

Thanks.

seanmavley:
Can one expect feedback if their showcase doesn't get included? Submitted a showcase, and I have no idea if the email even went through.

PR #1211, opened 2015-04-22
Removed unreachable sites

edmael:
Hi,
I've removed the unreachable sites from the showcase (as I did in my previous commit, but without adding anything new).


Dogfalo:
We've decided to leave them up for now.

PR #1215, opened 2015-04-22
Add push and pull columns

acidvertigo:
adding push and pull columns. With this is possible to better control responsive divs order 
Example "push-l8" or "pull-l8"


acidvertigo:
Closing as this has been already proposed

PR #1234, opened 2015-04-24
Double Quotes on href

DaEckster:
XHTML technically requires double quotes on hrefs, also, was kind of weird having both types one line after the other. Changed just to be normalized at least. Was throwing errors with HTMLint


Dogfalo:
Thanks for the fix!

PR #1235, opened 2015-04-24
Remove closing input tag from docs

DaEckster:
Was reusing the code and had Meteor throw a major hissyfit, don't need that closing input tag.


Dogfalo:
Thanks again!

PR #1256, opened 2015-04-27
Use `script.` instead of the deprecated version

mondaychen:
Stop the grunt task from reporting
````
jade/_head.jade, line 31:
Implicit textOnly for `script` and `style` is deprecated.  Use `script.` or `style.` instead.
````


Dogfalo:
Thanks!

Issue #436, opened 2015-01-11
add touch gestures to range input


Bound3R:
add an input range but when used in a mobile device, I can not see the value you are selecting, seeing the code I see it only works with mouse events :(

theY4Kman:
Is a fix for this in the roadmap?

ABeltramo:
No one found a solution??

ABeltramo:
Quick fix. I can turn this into a pull request if someone approve this code.
![untitled](https://cloud.githubusercontent.com/assets/1122217/7378120/74e954e2-edeb-11e4-9725-3edb12df0b95.gif)

Put the following code in a .js file and just include it. You don't need to mess with the core library.

```javascript
	var range_type = 'input[type=range]';
    var range_mousedown = false;

    $(range_type).each(function () {
      var thumb = $('<span class="thumb"><span class="value"></span></span>');
      $(this).after(thumb);
    });

    var range_wrapper = '.range-field';

      $(document).on("touchstart", range_wrapper, function(e) {
        var thumb = $(this).children('.thumb');
        if (thumb.length <= 0) {
          thumb = $('<span class="thumb"><span class="value"></span></span>');
          $(this).append(thumb);
        }

      range_mousedown = true;
      $(this).addClass('active');

      if (!thumb.hasClass('active')) {
        thumb.velocity({ height: "30px", width: "30px", top: "-20px", marginLeft: "-15px"}, { duration: 300, easing: 'easeOutExpo' });
      }
	  var left = e.originalEvent.touches[0].pageX - $(this).offset().left;
      var width = $(this).outerWidth();

      if (left < 0) {
        left = 0;
      }
      else if (left > width) {
        left = width;
      }
      thumb.addClass('active').css('left', left);
      thumb.find('.value').html($(this).children('input[type=range]').val());

    });

    $(document).on("touchend", range_wrapper, function() {
      range_mousedown = false;
      $(this).removeClass('active');
    });

    $(document).on("touchmove", range_wrapper, function(e) {

      var thumb = $(this).children('.thumb');
      if (range_mousedown) {
        if (!thumb.hasClass('active')) {
          thumb.velocity({ height: "30px", width: "30px", top: "-20px", marginLeft: "-15px"}, { duration: 300, easing: 'easeOutExpo' });
        }
        var left = e.originalEvent.touches[0].pageX - $(this).offset().left;
        var width = $(this).outerWidth();

        if (left < 0) {
          left = 0;
        }
[...]

PR #1266, opened 2015-04-28
BugFix: Now slider works also on mobile browser.

ABeltramo:
Just added event support for mobile plus the change to detect touch position on the screen.
See #436


Dogfalo:
Please edit forms.js instead or we can't merge this.

PR #1275, opened 2015-04-30
Fixes #801 - Range Slider shows old value

### LINKED ISSUE #801: Range Slider shows old value
When clicking on any point on the slider the bubble shows the old value of the slider.

gthapar:
Fixes Range Slider shows old value #801


Dogfalo:
We fixed this ourselves along with a few other problems the range input had on mobile. Thanks for the help though.

Issue #1272, opened 2015-04-30
Tabs - callback or trigger on active tab click


f1ana:
One nice feature for tabs would be the ability to run a function when the tab change process has been completed.  A valid use case would include set up that needs to happen once the new tab content is visible (e.g.  ajax calls, resize of 3rd party UI tools that cannot determine size in display:none)
I'm not sure which would be a better implementation: providing a callback on tab init or using a trigger.

Dogfalo:
Callback on tab doesn't really make sense to me. I think you should just implement it on a trigger.

f1ana:
If I implement the fix, is there a best practice for what the trigger should be named or bound to?

acburst:
Callback has been added called onShow

Issue #1278, opened 2015-04-30
Form reset button affects all forms on page


raptor008:
Resetting a single form will affect all forms on the page by removing the "active" class from labels. I believe it should only affect the form of which you are hitting the reset button.

http://codepen.io/anon/pen/MwwoXb

1. Enter text into first field of first form.
2. Enter text into first field of second form.
3. Hit the second reset button.
4. Notice that the active class for the first form's label is removed.

f1ana:
I have a fix for this in my fork- will create pull request shortly

raptor008:
Good deal, thanks.

acburst:
Fixed in c5de306

PR #1280, opened 2015-05-01
Add tab change trigger

f1ana:
Reference issue 1272
https://github.com/Dogfalo/materialize/issues/1272


f1ana:
Also references issue https://github.com/Dogfalo/materialize/issues/1278

PR #1288, opened 2015-05-03
Pagination links

Tschuky101:
Makes the links on pagination fill the <li> container for easier click ability on page numbers


nightskylark:
I was going to make the same fix and it is already done here. Great!

shinbuntu:
The PR #2931 has been merged and fixed it. Could you please confirm it ?

Issue #1323, opened 2015-05-08
Packaged Dark Theme following Material Design guidelines


DaEckster:
Material Design specifically goes out of its way to specify what a [dark theme](http://www.google.com/design/spec/style/color.html#color-themes)(see bottom), including [typography specifications](http://www.google.com/design/spec/style/color.html#color-ui-color-application), should look like in the documentation. It seems as if a dark theme implementation should be created, since it requires more than just changing the secondary and primary colors.

acburst:
I'll look into the pull request soon, I just added this into our new templates issue

kinsi55:
Hmm, did this ever get anywhere?

PR #1325, opened 2015-05-08
Adding initial dark theme changes

Ecksters:
Based on my issue #1323, I started adding a dark theme to the scss files, using a the `$dark-theme` variable, which is set to false by default, turning it to true changes the theme to dark.

Not certain if this implementation method is ideal, but I suppose I just want to try something out for it, and this at least can show what changes would need to be made to create a dark theme.

### DIFF
--- a/sass/components/_dropdown.scss
+++ b/sass/components/_dropdown.scss
@@ -1,6 +1,10 @@
 .dropdown-content {
   @extend .z-depth-1;
-  background-color: #FFFFFF;
+  @if $dark-theme == true {
+    background-color: color("grey", "darken-3");
+  } @else {
+    background-color: #FFFFFF;
+  }
   margin: 0;
   display: none;
   min-width: 100px;
@@ -21,7 +25,11 @@
     text-transform: none;
 
     &:hover, &.active {
-      background-color: #eee;
+      @if $dark-theme == true {
+        background-color: #666;
+      } @else {
+        background-color: #eee;
+      }
     }
 
     & > a, & > span {
--- a/sass/components/_global.scss
+++ b/sass/components/_global.scss
@@ -14,6 +14,7 @@ body {
   // display: flex;
   // min-height: 100vh;
   // flex-direction: column;
+    @if $dark-theme == true { background-color: #303030; }
 }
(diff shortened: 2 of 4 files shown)


DaEckster:
image

Here's just a look at what I'm thinking of.

HenriBeck:
Looks great @DaEckster

zot24:
:+1:

carloscarcamo:
:+1:

acburst:
I think instead of making this an inherent part of the framework, you can make your own theme with these dark color values replaced in the variables.scss file. I don't think we need more color variables when really this is just a color theme.

truenorth:
:+1:

Neotrixsx:
:+1:

DaEckster:
@acburst Yeah, that was something I was a bit afraid of, I don't want to clutter up the main scss code with theme, the only reason I thought it would possibly be okay is because Material design specifically shows dark and light themes, while adding colors appears to be intended for simply accenting, not as a main color scheme. This is actually why I haven't continued making more commits, since I assumed this wasn't an ideal implementation strategy in the long-run.

Would love to see an example of implementing a full theme redo, because the current SCSS makes a lot of assumptions that the background will be white, so some major changes would need to be made for the variables.scss file to support full theme changes.

acburst:
I was thinking about this, but if you wanted to make a template of sorts using a nice dark color scheme, we can look into adding it into the templates section of our documentation.

Ecksters:
Well, the main dilemma with that is that keeping that template up-to-date with future changes would be extremely difficult, since various of the scss need to be modified to achieve the dark look. Like I said, the current implementation assumes a light-colored background in many locations in the code. I think that would need to be corrected before it'd be worth putting much effort into a full dark theme. :tired_face:

ProfNandaa:
@Ecksters  

The theme looks great!

A quick one, as per the conversation above, should this PR be closed now?

ManuZiD:
Awesome work :+1:

Ecksters:
Yes, it should probably be closed as it requires major changes to the framework to add dark theme support.

PR #1277, opened 2015-04-30
Added option to dropdown, to allow the menu to expand upwards instead.

madsnedergaard:
This pull-request fixes #960 and possibly the duplicates #780 and #903.
It adds an option to make the menu appear above the activator. It might need more testing and I haven't changed the animation to slide upwards instead of down, but I hope it might be useful anyway.

PR #1330, opened 2015-05-09
add upwards dropdown functionality

edge:
I added functionality for dropdowns to drop upwards instead downwards. This is useful for dropdowns placed in footers and elsewhere.

It uses the option `up` and `data-up` for the options object and jQuery data, respectively.

When used with `belowOrigin`, the dropdown starts at the bottom of the origin.

I'll submit a follow up PR for the docs if this goes through.


Dogfalo:
Duplicate of #1277

Issue #1317, opened 2015-05-07
Hide Toasts with JS


gjuchault:
Hi,

When we use a custom HTML in Toast (with a link as in the docs), clicking on the button doesn't hide the Toast. Could you expose one method (as there is one for modals) to hide the toast ?

Thanks

zeeshanjan82:
+1

TomFreudenberg:
+1

jacek213:
+1

TomFreudenberg:
As @jacek213 pointed out, you may use jquery to drop ALL toasts by his example.

If you are using more toasts and just want to drop a specific one, you can add a virtual class as identifier to your toast message like

````
toast("message", 5000, 'red darken-4 toast-identifier-for-message');
````

That will be appended as class to your toast message, so you may run

$('.toast-identifier-for-message').fadeOut()

to just hide that message. This worked for us most the time

**Comment by yanickrochon** [comment]
Toasts may return an object that can be used to change the toast's behavior. Something like

```js
var $toastContent = $('<span>I am toast content</span>').on('click', function () {
  // toast.delay( 30000 );    change toast's delay to 30s
  // toast.discard();         discard
});
var toast = Materialize.toast($toastContent, 60000);  // 1 minute
```

Dogfalo:
Added in 87373d6b4551917e9b6d5b96afd031d521a8a285

alexisdiel:
After use` $('.toast').remove();`, I get te following error when the timeout.
![image](https://user-images.githubusercontent.com/7047656/28988854-59917824-7948-11e7-9464-e1562bb30b23.png)

retorquere:
Can something similar be used to change the text of the toast?

Asg21:
Duplicate of #

PR #1351, opened 2015-05-13
Added a toast dismiss

Shustoff:
#1317 Class toast-dismiss for toast dismissing


marcoscssno:
Is this really working?

Issue #1298, opened 2015-05-05
Fix badges position


Isprago:
Change CSS: position:absolute to CSS: float:left.
It works if you want to add 2 or more badges.

Actually, if you add 2 badges, they overlaps.

Also, let change the text "new" in order to translate the badges.

And finally, fix collections height when they have a long title and a few badges on small screen.

jpggvilaca:
I fixed it but i haven't done a pr. 
spans: position: relative  - display: inline-block  - float: right

ricardodovalle:
I've getting the same error and created a jsfiddle to test.
https://jsfiddle.net/6kqL1y7u/

0xbaadf00d:
I think .badge class don't have to define any position (relative or not) , classes .left and .rigth already exists and defining position can lead to bug problem to add tag everywhere in html page without embbed it in a position:relative div.

Merovex:
I'm running into a related issue where the badge is always on page far right, regardless of where I want it. Ricardo's fiddle above shows this isn't local to my situation.

developeranirudhprabhu:
may be need to provide nesting of badges. That would be the ideal fix?

developeranirudhprabhu:
@Isprago Can you create a fiddle and share these changes?

Native-Coder:
Is anyone having issue putting badges on the nav bar? When i do it [this](http://i67.tinypic.com/f9mau0.png) happens. I've also had problems with my search bar not taking up the entire nav bar, but I worked around that by setting a static height on the element.

acburst:
These issues should be fixed.

@Native-Coder http://materializecss.com/badges.html#navbar

Native-Coder:
I have the latest version. Still having both issues. V.98

PR #1356, opened 2015-05-14
sass/global: badge: fix positioning (Dogfalo/materialize#1298)

kabaka:
Not all points in Dogfalo/materialize#1298 are addressed -- the "new"
text is still hard-coded and cannot easily be replaced for translation
purposes.

This also resolves positioning issues when badges are used on
`.collapsible-header` containers, though line height will need to be
addressed separately before that working completely.


ricardodovalle:
:+1:

PR #1366, opened 2015-05-15
hotfix/code-improvements-on-parsing

chukitow:
Parsing integers in this way is faster. Take a look at this.

``` javascript
console.time("parseInt1");
for(var i=0; i<=10000;i++){
  var length = parseInt("100");
}
console.timeEnd("parseInt1");

console.time("parseInt2");
for(var i=0; i<=1000;i++){
  var length = +"100";
}
console.timeEnd("parseInt2");
```

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -67,7 +67,7 @@
 
     validate_field = function(object) {
       var hasLength = object.attr('length') !== undefined;
-      var lenAttr = parseInt(object.attr('length'));
+      var lenAttr = +object.attr('length');
       var len = object.val().length;
 
       if (object.val().length === 0 && object[0].validity.badInput === false) {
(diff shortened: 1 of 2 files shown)


tiekuhn:
proven by jspertest. Please see
http://jsperf.com/parsing-integers-test

tomscholz:
@DanielRuf

DanielRuf:
Direct type casting (using unary plus, double NOT bitwise, right shift unsigned and other bitwise operators) is generally faster than parseInt and other high level abstractions in programming languages.

The PR looks good.

acburst:
First off, your console test don't use the same amount of iterations. And second, the time saved is definitely negligible given the amount of times these parseInts are going to be called. I think its fine to stay with parseInts because they are definitely more legible and make the code more readable.

PR #1378, opened 2015-05-17
Added companion .css with the full color palette for waves.

alexyans:
I needed to use a white wave effect and noticed only the 7 base colors were available. I wrote a script to generate styles for the entire palette, but because of the size I added it in a new file, to be used optionally. 

I thought people might find it useful.


Dogfalo:
It's a bit much to add. But at least if people come looking for it they may stumble across it here.

PR #1383, opened 2015-05-19
Added cursor pointer to card action

zippoxer:
When the anchor tag doesn't have an href, there's no pointer cursor. This commit makes sure there's always a pointer cursor.


tomscholz:
@acburst Seems like a good idea to me. Your thoughts?

Issue #1400, opened 2015-05-22
<select> value with js


Sajgoniarz:
I'm struggling with it quite a while, and I can't find any solution to retrieve dropdown / select value. Closest  as I get was when I closed select inside container with id, and onClick event on it. Then I tried to retrieve value of dynamically created ul list with css path, but your lib is adding active class to previously selected element, what means I've got previous value. When you click dropdown/select item value of select is not changing, or even html in any dom element, it's only visual change, which can be caught after next selection.

Sajgoniarz:
#1401 FIXED w8 for merge

Dogfalo:
Just use  .val() on the original select

Dogfalo:
Don't see why this is necessary

Sajgoniarz:
It would be nice to see it in documentation.

Dogfalo:
It's the same as on any select

jdcauley:
Thats not true on a select you set a change handler, we don't have that.

PR #1401, opened 2015-05-22
[BUGFIX] Fixed material_select function to set clicked item as active (Issue #1400)

RakanNimer:
The _active_ class was being set on list items in a select on select _focus_ event instead of the li _click_ event. 
I adapted the logic in forms.js for this.
This code does not change anything in the UI, it is for programatically retrieving the selected list item.


Sajgoniarz:
Now, just need to w8 :) Thanks, a lot.

Dogfalo:
This probably breaks keyboard support

RakanNimer:
Hey @Dogfalo ! 
Thanks for your work on materialize it's great ! 
You were right the fix broke keyboard support when no option is selected.
Fixed that by setting the first element's class to active keeping it consistent.
Cheers !

acburst:
Should be fixed on latest release

PR #1415, opened 2015-05-24
Four undefined variables in strict mode

Lucasljj:
making it compatible with strict mode and consequently Drupal 8 :)

"doc" about strict mode: http://www.w3schools.com/js/js_strict.asp


Lucasljj:
Ops... *Three undefined variables* :)

PR #1416, opened 2015-05-24
Fix undefined variable

Lucasljj:


Dogfalo:
You should open this on https://github.com/amsul/pickadate.js

Lucasljj:
Oh, Thanks for your help :)

Issue #1417, opened 2015-05-24
two sideNav`s buggy on close


iSuslov:
If having 2 panels, on the right and one the left, you can not close sideNav by clicking outside, since you are clicking on another sideNav`s drag-target. 
Line № 3627
```
$('body').append( $('<div class="drag-target"></div>'));
``` 

should be replaced with

```
var dragTarget = $('<div class="drag-target"></div>');
$('body').append(dragTarget);
```

and then, all `$('.drag-target')` should be replaced with `dragTarget`

And finally, when clicking outside sideNav, it behaves badly since you can click to dragTarget of another sideNav if it lays 'higher'.
Workaround is: near the line 3880 insert `$('body').append(dragTarget)`, so your active drag target will be always on top.

```
$this.click(function () {
                    if (menuOut == true) {
                        menuOut = false;
                        panning = false;
                        removeMenu();
                    }
                    else {
                        if (options.edge === 'left') {
```

replace with 

```
$this.click(function () {
                    if (menuOut == true) {
                        menuOut = false;
                        panning = false;
                        removeMenu();
                    }
                    else {
                        $("body").append(dragTarget)
                        if (options.edge === 'left') {
```

imjoeybrennan:
+1

iSuslov:
Guys! I told you exactly what causes this bug. I even did a pull request, are you thinking it's good to just ignore it or you going to fix it another way? Please make something

erredeco:
this solution seems to mess the tabs component... can someone check?

PR #1418, opened 2015-05-24
bugfix #1417

iSuslov:
See https://github.com/Dogfalo/materialize/issues/1417

### DIFF
--- a/js/sideNav.js
+++ b/js/sideNav.js
@@ -19,17 +19,18 @@
         }
 
         // Add Touch Area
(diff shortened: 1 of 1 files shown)


akyunus:
I'm not sure that we really need to call  twice $('body').append(dragTarget);

akyunus:
iSuslov you are right. It was my fault that i dont understand how the code works.

Dogfalo:
Good fix, thank you @iSuslov

PR #1425, opened 2015-05-25
correct medium-and-up and large-and-up media queries

kaymmm:
I might be mistaken here, but it seems like the media query breakpoints `$large-and-up` and `$medium-and-up` are using the wrong breakpoints.

They're currently defined in _variables.scss as:
```
$medium-and-up: "only screen and (min-width : #{$small-screen-up})" !default;
$large-and-up: "only screen and (min-width : #{$medium-screen-up})" !default;
```
but I believe they should be:
```
$medium-and-up: "only screen and (min-width : #{$medium-screen-up})" !default;
$large-and-up: "only screen and (min-width : #{$large-screen-up})" !default;
```

### DIFF
--- a/sass/components/_variables.scss
+++ b/sass/components/_variables.scss
@@ -82,8 +82,8 @@ $small-screen: 600px !default;
 $medium-screen: 992px !default;
 $large-screen: 1200px !default;
 
-$medium-and-up: "only screen and (min-width : #{$small-screen-up})" !default;
-$large-and-up: "only screen and (min-width : #{$medium-screen-up})" !default;
+$medium-and-up: "only screen and (min-width : #{$medium-screen-up})" !default;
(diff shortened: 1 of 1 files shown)


Dogfalo:
It is right as is, your code would make medium and up non-inclusive of itself

kaymmm:
Sorry, I had misinterpreted the breakpoints as minimums, not maximums (e.g., medium = 601-992, not 992-1200). In this case, would it make sense to add another media query variable for extra-large screens:
```
$extra-large-and-up: "only screen and (min-width : #{$large-screen-up})" !default;
```
since currently there is no media query defined that uses the $large-screen-up breakpoint? The main use case is limiting the size of things on the screen so that they don't get super wide and short. Thanks!

PR #1437, opened 2015-05-28
fixed dropdowns misbehaving because of relocating dom

kdar:
The issue is dropdowns are relocated in the dom to be right next to the element that is triggering the dropdown. If the surrounding dom has CSS elements that negatively affect the dropdown (like position: relative, or overflow: hidden), then the dropdown appears strange.

This fix gets the offset of the origin in relation to the entire page, and then it absolutely positions the dropdown from that.

### DIFF
--- a/js/dropdown.js
+++ b/js/dropdown.js
@@ -41,8 +41,6 @@
 
     updateOptions();
 
-    // Attach dropdown to its activator
-    origin.after(activates);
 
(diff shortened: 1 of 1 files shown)


acburst:
We attach it to the trigger for a reason. Doing this breaks a lot of cases for dropdowns.

PR #1461, opened 2015-05-31
Add scrollspy options

jmversteeg:
Added the option of `scrollOffset` to the `scrollSpy` method. This allows you to define the offset to use when automatically scrolling to the designated part of the page (instead of the hardcoded 200 offset, which is still used as a default)


js/scrollspy.js
```diff
@@ -181,14 +186,12 @@
...
-			
-	    	$('html, body').animate({ scrollTop: offset - 200 }, {duration: 400, queue: false, easing: 'easeOutCubic'});
+
+              var scrollOffset = typeof options.scrollOffset !== 'undefined' ? options.scrollOffset : 200;
+              $('html, body').animate({ scrollTop: offset - scrollOffset }, {duration: 400, queue: false, easing: 'easeOutCubic'});
 			
 		  });
 		});
-		options = options || {
-			throttle: 100
```
jmversteeg: no need to specify this default value here, since later on there is a fallback (`var throttledScroll = throttle(onScroll, options.throttle || 100);`) either way. a default value of `{}` will suffice

PR #1467, opened 2015-06-02
Fix html preview

ahi:


Dogfalo:
You didn't modify the source files so we couldn't merge it. But i made the fix myself. 7d0a1336f7da
Thanks!

PR #1482, opened 2015-06-04
Fix bug in Firefox with hover and below origin

elboman:
Added browser compatibility for target element on mouseleave that was causing a bug in Firefox when using dropdown with hover and below origin options.


Dogfalo:
Great work, thanks!

PR #1496, opened 2015-06-09
Card hoverable

kaliel86:


Dogfalo:
Great idea! I added it with a slightly different implementation, 6058925a185

PR #1508, opened 2015-06-11
Fix #1503 - Textarea not resizing automatically with dynamic content

### LINKED ISSUE #1503: Textarea not resizing with dynamic content
When filling up the textarea dynamically, the textarea is not resized automatically.

Even when the textarea gains focus, it is not resized.

It would be great if the textarea was resized on focus and not only on keyup keydown.

klugjo:
Added a focus event on textarea resize.

https://github.com/Dogfalo/materialize/issues/1503

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -142,7 +142,7 @@
       }
     });
 
-    $('body').on('keyup keydown', text_area_selector, function () {
+    $('body').on('keyup keydown focus', text_area_selector, function () {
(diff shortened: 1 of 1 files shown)


tomscholz:
Should be fixed in https://github.com/Dogfalo/materialize/commit/4a3c15c9fcd1bdd778f2840e932700254522fda7

newoutlaw98:
This is still not fixed in v0.100.1.

PR #1510, opened 2015-06-11
Fix for Issue 1509

FrancisStokes:
Changing the ease function from easeOutSine to easeInSine fixes firefox's flickering problem without any real perceptible difference in the look of the dropdown.


Dogfalo:
This was fixed

PR #1524, opened 2015-06-14
Fix sideNav Variations ID

Johann-S:
When you click on the item Variations of the sideNav of this page : http://materializecss.com/side-nav.html
You cannot go to the section mentionned

Tested on : Chrome 43.0.2357.124 m and Firefox 38.0.5


Dogfalo:
I uncapitalized the id instead in  8375acd93a5d62

Johann-S:
Yes but it doesn't work currently. Why did you uncapatilize ?

PR #1528, opened 2015-06-15
Update ghpages-materialize.css

john-dev:
Removed overflow:hidden from .card class since selects (formItem) are beeing cut when higher than .card


acburst:
This is fixed on master. We will try to release soon. For now just override it manually

john-dev:
Allright! Thanks

PR #1536, opened 2015-06-17
Add border-radius mixin into prefixer.scss

RaidoS:
Add mixin border-radius, that exist in the docs http://materializecss.com/sass.html , but missed in the _prefixer.scss

### DIFF
--- a/sass/components/_prefixer.scss
+++ b/sass/components/_prefixer.scss
@@ -156,6 +156,13 @@
     @include box-sizing(content-box);
 }
 
+// Border Radius
+
+@mixin border-radius($args) {
+    -webkit-border-radius: $args;
+    -moz-border-radius: $args;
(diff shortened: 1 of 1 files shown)


gnicol-bzh:
*border-radius* do not need any prefixe on recent browsers (see [Can I Use](http://caniuse.com/#feat=border-radius)). 

As far as I'm concerned, I don't use prefixer mixins anymore since I've discover [Autoprefixer](https://www.npmjs.com/package/autoprefixer) ; available for Gulp ou Grunt. It's really worth a try ;)

RaidoS:
@gnicol-bzh , You right. 
So, in this case should remove border-radius mixin from the docs, and other similars.

Dogfalo:
It is removed and will be update when the release comes

PR #1541, opened 2015-06-17
Updated tooltip styling

kbouw:
Updated tooltip styling to match google material design guidelines.

You can view the desktop style guideline here: http://www.google.com/design/spec/components/tooltips.html#tooltips-usage


tomscholz:
I like the idea, but you edited the wrong files. If you still have intrest in this, open a new PR that follows our CONTRIBUTING guidelines

Issue #1386, opened 2015-05-20
tooltop.js problem when another tooltip is initialize


itsmechlark:
This happen when another tooltip is setup after the previous tooltip is setup.

Problem at [tooltip.js#L16](https://github.com/Dogfalo/materialize/blob/ce5e0ba9f0ac2e32e0f2d9f8859928af32c6b0dd/js/tooltip.js#L16)

The previous container of the tooltip is remove after the second tooltip is setup.

``` html
<a id="tp1" class="btn tooltipped col s4 offset-s4 l2 offset-l1" data-position="bottom" data-delay="50" data-tooltip="I'm TP1!">TP1</a>
<a id="tp2" class="btn tooltipped col s4 offset-s4 l2 offset-l1" data-position="bottom" data-delay="50" data-tooltip="Goodbye TP1!">TP2</a>
<script>
$(document).on("click", "#tp1", function(){
    if (!$("#tp2").is(".ok")){
        $("#tp2").addClass("ok").tooltip();
    };
});
$("#tp1").tooltip();
</script>
```

Check this https://jsfiddle.net/8wxa2pym/

itsmechlark:
PING

toben:
I had the same issue. You have to remove the line :

Fed03:
Very serious bug, @Dogfalo

webbiesdk:
@toben, you cant just remove that line, then we risk having a lot of .material-tooltip divs floating around, that are never used. 

My pull-request should however fix this problem in a good way.

carlosperate:
Has this issue been resolved since the merge of https://github.com/Dogfalo/materialize/pull/1547 ?

webbiesdk:
I think it has, now the API is such that previous tooltips are only ever removed if you call tooltip like: `.tooltip("remove");`. 

So yeah, IMO it has been fixed.

itsmechlark:
#1547, I think doing this every tooltip initialisation is not right it should be like [bootstrap/tooltip.js](https://github.com/twbs/bootstrap/blob/master/js/tooltip.js). We all know that `tooltip.js` is `materializecss` is just the same but differ on how it animate and looks. 

All js library should be structured well, just a suggestion.

merlano17:
There is simple solution for this for the time being till they fix it
I use this 

```
setInterval(clearTooltip, 6000);
function clearTooltip() {
    $('.material-tooltip').hide();
  }

[...]

PR #1547, opened 2015-06-19
Not removing existing tooltips when we re-initialize. 

webbiesdk:
Currently, when you initialize a tooltip, all the previously initialized tooltips are removed. 

So when you add a tooltip dynamically, like we do here: krescruz/angular-materialize#34, you have to initialize all the tooltips again. 
This is also dicussed a bit here: #1386

With this change, a tooltip is only removed if its anchor is detached, or we are about to re-initialize it. 

The last part keeps backwards-compatibility with people that re-initialize everything when a new tooltip is added.

### DIFF
--- a/js/tooltip.js
+++ b/js/tooltip.js
@@ -13,7 +13,16 @@
       options = $.extend(defaults, options);
 
       //Remove previously created html
-      $('.material-tooltip').remove();
+      var that = this;
+      $('.material-tooltip').each(function (index, element) {
+        var origin = $(element).data("origin");
+        // Checking if the element is still attached to the dom. http://stackoverflow.com/questions/3086068/how-do-i-check-whether-a-jquery-element-is-in-the-dom#answer-3086084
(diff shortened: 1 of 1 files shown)


Fed03:
+1 @Dogfalo

Dogfalo:
Thanks for improving the logic.

Floby:
on which version is this available ?

carlosperate:
Not yet released, if you need to use them right away you'll have to import the files from https://github.com/Dogfalo/materialize/tree/master/bin

Floby:
cannot this be release as a bugfix with version 0.97.1 ?

webbiesdk:
The logic was changed again in 30b2e1d525c5c4a721c9c84f2e33524831331e9a. So if you download the latest version, you have to explicitly remove all the "old" tooltips yourself. 

Also +1 for release, I could use it for krescruz/angular-materialize#34.

Floby:
I copied the fixed version to my project. I hate this.

PR #1550, opened 2015-06-19
Fixed white space inside card

onerinas:
cards materialize
similar issue is seen in this SO : http://stackoverflow.com/questions/4042137/unknown-whitespace-appearing-inside-div-without-padding-or-margins?answertab=votes#tab-top


onerinas:
@Dogfalo any thought on this?

PR #1564, opened 2015-06-23
fix(jQuery): add missing jQuery wrapper on global.js

ben-barbier:
I propose to add the jQuery wrapper on global.js because if it's not present, Jquery with no $ alias doesn't work fine.
capture_d ecran_2015-06-23_a_02_15_32

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -28,9 +28,11 @@ Materialize.elementOrParentIsFixed = function(element) {
 
 // Velocity has conflicts when loaded with jQuery, this will check for it
 var Vel;
-if ($) {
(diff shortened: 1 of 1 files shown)


Lucasljj:
A more elegant way in [#1565](https://github.com/Dogfalo/materialize/pull/1565)

ben-barbier:
Sorry but #1565 doesn't work (detail on #1565 comments)

ben-barbier:
Problem fixed more elegantly on https://github.com/Dogfalo/materialize/pull/1595.

PR #1565, opened 2015-06-23
Use JavaScript Module pattern on global.js

Lucasljj:
Other parts of the code in this file also use jquery, let's add a wrap and ensure that the code to run properly

This solves the problem in a more elegant way that https://github.com/Dogfalo/materialize/pull/1564

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -1,36 +1,32 @@
-Materialize = {};
+(function ($) {
+  Materialize = {};
 
-// Unique ID
-Materialize.guid = (function() {
-  function s4() {
-    return Math.floor((1 + Math.random()) * 0x10000)
-      .toString(16)
-      .substring(1);
-  }
-  return function() {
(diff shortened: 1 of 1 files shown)


ben-barbier:
Hello !

I also had this approach at first time but there is a problem with your solution...
The 'Vel' variable is not usable outside of global.js file.
It's the reason of my solution ('Vel' definition outside of the jQuery wrapper).

Lucasljj:
I checked all .js files in materialize and saw that only the [toasts.js](https://github.com/Dogfalo/materialize/blob/master/js/toasts.js) file does not depend on jquery

I'll think of another way to do it :)

ben-barbier:
I agree... and I'll find it ;-)

ben-barbier:
Hi !

What do you think about https://github.com/Dogfalo/materialize/pull/1595 ?

PR #1569, opened 2015-06-23
Added caret deletion on material_select destroy and redraw

cuckoopt:
I was trying your latest release and upon close inspection I found out that the select's on destroy and redraw were leaving behind the caret, this line is just to remove it.


acburst:
Oops, I resolved this issue without looking at this. Sorry about that

Fixed in cbcf4c8f6c229f42dba1dd3ccde1788d7f96ff80

PR #1573, opened 2015-06-23
Remove css specificity in grid classes such as .col and .s1

MCooke:
I haven't found a reason for the specificity of .col and column sizes. I needed to remove padding on some columns.


Dogfalo:
In this case the higher specificity can be good as it allows the grid system to have a higher "priority" over other widths that may be being applied. This way the grid system will not be broken as easily by other rules.

PR #1579, opened 2015-06-24
Update buttons.html

coreygo:
Line 159 had an extra closing icon match.


acburst:
I will need to update our templates, but thanks for pointing this out.

Fixed in 3569dbb

PR #1587, opened 2015-06-24
Un-initialized variable causing issues

aequasi:
Un-initialized Materialize variable causing issues when using browserify. Assumign due to strict mode?

This should fix the one issue i've seen

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -1,4 +1,4 @@
-Materialize = {};
+window.Materialize = {};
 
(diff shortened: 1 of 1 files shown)


abhijitghogre:
Even afer fixing that I get:

```
Uncaught TypeError: Cannot set property 'Picker' of undefined
Uncaught ReferenceError: validate_field is not defined
```

PR #1590, opened 2015-06-25
Fixed duplicate leanModal overlays

JonathanGuo:
The bug is by opening opened modals multiple times, it will generate
duplicate overlays. These overlays will not be removed properly.

Fixed:
Updated openModal, added _stackModals array to store references of each
modal. If a modal is opened then it will retrieve overlayID.
 Updated closeModal, it removes references of modals when close.

I created a plunk to illustrator bug.
http://embed.plnkr.co/NMosr6KBPlR1NTyX3FLj/
(Somehow Codepen doesn’t run it properly)


JonathanGuo:
Merged latest master branch. I have tested the code with the latest version 0.97.1.

romant:
any insight to when/**if**  this will be merged?

PR #1595, opened 2015-06-25
fix(Velocity): use jQuery wrapper instead of incorrect global velocity definition (Vel).

ben-barbier:
The previous global velocity definition do not treat the case of jQuery has no $ alias (on global.js).

This PR fix the same problem as https://github.com/Dogfalo/materialize/pull/1564 and https://github.com/Dogfalo/materialize/pull/1565 but more elegantly (I'll close my previous solution https://github.com/Dogfalo/materialize/pull/1564).

Tkx to @Lucasljj for his feedback (on https://github.com/Dogfalo/materialize/pull/1565).

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -25,12 +25,3 @@ Materialize.elementOrParentIsFixed = function(element) {
     });
     return isFixed;
 };
-
-// Velocity has conflicts when loaded with jQuery, this will check for it
-var Vel;
-if ($) {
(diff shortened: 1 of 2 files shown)


ben-barbier:
**Pull-Request content :**

- global.js
 - remove Vel definition

- toasts.js
 - add jQuery Wrapper
 - replace Vel by $.velocity

PR #1597, opened 2015-06-26
Allow jquery 1.x to be used

truenorth:
The existing version is over-specific. jquery `>= 1.11.x` (`1.11.1` for example) works just fine

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -61,6 +61,6 @@
 
   ],
   "dependencies": {
-        "jquery": ">=2.1.1"
+        "jquery": ">=1.11.1"
     }
 }


hebbet:
did you try materializecss with jquery 1.11?

i got some errors and some features didn't work.

Dogfalo:
What is the reason people need to use the old jquery version? The framework is only tested on the jQuery 2 versions although it may work on older versions.

hebbet:
IE8 support.

Johann-S:
Because they want a better browser compatibility.

I think it would be useful to have a JSBin using JQuery 1.11 to show errors if they are.

PR #1603, opened 2015-06-26
Clarify icon usage.

Garbee:
This documentation update clarifies how to use the Material Design icons. Some [confusion exists](https://plus.google.com/u/0/115994601702879969997/posts/6XcHKNyrR13) (see comments) based on the current wording which does not work as it says.

### DIFF
--- a/jade/page-contents/icons_content.html
+++ b/jade/page-contents/icons_content.html
@@ -8,7 +8,7 @@
             <p>To be able to use these icons, you must include this line in the <code class="language-markup">&lt;head></code>portion of your HTML code</p>
             <pre><code class="language-markup">
   &lt;link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet"></code></pre>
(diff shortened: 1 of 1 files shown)


cbruyndoncx:
the mdi-action... style of icons from v.96 give now unwanted results when upgrading to version 0.97
I understand from somewhere else that the hex codes changed.
should these icons be dropped/replaced with other references - or will you keep mdi-xxx style icons ?

Garbee:
Looking at the source the `mdi-` icons are still in place at the moment. I presume to make for easy upgrading during pre-release.

As far as if they will remain when 1.0 hits or whatever the future plans are, I have no idea myself (since I don't use or work on this project.) Someone else will need to answer that/update the docs accordingly.

cbruyndoncx:
for the moment, i just used <i class="material-icons">hexcode</i> for the few icons that i use. Not the most elegant, but the safest imo

PR #1605, opened 2015-06-27
Fixing misleading details about tab selection api

lelandbatey:
Currently, [documentation on tabs](http://materializecss.com/tabs.html) says:

> You can programmatically trigger a tab change with our select_tab method. You have to input the ***id*** of the tab you want to switch to.   

Specifically:

> input the ***id*** of the tab

This is very false. [As you can see by the source here, selection of tabs has nothing to do with their id, and is instead entirely based on the contents of the `href` attribute of the element.](https://github.com/Dogfalo/materialize/blob/1f27f1162d0187c905bbb883bff15cf1e5e16a8f/js/tabs.js#L111)

I've updated the documentation here to be correct.

### DIFF
--- a/jade/page-contents/icons_content.html
+++ b/jade/page-contents/icons_content.html
@@ -8,7 +8,7 @@
             <p>To be able to use these icons, you must include this line in the <code class="language-markup">&lt;head></code>portion of your HTML code</p>
             <pre><code class="language-markup">
   &lt;link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet"></code></pre>
-            <p>To use these icons, just place the name of the icon into the class of an i tag</p>
+            <p>To use these icons, use the material-icons class on an element and provide the ligature as the text content.</p>
             <pre><code class="language-markup">
       &lt;i class="material-icons">add&lt;/i>
             </code></pre>
--- a/jade/page-contents/tabs_content.html
+++ b/jade/page-contents/tabs_content.html
@@ -55,7 +55,7 @@
 
       <div id="method" class="section scrollspy">
         <h4>jQuery Plugin Methods</h4>
-        <p>You can programmatically trigger a tab change with our <code class="language-javascript">select_tab</code> method. You have to input the id of the tab you want to switch to. In the case of our demo it would be either test1, test2, test3, or test4.</p>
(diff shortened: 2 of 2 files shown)


Dogfalo:
See here

There are id's and they will also be the same as the href's, they are essentially the same thing

lelandbatey:
@Dogfalo In the example code in the tutorials/in your screenshot, it is true that the `id`'s and the the `href`s are the same. Since they are the same, attempting to select based on the `id` will *appear* to work.

However, if the example html where changed so the `id`s are different from the `href`s and you attempt to select a tab now based on the `id`, the js will not work and the selection will not change.

So if the html where changed to the following:

```html
<div class="row">
	<div class="col s12">
		<ul class="tabs">
			<li class="tab col s3"><a href="#test1">Test 1</a></li>
			<li class="tab col s3"><a class="active" href="#test2">Test 2</a></li>
			<li class="tab col s3 disabled"><a href="#test3">Disabled Tab</a></li>
			<li class="tab col s3"><a href="#test4">Test 4</a></li>
		</ul>
	</div>
	<div id="testone" class="col s12">Test 1</div>
	<div id="testtwo" class="col s12">Test 2</div>
	<div id="testthree" class="col s12">Test 3</div>
	<div id="testfour" class="col s12">Test 4</div>
</div>
```

Given the above html, attempting to change the selected tab like so:

```javascript
$('ul.tabs').tabs('select_tab', 'testone');
````

This javascript will do nothing and no new tab will be selected.

carlosperate:
I'm not quite sure your point is completely valid, as the example that you give breaks the tabs implementation altogether: http://codepen.io/anon/pen/pJxaKa
For it to work it needs to be the same code as the documentation: http://codepen.io/anon/pen/NqOyzg
In which case the id and href are essentially the same thing as indicated by @Dogfalo

lelandbatey:
@carlosperate @Dogfalo  The context that I brought this issue up in was a project that made use of the tabs but without taking advantage of the built in ability to hide and show elements based on the currently selected tabs.

Revisiting this, I see that the way the mentioned project was using these tabs was not how they where intended to be used by the Materialize developers. I see now why the documentation for Materialize is the way it is. Thank you both for your comments!

Issue #1426, opened 2015-05-26
fixed sidenav with closeOnClick enabled


rgoomar:
So, currently, if you have a fixed sideNav and you want `closeOnClick` "enabled" for your mobile side, you have to add in a work around to make sure that if a user clicks and is on a full desktop version, it doesn't hide the side nav when it should be fixed.

TLDR; You can't have `closeOnClick` enabled with a fixed side nav

A workaround I created for this is:

``` javascript
  // Side Navigation fix
  $('.side-nav li a').on('click', function(e) {
    windowsize = $(window).width();
    if (windowsize < 992) {
      $('.button-collapse').sideNav('hide');
    }
  });
```

This is probably something to look into since it is useful to have the sidenav close on click on mobile, but stay fixed on desktop screens.

JulianKingman:
Thanks for the fix, this is kind of annoying.

rgoomar:
@Dogfalo any update on this?

lijoantony:
@rgoomar Thanks for the work around. However, it didn't fully work for me.
Side menu is getting hidden, but #sidenav-overlay remains.
I had to do something like this,

```javascript
var windowsize = $(window).width();
if (windowsize < 992) {
     $('.button-collapse').sideNav('hide');
     setTimeout(function(){
         $('.button-collapse').sideNav('hide');
     }, 500);
}
```
Second hide is to hide #sidenav-overlay.
Even `setTimeout(..., 0);` didn't help there.

ronniebermejo:
#### angularized fix
```
(function () {
  'use strict';

  angular
    .module('buggy')
    .directive('sideNavBar', sideNavBar);

  sideNavBar.$inject = [];
  function sideNavBar() {
    return {
      restrict: 'C',
      require: '?ngModel',
      link: function (scope, elem, attrs, ctrl) {
        $(".button-collapse").sideNav();


        var SIDEBAR_BREAK_POINT = 992;
        var windowsize = $(window).width();
        if (windowsize < SIDEBAR_BREAK_POINT) {
          $('.button-collapse').sideNav('hide');
          setTimeout(function(){
            $('.button-collapse').sideNav('hide');
          }, 500);
        }
      }
    }
  };
})();
```

jacek213:
this is a major issue, can we expect a fix?

MuyBien:
+1

Almaju:
+1

belgacea:
+1

asonj:
[...]

PR #1615, opened 2015-06-27
Character count and Sidenav issues fixed

### LINKED ISSUE #1612: Character Count works with invalid html attribute
The character counter works with an invalid html attribute (length) instead of using maxlength or a data-attribute.

I couldn't find the length anywhere in the specs (http://www.w3.org/TR/html5/forms.html#the-input-element). I'm using React which eliminates all attributes that are not in the specs.

In case you guys would like to get this solved, I'm more than willing to transform it towards whichever option (maxlength or data-) has your preference.

janpieterz:
Fixed #1612 and #1426.

With #1612 I chose for using a data attribute, since the maxlength attribute would prohibit the user from getting the validation errors and feels a bit too restrictive (it would never show the red invalid line). 

In case you want me to make any other changes to this just let me know!

### DIFF
--- a/js/character_counter.js
+++ b/js/character_counter.js
@@ -26,6 +26,10 @@
   }
 
   function addCounterElement($input){
+    if($input.parent().find(".character-counter").length > 0){
+		return;
+	}	
+	  
     var $counterElement = $('<span/>')
                         .addClass('character-counter')
                         .css('float','right')


mondaychen:
+1 for this. I hate the fact that a fixed SideNav for desktop will also be closed when `closeOnClick` enabled.
But probably a separate option is better? Some people may still need it on desktop version

janpieterz:
@mondaychen ideas for naming? preventCloseOnClickOnDesktop: true ?
Seems a little bit too extensive, though it describes exactly what happens ;)

janpieterz:
I've also added a small css fix. A pagination element was only clickable on it's link element, while the whole li had a pointer.

mondaychen:
How about turn the option `closeOnClick` into [Boolean or Function], instead of just Boolean?
Like this:
````
options.closeOnClick = $.isFunction(options.closeOnClick) ?  options.closeOnClick() : options.closeOnClick
````
In this way it keeps the current behavior as it is. And when you need to prevent it on desktop version, you can use
````
$('.button-collapse').sideNav({
    closeOnClick: function() { return window.innerWidth <= 992; }
});
````

janpieterz:
@mondaychen I'll wait for the official dev's to pitch in so I'm not working on it too much without any guidance on how they'd like it. Thanks for the idea, would never have thought of doing it that way myself, neat way!

The other commit just added ensures the ability to call .characterCounter() multiple times over the same form without creating two elements. I came onto this in my react app which updated the view dynamically, calling the method then would result in a double element being created since the plugin already initializes itself.

Other solutions I currently see are:
Doing it at the top of the .each function, making sure we don't double-monitor and execute the character counting (cleaner but wasn't sure if this was wanted)
Removing the initial call from the library which then forces every single user to call it manually.

tomscholz:
#1426 fixed in https://github.com/Dogfalo/materialize/commit/f97bf0b5e4061b10c5e6a9624de97f138583c003

tomscholz:
I'm going to close this pull request now because it's very old and some things are already fixed in the master branch. Any new discussion should probably happen on a new issue with a more specific title!

PR #1617, opened 2015-06-28
.collapsible-header height changed to auto

gregaubs:
Causing an issue for the collapsible component (especially on mobile devices).

Text was spilling over into the row below. 
Rows weren't able to flexibly change height to accommodate the text:
image

Removing the 'height:3rem' means it can accommodate the text as the viewport shrinks:
image

Thanks!


acburst:
Thanks for the solution. But in your pull request you only changed the compiled css and not the Sass. I did it myself and merged it in here 20d68a7.

Next time if you could edit the Sass and then compile the css, I can automatically merge through github.

PR #1625, opened 2015-06-29
Fix documentation typos on getting_started_content.html

billfienberg:
In jade/getting_started/getting_started_content.html, I updated line 27 from '...all you have to is...' to '...all you have to do is...,' and capitalized the word 'this' in lines 54 and 62.


acburst:
thanks

PR #1628, opened 2015-06-29
Tooltip javascript-defined position

Roundaround:
Added default tooltip position and linked position to look at the javascript parameters when not defined in the attribute.

The general reason for this change is to reduce unnecessary bloat in the DOM when trying to use the Materialize tooltips.  Having to specify a class and at least 2 data-attributes makes an originally small button definition to need line wrapping in some IDEs and code styles.  This change makes the data-position attribute optional in the same fashion as the data-delay attribute to possibly reduce this clutter.

### DIFF
--- a/js/tooltip.js
+++ b/js/tooltip.js
@@ -8,7 +8,8 @@
 
       // Defaults
       var defaults = {
-        delay: 350
+        delay: 350,
+        position: 'bottom'
       };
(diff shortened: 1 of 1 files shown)


tan9:
:+1: for this PR!

Roundaround:
Remade a mergable version of this pull request at #3004

PR #1630, opened 2015-06-29
[Fix #1621] Update the current displayed value during slide of range input

### LINKED ISSUE #1621: Range Slider (Forms) doesn't update the current value during slide
The range slider doesn't update the current value of range bar during slide. The value in [.thumb] is updated only when the (mouseup/touchend) event is triggered.

The current value in [.thumb] must be updated as the user slides through the bar. This would help to select accurate value easily. In fact, it's a very basic feature that must be provided to user.

The current issue can be seen on codepen at http://codepen.io/anon/pen/QbaqVK

The solved issue can be seen on codepen at http://codepen.io/anon/pen/GJyJGK

Johann-S:
Fix https://github.com/Dogfalo/materialize/issues/1621

Thanks to @yatharthk.

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -243,9 +243,8 @@
           left = width;
         }
         thumb.addClass('active').css('left', left);
-
+        thumb.find('.value').html(thumb.siblings(range_type).val());
       }
-
     });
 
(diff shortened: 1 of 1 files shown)


acburst:
Nicely done, thanks

PR #1635, opened 2015-06-30
Using correct mixin to init flexbox for .tabs

marvwhere:
@include flex() was incorrect because we would need to pass values, instead doing this, use mixin flexbox


marvwhere:
forgot to add the original error: 

>Module build failed: 
  >@include flex($values);
                              ^
      Undefined variable: "$values".

PR #1644, opened 2015-07-01
Remove list style from Collection Header by default.

MichaelObi:
Set list-style for collection-header to none to remove bullet.


tomscholz:
Why did you commited no changes to the .travis.yml file?

PR #1654, opened 2015-07-02
Added support for file input with 'multiple' property

hotzenklotz:
I added support for file uploads with the 'multiple' property set. `<input type='file' multiple>`.

Previously file uploads would only show the name of the first selected file. Now it shows the file name of all selected files separated by comma.

Working demo:
http://codepen.io/therold/pen/WvMZZN


acburst:
Looks great, Thanks!

Issue #1618, opened 2015-06-28
problem adding dynamic select lists


jots:
I am cloning a table row that contains select lists.  after I append it, the select list is disabled.
if I add: $('select').material_select();  after appending it, it creates a duplicate select list which works, but leaves the select list that doesn't work.

uncomment out the $('select').material_select(); in:
http://jsbin.com/hutuzagaqu/edit?html,js,output

to see what I mean.

rexopl:
Hello, yeah there are some problems with selects. I even do material_select('destroy') and then create new one material_select() but it seems that after destroy some code left: "<span class="caret">▼</span>". 
After few destroy, create i have many carets above my select.

smark91:
same here.
Dropdown not dynamically create on a page. when material_select() a ▼ appears for every time the function is call.

Stroncickiy:
i have same problem

kai23:
+1

smark91:
find a fix.
already pull request.

https://github.com/Dogfalo/materialize/pull/1656

carlosperate:
If this issue is only regarding the left over caret, it has already been fixed. You can get the latest built version from the bin folder.
Original issue: https://github.com/Dogfalo/materialize/issues/1570
Commit: https://github.com/Dogfalo/materialize/commit/cbcf4c8f6c229f42dba1dd3ccde1788d7f96ff80

acburst:
The caret bug has been fixed already. I think the problem lies with you cloning the select after you've initialized it with our plugin. Initializing it with our plugin is what creates the structure so if you clone an already initialized select it will duplicate the structure. Try instead of cloning it, creating it from scratch or some different way of cloning .

rexopl:
The problem still occurs on master version.

sheki:
I think this exists and is a big issue for react.

PR #1656, opened 2015-07-02
bugfix  issue #1618

smark91:
in the last update 0.96.1 -> 0.97 you change the `dropdownIcon` from `<i>` to `<span>` but forget to change the `$select.parent().find('i').remove();` on `$select.parent().find('span').remove();`

### DIFF
--- a/dist/js/materialize.js
+++ b/dist/js/materialize.js
@@ -2725,7 +2725,7 @@ $(document).ready(function(){
       // Tear down structure if Select needs to be rebuilt
       var lastID = $select.data('select-id');
       if (lastID) {
-        $select.parent().find('i').remove();
+        $select.parent().find('span').remove();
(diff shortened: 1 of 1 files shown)


carlosperate:
This issue has been already been fixed in https://github.com/Dogfalo/materialize/commit/cbcf4c8f6c229f42dba1dd3ccde1788d7f96ff80.
The javascript in the dist folder is built from the individual files in the `js` folder, so for future patches please ensure you are editing the original source code files.

rkusuma:
@Dogfalo when will this bug fix released ? maybe in 0.97.1

carlosperate:
Is already fixed, so it will be packed in the next release. You can get the latest files from the bin folder in this repo if you'd rather not wait.

rkusuma:
thanks @carlosperate

PR #1671, opened 2015-07-04
Fix dead link in showcase, and add thumbnail to images folder.

carlosperate:
There is also a couple of dead pages, but based on responses to previous pull requests it seems that you prefer to leave them.


carlosperate:
The showcase page has been updated, but this link is still dead. I could update the pull request, but will you guys review it or just let it sit here?

Dogfalo:
I looked through all the links and made the decision to leave a few dead
ones temporarily. The rest of the dead ones will be removed at a later date

carlosperate:
But this pull request was actually changing a dead link to a working link...

macsupport:
Leaving links when you know they are dead for over 11 days?!! That makes no sense on an active page. Do you just not care or don't have the time or what?

PR #1685, opened 2015-07-07
Update materialize.js

Cerealkillerway:
Updated tooltip function to autoremove tooltip's elements when their respective tooltipped are removed from the DOM


carlosperate:
I believe you should probably change the https://github.com/Dogfalo/materialize/blob/master/js/tooltip.js instead of the built js file. Also, it has been a while since it was last built, there might be discrepancies with the current code base.

Issue #1686, opened 2015-07-07
"$('select').material_select();" - Issue


fcgomes92:
Hi. When I use the select box dynamically in a form, in my case it is a simple gender select, whenever I create a new form and call '$('select').material_select();' another select box appears.
I thought it was related to the fact that I'm using each dynamic form inside a collapsible element, but when I changed the class to 'browser-default' everything worked fine.
I took a look around but didn't find anything like this, if it's an old issue or something like that I apologize xD

Here is what it looks like when I add a second form. The upper select box doesn't work, just the bottom one.
![screenshot from 2015-07-07 01-34-13](https://cloud.githubusercontent.com/assets/6561278/8538790/7422b3e4-2448-11e5-951c-35cb24148dab.png)

Thanks (:

carlosperate:
This was an issue resolved a long time ago (https://github.com/Dogfalo/materialize/issues/819, https://github.com/Dogfalo/materialize/issues/452), are you sure you are using the latest version?

Based on the extra caret added at the top, I would assume you are using 0.97.0 as that was an issue on that version (https://github.com/Dogfalo/materialize/issues/1570), so it should be working. If you are changing the select forms, are you sure you are destroying the old one first? 

`$('select').material_select('destroy');`

Otherwise please crate a codepen with an example of your issue.

ahmadfikrizaman:
Encounter the same problem on 0.97.0..

```javascript
            $('select').material_select('destroy');

            $.each(content, function(index, value)
            {
                $("#" + index).val(value);
            });

            $('select').material_select();
```

ahmadfikrizaman:
owh.. sorry @carlosperate .. just saw the #1570

Issue #1638, opened 2015-06-30
material_select() options ?


razy69:
Hi, is there any options like dropdown for material_select ? 

``` js
$('select').material_select({
    maxHeight: 200px,
    belowOrigin: true
}); 
```

Johann-S:
+1 for this two features

TyrionGraphiste:
Hi, I included this in my PR.

fega:
Closed due inactivity, feel free to reopen it if it is still necessary.

Issue #1570, opened 2015-06-23
Material Select remaining caret icon on destroy


avcreation:
When using the $('select').material_select('destroy') command, the select correctly disapear, but letting the caret icon in parent DOM element.

Here is a fiddle for example : 

https://jsfiddle.net/4nbq2ymv/

Thx,

philipraets:
In addition:
Updating adds extra caret icons to the select

http://codepen.io/anon/pen/XbzMva

carlosperate:
Can confirm I am experiencing the same issue.

ricardodovalle:
:+1:  I am getting the same issue:
http://pastebin.com/zyS0yjTx

Cerealkillerway:
having the same issue

acburst:
Fixed in cbcf4c8

rkusuma:
@acburst is it not yet fixed in latest release 0.97.0.

uok:
if you click on https://github.com/Dogfalo/materialize/commit/cbcf4c8f6c229f42dba1dd3ccde1788d7f96ff80 you can see at the top left "master" which means that it is commited, but not released. if you see a number there in the future, e.g. 0.97.1, it's released

meichenblatt:
When will it be released?

OminStyle:
I'm experiencing the same problem and waiting for the release.

SomilKumar:
I too experienced the same issue , waiting for the release ...

OminStyle:
You can add $select.parent().find('input, i, .caret').remove(); on re-initialization to fix this issue until the next release.

serkandurusoy:
I just tried a fresh checkout of the master and the problem still seems to be there. No change of the extra caret behaviour.

I'm using meteor, if that matters.

lasseborly:
Has been fixed in v0.97.1 @serkandurusoy.

PR #1687, opened 2015-07-07
Refactor select plugin and add options

Johann-S:
Hi,

I know it's a big PR but i think it's valuable.

This PR allow to add a maxHeight to the select plugin (https://github.com/Dogfalo/materialize/issues/1638) and maybe other options later.

But it fixe too the problem of destroying a select (related to https://github.com/Dogfalo/materialize/issues/1686 and https://github.com/Dogfalo/materialize/issues/1570).

And to finish I think it's better for the code to split each plugins in different file this is why
I create a select.js file.


Johann-S:
PING @Dogfalo @acburst 
Any feedbacks on this PR ?

PR #1691, opened 2015-07-08
Reuse the lean overlay when opening an already opened modal

### LINKED ISSUE #1647: Open modal multiple times and closing it leaves behind the dark overlay permanently
So if a modal is opened multiple times, and then exited only the latest overlay is removed and all other stay there permanently.

To test you can go to http://materializecss.com/modals.html and in the javascript console just run `$('#modal1').openModal();` multiple times. You can see that after clicking any of the modal options only one overlay is removed.

pascal-de-ladurantaye:
fixes #1647

### DIFF
--- a/js/leanModal.js
+++ b/js/leanModal.js
@@ -24,11 +24,11 @@
       
       if ($modal.css("display") == "block") {
         // The modal is already opened, just put it back on top
-        associatedOverlayID = $modal.data('overlay-id'),
+        associatedOverlayID = $modal.data('overlay-id');
         $associatedOverlay = $('#' + associatedOverlayID);
         $associatedOverlay.css('z-index', 1000 + lStack * 2);
         $modal.css('z-index', 1000 + lStack * 2 + 1);
-        return
+        return;
       }
       else {
         overlayID = _generateID(),
--- a/js/leanModal.js
+++ b/js/leanModal.js
@@ -30,13 +30,12 @@
         $modal.css('z-index', 1000 + lStack * 2 + 1);
         return;
       }
-      else {
-        overlayID = _generateID(),
-        $overlay = $('<div class="lean-overlay"></div>');
-        // Store a reference of the overlay
-        $overlay.attr('id', overlayID).css('z-index', 1000 + lStack * 2);
-        $modal.data('overlay-id', overlayID).css('z-index', 1000 + lStack * 2 + 1);
-      }
+      
+      overlayID = _generateID(),
+      $overlay = $('<div class="lean-overlay"></div>');
+      // Store a reference of the overlay
+      $overlay.attr('id', overlayID).css('z-index', 1000 + lStack * 2);
+      $modal.data('overlay-id', overlayID).css('z-index', 1000 + lStack * 2 + 1);
       
       $("body").append($overlay);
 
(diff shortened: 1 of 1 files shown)


js/leanModal.js
```diff
@@ -20,15 +19,25 @@
...
       lStack = (++_stack);
-
-      // Store a reference of the overlay
-      $overlay.attr('id', overlayID).css('z-index', 1000 + lStack * 2);
-      $modal.data('overlay-id', overlayID).css('z-index', 1000 + lStack * 2 + 1);
-
+      
+      if ($modal.css("display") == "block") {
+        // The modal is already opened, just put it back on top
+        associatedOverlayID = $modal.data('overlay-id'),
```
carlosperate: Should end in `;` not `,`

js/leanModal.js
```diff
@@ -20,15 +19,25 @@
...
+        $associatedOverlay = $('#' + associatedOverlayID);
+        $associatedOverlay.css('z-index', 1000 + lStack * 2);
+        $modal.css('z-index', 1000 + lStack * 2 + 1);
+        return
```
carlosperate: don't forget the `;`

pascal-de-ladurantaye:
@carlosperate fixed the line endings

js/leanModal.js
```diff
@@ -20,15 +19,25 @@
...
+      }
+      else {
```
carlosperate: Since the previous `if` statement ends in a `return` there is no point to have an `else` statement here. What's encapsulated within the `else` curly brackets is no-way separated from the rest of the function.

js/leanModal.js
```diff
@@ -20,15 +19,25 @@
...
+      
+      if ($modal.css("display") == "block") {
+        // The modal is already opened, just put it back on top
+        associatedOverlayID = $modal.data('overlay-id');
+        $associatedOverlay = $('#' + associatedOverlayID);
+        $associatedOverlay.css('z-index', 1000 + lStack * 2);
+        $modal.css('z-index', 1000 + lStack * 2 + 1);
+        return;
+      }
+      else {
```
carlosperate: Since the previous if statement ends in a return there is no point to have an else statement here. What's encapsulated within the else curly brackets is no-way separated from the rest of the function.

carlosperate:
Looks good, I'll see if I can test it with the fixed footer and I'll report back.

carlosperate:
Yep, working fine for me with fixed footer as well. And it brings the modal to the top if multiple modals opened, works as expected.
:+1:

pascal-de-ladurantaye:
Perfect

darbs:
:+1:

carlosperate:
@Dogfalo  @acburst do you guys have any feedback on this? relatively simple update to fix an open issue.

phylor:
This PR also solves the symptoms of another issue which is: if you call `leanModal()` multiple times on the same element(s), opening a modal results in multiple overlays to be shown. In the worst case, the background is black and stays black if the modal is closed.

Ideally, `leanModal()` would be idempotent, ensuring that the handler is only added once to every element, even if `leanModal()` is called several times on the same element.

If you're working with SPA frameworks like Meteor, it is really hard to ensure that `leanModal()` is only called once. Not being able to do that makes modals completely unusable.

A merge would be greatly appreciated!

pascal-de-ladurantaye:
Closing this now

PR #1695, opened 2015-07-08
Fix 0 width for empty elements.

peinorim:
If an empty element has class ".col .sx", then the element will have no width. With a 1px min-height, the width applied by the responsive classes .col .sx will be applied.


Dogfalo:
Having 0 width when the element is empty is expected and desired behavior

PR #1696, opened 2015-07-08
Improved keyboard select search

GianlucaGuarini:
The select search is not really efficient. With this pull request the selected items will be filtered using a regular expression containing the characters typed by the user returning the first item found.
I have also optimized the search timer implementing using a simple debouce


acburst:
Does not pass automated tests.

PR #1699, opened 2015-07-08
Using ES6 module and a compilation step to support globals and amd

Munawwar:
Hi. I've attempted to bring AMD support (through refactoring to ES6 modules syntax).

Idea is to use a specific subset of ES6 modules to support AMD and globals.
Here are the following coding restrictions/conventions that will apply:
1. Only default exports and default imports can ever be used. (Read more [here](https://github.com/Munawwar/es6-module-clean-transpilation) on why).
2. Every file should have exactly one export (and it should be a default export as stated in 1). Only then AMD can be nicely supported.
3. (This one is a restriction due to the dumb compiler I wrote) - Every import and export should be on a separate line.

[The reason why I had to write a compiler is so that libraries like these can play nice with existing code bases.  From what I gather, the others like esperanto, babel etc would require users of this library to refactor their code as well.]

I had to do some additional stuff as well - e.g. split global.js to two files..one to return Materialize and one to handle the Velocity fix.

Please check whether I got the dependencies right.

The diff looks horrible because of indentation. I suggest anyone reviewing this to use kdiff3 (since it detect indentation):
```
git difftool -y -t kdiff3 HEAD~2:HEAD
```

I haven't added compilation step for AMD yet, but the way to generate AMD is to do:
```
npm install
node node_modules/es6-module-clean-transpilation/transpiler.js -t amd -i js/ -o js-amd/
```

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -72,6 +72,9 @@ module.exports = function(grunt) {
 
 //  Concat
     concat: {
+      process: {
+        //compile-modules animation.js --to test-c.js --type globals --imports "jquery:jQuery,./jquery.easing.1.3:jQuery"
+      },
       options: {
         separator: ';'
       },
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -72,10 +72,11 @@ module.exports = function(grunt) {
 
 //  Concat
     concat: {
-      process: {
-        //compile-modules animation.js --to test-c.js --type globals --imports "jquery:jQuery,./jquery.easing.1.3:jQuery"
-      },
       options: {
+        process: function(src, filepath) {
+          var compiler = require('es6-module-clean-transpilation');
+          return compiler({src: src, type: 'globals'});
+        },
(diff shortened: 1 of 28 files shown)


Munawwar:
I think the .travis.yml error isn't related to my code. However I broke grunt testem, Will look into how to compile the files before testing..

Munawwar:
Fixed the tests as well. I hope this will be merged since users of module systems can start using materialize seamlessly with any build system (and only include what is needed).

Munawwar:
@Dogfalo: @acburst: Ping...

Munawwar:
You'd get least conflicts with:
```git merge -X ignore-all-space 95ab2baad```

If you aren't interested in this change then please close it.

Munawwar:
I am giving up on this. Not only because there is no response on this, but also the JS has many `$(document).on`s (That is, code assumes markup is ready..but for web apps that may not be the case).

PR #1731, opened 2015-07-14
Problem when using Webpack with Materialize

xhanrot:
Here are the error messages : "assignment to undeclared variable: Materialize" and "assignment to undeclared variable: validate_field".
I guess it stems from strict mode.

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -69,7 +69,7 @@
       validate_field($inputElement);
     });
 
-    validate_field = function(object) {
+    window.validate_field = function(object) {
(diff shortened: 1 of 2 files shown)


kriscarle:
In case it helps I've also worked around this by adding the following to my webpack config
```
plugins: [
new webpack.ProvidePlugin({
          $: "jquery",
       jQuery: "jquery",
       "window.jQuery": "jquery",
       Materialize: "materialize-css",
       "window.Materialize": "materialize-css"
    })
]
```

Issue #1594, opened 2015-06-25
[JS] Input file don't work when loading with ajax


reidark:
Hello guys.
In my application, I have a form that loads with ajax, and in this form has some input file.
I've tested, when i put this input file in a normal page, it works fine, but, when it load with ajax, it don't work as expect.

Have a fix for this?

carlosperate:
Fixed in https://github.com/Dogfalo/materialize/pull/1733 https://github.com/Dogfalo/materialize/commit/7bec3393da3888ebdb83abf45680182d049ded6e.

PR #1733, opened 2015-07-15
file input for dynamically loaded elements

dritterman:
Fix for issue #1594 - [JS] Input file don't work when loading with ajax


acburst:
Thanks!

PR #1737, opened 2015-07-15
Empty placeholder causing active label

drainx:
If the placeholder on an input is empty, we shouldn't classify that as active.

Signed-off-by: Daniel Rix Drainx1@live.com

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -59,7 +59,7 @@
 
     $(document).on('blur', input_selector, function () {
       var $inputElement = $(this);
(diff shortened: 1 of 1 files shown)


Dogfalo:
Fixed by #1709

revati:
Are there any updates with this issue?

drainx:
I just outright abandoned using this project and moved to MDL.
Another reason I abandoned was because of how #1746 wan handled and the general idea of a broken master and a broken  main site that maintainers kept refusing to fix.
This issue is just a testament to the maintainers that they don't care about the project anymore or they are more interesting in doing the work themselves as opposed to using PRs properly.

PR #1738, opened 2015-07-15
Add a css attr to limit the max height of the dropdown by the availab…

victoreassis:
Add a css attr to limit the max height of the dropdown by the available space in the document. This avoid the dropdown to break the rest of the layout in some cases. If this value is higher than the default max height set in the css, then it preserve the default.


Dogfalo:
Later we plan to add more sizing logic to the dropdown ourselves to ensure it stays within the viewport

Issue #1669, opened 2015-07-03
search bar broken - close button


TobyEalden:
See the search bar at http://materializecss.com/navbar.html

The close button appears on the line below, obscuring the first word on next line of text.

Reproduced on Chrome 43.0.2357.130 and Firefox 33.0

TobyEalden:
Appears to be the material-icons causing the problem. If you replace:

```
<i class="material-icons">close</i>
```

with

```
<i class="mdi-navigation-close close"></i>
```

it works as expected.

acburst:
It is working on master, We haven't updated the docs yet

PR #1746, opened 2015-07-16
Lets update the documentation for navbar search.

drainx:
How about a fix for #1669 . Just because it works in master doesn't mean the docs shouldn't be updated. It in fact does not work as intended in master, which was the whole point of that issue in the first place.

Signed-off-by: Daniel Rix <Drainx1@live.com>

### DIFF
--- a/jade/page-contents/navbar_content.html
+++ b/jade/page-contents/navbar_content.html
@@ -272,7 +272,7 @@ $(".dropdown-button").dropdown();
               <div class="input-field">
                 <input id="search" type="search" required>
                 <label for="search"><i class="material-icons">search</i></label>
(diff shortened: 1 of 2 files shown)


Dogfalo:
You are using the wrong icon files

drainx:
Please elaborate.
It is clearly broken as pointed out in #1669, and this is the appropriate fix for it, again as stated in #1669

carlosperate:
You are pulling in the use of the old icons. The google material icons should be used instead: http://materializecss.com/icons.html

drainx:
So then why is it broken on the main site, which is using that icon family?

PR #1760, opened 2015-07-20
Mistyped variable $secondary_color instead of $secondary-color

brunoarueira:
Hey guys,

I found some mistyped variables and open this PR to fix that.


tomscholz:
See: https://github.com/Dogfalo/materialize/blob/master/sass/components/_variables.scss#L118

PR #1772, opened 2015-07-22
Update _materialbox.scss

arist0tl3:
Added max-width:none to prevent cropping of zoomed images when used in columns.


Dogfalo:
Not really sure what issue this is fixing, we need to see a codepen illustrating this.

PR #1779, opened 2015-07-23
Added support for left and right for dropdown

emilien-puget:
Added the possibility to define a container for a dropdown instead of using window to know if the activated part of the dropdown can fit in the container (usefull for secondary action in a collection or in a card).
Added the possibility to force the activated part of a dropdown to be on the left or right of the origin (usefull for icon origin).


Dogfalo:
alignment was added a while ago, just forgot to close this PR.

PR #1782, opened 2015-07-23
rebuilt select caret, on update select structure

fernandoPalaciosGit:
Hello, I notice when not added a new <option> element, only refresh seleted item, the caret will repeated. example to replicate bug:

``` javascript
$inputSelect.find(':selected').text('rename text');
$inputSelect.material_select();
```

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -284,9 +284,7 @@
       // Tear down structure if Select needs to be rebuilt
       var lastID = $select.data('select-id');
       if (lastID) {
-        $select.parent().find('span.caret').remove();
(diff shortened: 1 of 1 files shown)


javidalkaruzi:
This solved my problem.

carlosperate:
The repeating caret was fixed in https://github.com/Dogfalo/materialize/commit/cbcf4c8f6c229f42dba1dd3ccde1788d7f96ff80 (or at least it did at the time).
Now, I haven't had a chance to test this pull request, but could you provide an example of where the issue occurs with the latest code in master and it gets fixed with this pull request?

PR #1785, opened 2015-07-24
Add waves effect to fixed floating buttons demo.

qsysmine:
In the buttons demo, there is no `waves-effect` class on the floating action button.

### DIFF
--- a/buttons.html
+++ b/buttons.html
@@ -153,14 +153,14 @@
         </code></pre>
 
         <div class="fixed-action-btn" style="bottom: 45px; right: 24px;">
-          <a class="btn-floating btn-large red">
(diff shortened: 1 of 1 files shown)


carlosperate:
If you'd like to submit a pull request for the website documentation, I think you are supposed to edit the files in the `jade` folder, as the HTML pages would be automatically generated from those.

Apart from that, the floating action button on the documentation page does not have the waves effect added in both the page itself and the documented HTML. I assumed this was deliberate, but I do not know, either way both the documentation and actual page code should correlate and this pull request only edits one of the two.

qsysmine:
@carlosperate Oh, I didn't realise that. I'll close this.

PR #1795, opened 2015-07-25
Removed bug for RTL sidenav in responsive where it will go out of the screen

rajool:
In wide displays (more than 992) if the menu is right-aligned, it will be disappear just after the page loads.


acburst:
I can't seem to recreate the bug. Do you mind making a codepen to demonstrate it?

PR #1798, opened 2015-07-25
Add waves effect to fixed floating buttons demo.

qsysmine:
[Fixed from previous pull request, thanks.]


carlosperate:
You are only editing the page code to add the waves, and the example code does not have the effect added. As I mentioned in the other pull request, I don't know if leaving the wave effect out of the floating action button was a deliberate decision, but if you edit the page code, then the example code should be edited too.

qsysmine:
@carlosperate Fixed.

PR #1791, opened 2015-07-24
Removing link to non-existent showcase example - Material Admin Template

akshaymohite:
Showcase example link is dead !
http://demo.geekslabs.com/materialize-v1.0/index.html

PR #1813, opened 2015-07-28
Fixed navbar Mobile Collapse not giving correct example

TheSerenin:
The example on the NavBar page, at Mobile Collapse, doesn't give the right example. It's saying you need to add

```html
<i class="material-icons">menu</i>
```

But it should be

```html
<i class="mdi-navigation-menu"></i>
```

Like how it is on the SideNav page under JavaScript.

This commit fixed that.

### DIFF
--- a/jade/page-contents/navbar_content.html
+++ b/jade/page-contents/navbar_content.html
@@ -326,7 +326,7 @@ $(".dropdown-button").dropdown();
   &lt;nav>
     &lt;div class="nav-wrapper">
       &lt;a href="#!" class="brand-logo">Logo&lt;/a>
-      &lt;a href="#" data-activates="mobile-demo" class="button-collapse">&lt;i class="material-icons">menu</i>&lt;/i>&lt;/a>
+      &lt;a href="#" data-activates="mobile-demo" class="button-collapse">&lt;i class="mdi-navigation-menu"></i>&lt;/i>&lt;/a>
       &lt;ul class="right hide-on-med-and-down">
         &lt;li>&lt;a href="sass.html">Sass&lt;/a>&lt;/li>
         &lt;li>&lt;a href="badges.html">Components&lt;/a>&lt;/li>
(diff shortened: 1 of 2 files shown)


drainx:
I'll just reference #1746 
I have given up on Materialize and have moved to MDL, since the devs working on MDL actually know what they are doing.

TheSerenin:
In regards to that PR you made, I'm actually going to make all those changes for the icons...

I'm closing this PR, in favor is a newer one.

drainx:
I honestly wouldn't bother, because they will almost certainly give you the same reply they did me, and the PR will end up sitting for weeks, or just outright being rejected. Merging (no pun intended) all of the changes is rather useless, as others have tried to get them pulled and have been rejected.

In addition, the dev team here has outright refused to remove (and replace) dead demo links #1791 et al.

TheSerenin:
Fair enough.

carlosperate:
The `<i class="mdi-navigation-menu"></i>` is the old icon system which is being phased out for the official google material icons, the ones using `<i class="material-icons">menu</i>`.

If there is any old documentation from the jade folder that is still using the old `<i class="mdi-xxxx"></i>` then a pull request will certainly be appreciated. If there are any problems using the new icons, then if you open an issue then the team can look into fixing those.

TheSerenin:
Those old icons aren't working indeed.

If what you're saying is true, then the documentation is ahead of the implementation.

drainx:
@carlosperate As I pointed out in #1746 (and several others, this issue included,) the new implementation is broken.

There is no excuse for #1791 et al. though.

carlosperate:
I'm not debating that there might be issues with the new icons, just that if anything is broken then it can be looked into and fixed, rather than revert back to the old implementation in the documentation.

About the specific icon issue you mentioned in your last comment from https://github.com/Dogfalo/materialize/pull/1746, has https://github.com/Dogfalo/materialize/commit/43d43c27960ba7fd6eff16b4daadcd7e2dba6294 fixed that one? If not then we can open a new issue to capture that problem.

TheSerenin:
I'm updating the docs atm and I'll make a new PR for it.

TheSerenin:
@carlosperate See #1814

TheSerenin:
And about the buttons not working, please see this commit: https://github.com/TheSerenin/periodic-table/commit/3cf19df9df8fad62e0217cf936649f9358dea3a3

PR #1814, opened 2015-07-28
Replaced all mdi-* occurences in the documentation

TheSerenin:
All mdi-* class references are now the new material-icons references.

I didn't touch the tests, as they seem to break when doing this. (old version?)
Same goes for the templates, which are broken as is.

### DIFF
--- a/about.html
+++ b/about.html
@@ -32,7 +32,7 @@
   </head>
   <body>
     <header>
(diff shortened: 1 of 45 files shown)


carlosperate:
Some of the changes from this commit are editing the html that is automatically generated by the template engine. Please edit only the html contents from `/jade/page-contents/` as indicated in the [contributing](https://github.com/Dogfalo/materialize/blob/master/CONTRIBUTING.md) file.

TheSerenin:
I rolled all page commits back to 1b92fb2. This is how I checked them out after cloning.

tomscholz:
We don't use the `mdi-*`class anymore

PR #1853, opened 2015-08-04
Init expanded-collapsible with JS even if data-collapsible isn't set

marco-kretz:
The problem is, that if you init a collapsible like this:

```
$(document).ready(function(){
  $('.collapsible').collapsible({
    accordion : false
  });
});
```

And DON'T set data-collapsible="expandable" it won't be expandable.
Fix attached :)

### DIFF
--- a/js/collapsible.js
+++ b/js/collapsible.js
@@ -95,7 +95,7 @@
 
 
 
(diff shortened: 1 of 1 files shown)


Dogfalo:
html attribute options have been removed

PR #1818, opened 2015-07-29
Fixed Extra End </i> Tag

danielhickman:
The fixed multi FAB example had an extra `</i>`.

PR #1864, opened 2015-08-06
Update buttons.html

ricocrivelli:
Fixed Action Button example has two &lt;/i> tags.


carlosperate:
This is a duplicate of https://github.com/Dogfalo/materialize/pull/1818/

PR #1865, opened 2015-08-06
SideNav produces multiple #sidenav-overlay divs

V07D:
When I click on side menu item, sideNav.js generates multiple sidenav-overlay divs. It's because of absence of checking near line 268 in sideNav.js. This commit fixes it.
materialize-bug

### DIFF
--- a/js/sideNav.js
+++ b/js/sideNav.js
@@ -255,7 +255,7 @@
               $('body').css('overflow', 'hidden');
               // Push current drag target on top of DOM tree
               $('body').append(dragTarget);
-              
+
               if (options.edge === 'left') {
                 dragTarget.css({width: '50%', right: 0, left: ''});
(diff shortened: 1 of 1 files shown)


Aghassi:
I think this is missing a small tidbit. Although this prevents multiple overlays, which is good, I don't think it prevents the creation of another sideNav object in memory. Every time you use `sideNav()` this gets called, and all though it may not create more overlays, it will create another sideNav to watch. So what should also happen is to check if a sideNav already exists in memory, and if it does don't do anything. That, plus what you have here, should work nicely. I made an issue ticket here regarding this issue because I have had it for a while:
https://github.com/Dogfalo/materialize/issues/1894

V07D:
Thank you for your answer, Aghassi. I have to use multiple siveNav() calls in my project too, so your way to fix the problem with memory leaks was very useful for me.

Dogfalo:
No longer an issue in 1.0.0 alpha

PR #1887, opened 2015-08-11
Decentralized icon on the nav bar for small screens

brunocalou:
The search bar icon gets decentralized when the screen is small.
I tested this behavior on
- Google Chrome 44.0.2403.130 (64-bit) - elementary OS 0.2.1
- Firefox 39.0.3 - elementary OS 0.2.1
- Google Chrome 44.0.2403.133 - Android 4.4.4

screenshot from 2015-08-10 21 55 01
screenshot from 2015-08-10 21 55 10

The problem is that on small screens, the line-height: 1 of the material-icons class is dominant. So, adding line-height: inherit to the icon inside the navbar solves the problem.

Working Codepen: http://codepen.io/brunocalou/pen/RPdYGR
Resize the page to see the bug

### DIFF
--- a/bin/materialize.css
+++ b/bin/materialize.css
@@ -5251,6 +5251,7 @@ nav {
     position: relative;
     height: 100%; }
     nav .nav-wrapper i {
+      line-height: inherit;
       display: block;
       font-size: 2rem; }
   @media only screen and (min-width : 993px) {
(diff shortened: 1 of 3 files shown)


brunocalou:
The same behavior is observed on any icon on the nav bar. So, changing the `nav .nav-wrapper i` instead of adding it just to the `nav .input-field label i` solves the problem.

screenshot from 2015-08-11 15 13 03
screenshot from 2015-08-11 15 13 10

Working Codepen: http://codepen.io/brunocalou/pen/waObxZ
Resize the page to see the bug

tomscholz:
This seems fixed in the current release

PR #1913, opened 2015-08-14
Add extra large class for floating buttons

andrefrd:
Simply add class .btn-extra-large to floating buttons


tomscholz:
The spec only specifies a mini and normal version
https://material.io/guidelines/components/buttons-floating-action-button.html#

PR #1924, opened 2015-08-17
Small error in length comparator

harrinsonmb:
If I create an input with a length of 10, I write 10 characters and I focus out the input, the entry is marked as invalid because there is a small error in the comparator.

http://codepen.io/anon/pen/vOwZXa


acburst:
Thanks, that's embarrassing haha

PR #1931, opened 2015-08-18
replace '&lt;' to '<'

andrefrd:


carlosperate:
The `&lt;` is used to be able to display html code inside the`pre` and `code` tags.

PR #1936, opened 2015-08-19
Fixed autofill

gfazioli:
Added the `input` event. The `input` event is fired when the autofill changed the input value.


Dogfalo:
autofill should be handles already.

PR #1954, opened 2015-08-24
added transition property to .hoverable, off of :hover pseudo-selector

LA1CH3:
The `.hoverable` class was including the `transition` rule on the `:hover` pseudo-selector. This was causing a smooth transition on hover-in, but a quick, harsh transition on hover-out (since transition is only included on hover). `transition` is instead added to the `.hoverable` class so that the transition is apparent on hover-in and hover-out.


Dogfalo:
Sorry for the late reply, but this has since been fixed.

PR #1969, opened 2015-08-28
Hide dropdown when button is clicked and 'dropdown-content' is active

iammary:


iammary:
@carlosperate what do you think about this pull request? This is evident when data-beloworigin is set to true.

PR #1973, opened 2015-08-28
Sticky footer visible in all the browsers

damianopetrungaro:
If you decide to use this class edit for sticy footer change it also in the css core.

In this way the browser will see good in all the browsers, also safari, an IE8


Dogfalo:
This doesn't work too well as the content in body can be covered up by the footer, since it does not fall into the flow of the DOM

PR #1974, opened 2015-08-28
line-height from .toast {} changed

felipe-douradinho:
On smaller resolutions, when text are large (a lot of chars), the line-height crashs and nothing is readable

before

after


acburst:
Fixed in 629231712132d3c6959cbdb8fab067d15790bc23

PR #1980, opened 2015-08-31
Fix path to materialize.js

saadq:


carlosperate:
The problem with this fix, is that then the demo from http://materializecss.com/templates/starter-template/preview.html will not load the js correctly either.

saadq:
Oh, I see. I just sent this pull request because it took me awhile in trying to figure out why the mobile nav wasn't working. It might be helpful if it was mentioned somewhere that some assets need to have their paths corrected.

PR #1981, opened 2015-08-31
Adds alignment option logic for Dropdowns

### LINKED ISSUE #1841: dropdown content do not align from right to left
The .dropdown-content seems to align only from left to right relatively to its activator(.dropdown-button).
The issue is when the activator is on the right corner of the page or div. Just like this:
materialize-dropdown-bug2

It would be nice to have an option to align the .dropdown-content from the right to left relatively to its activator

msolters:
Dropdowns now take an optional `alignment` property when initialized:
```
$('.dropdown-button').dropdown({
  alignment: 'left' || 'right'
});
```

which controls the edge of the parent button that the dropdown list aligns with.  Dropdown list still obeys gutter rules.  Default, if not specified, is `alignment: 'left'`.

This seems to fix #1841.

### DIFF
--- a/js/dropdown.js
+++ b/js/dropdown.js
@@ -14,7 +14,8 @@
       constrain_width: true, // Constrains width of dropdown to the activator
       hover: false,
       gutter: 0, // Spacing from edge
-      belowOrigin: false
+      belowOrigin: false,
+      alignment: 'left'
     };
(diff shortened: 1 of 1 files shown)


acburst:
Thanks for this!

PR #2022, opened 2015-09-10
Removed icons duplicates

NeXTs:
List of duplicates

clear_all
chat_bubble_outline
chat_bubble
comment
snooze
sort_by_alpha
stop
subtitles
surround_sound


carlosperate:
If the icons are correctly displayed with different key words, I don't see any need to remove them, it also makes it easier to find an icon you want by trying to search for a word instead of reading all of them.

NeXTs:
@carlosperate key words are identical. But you are right, I had to edit jade file.

PR #2031, opened 2015-09-13
indicatorsHeight

kendrick-k:
add indicatorsHeight option (default : 40px), can be set to 0 px for exemple il you want the indicators on the image + documentation if accepted


kendrick-k:
@Dogfalo hi, is it still useful ?

PR #2038, opened 2015-09-14
Fix broken meteor packaging due to use of minimized code

### LINKED ISSUE #2037: Upgrading to 0.97.1 from 0.97.0 fails meteor app
Just saw and upgrade this morning you pushed, but after update my app was a bit confused :) http://prntscr.com/8g3z9p  Just to let you know. Had to go back to 0.97.0. Great work otherwise with Materialize!

serkandurusoy:
Fixes #2037

### DIFF
--- a/package.js
+++ b/package.js
@@ -3,7 +3,7 @@
 Package.describe({
   name: 'materialize:materialize',  // http://atmospherejs.com/materialize/materialize
   summary: 'Materialize (official): A modern responsive front-end framework based on Material Design',
-  version: '0.97.1',
+  version: '0.97.1_1',
   git: 'https://github.com/Dogfalo/materialize.git'
 });
 
@@ -36,7 +36,7 @@ Package.onUse(function (api) {
     'font/roboto/Roboto-Thin.ttf',
     'font/roboto/Roboto-Thin.woff',
     'font/roboto/Roboto-Thin.woff2',
-    'bin/materialize.css',
-    'bin/materialize.js',
+    'dist/css/materialize.css',
+    'dist/js/materialize.js',
   ], 'client');
 });


iliasch:
This pull request hides the problem, doesn't fix it.

serkandurusoy:
@iliasch I'd rather see this accepted as a temporary fix than leaving all meteor applications permanently broken.

nelreina:
+1

serkandurusoy:
@Dogfalo any news? All updated meteor apps get cripled and this should be and easy workaround until the root cause can be fixed.

iliasch:
There root cause has been identified and I have submitted a pull request, but there seems to be no response from the author. For the time being you can apply this fix yourself:

    $ cd .meteor/packages/materialize_materialize/0.97.1/web.browser/packages
    $ chmod u+w materialize_materialize.js
    $ nano materialize_materialize.js
    > find: window.Materialize={}
    > replace with: Materialize={}
    > save changes
    $ chmod u-w materialize_materialize.js
    $ cd ../..
    $ chmod u+w web.browser.json
    $ nano web.browser.json
    > find: "length": 123264
    > replace with: "length": 123257
    > save changes
    $chmod u-w web.browser.json

Rebuild your meteor application and everything should be working fine. This has been tested on both a production and development environment. When a new version is available, you will be able to update as usual.

serkandurusoy:
Thanks @iliasch that's spot on.

But still, I cannot use this because I'm sharing code with other developers and build machines so applying this on every single environment woud be error prone.

What we've opted for is to fork the repo and use it as a shared local package.

Also it seems there is a more complete PR #2056 which seems to be both a safer fix and addresses the minification problem as well.

Anyone reading this, should go show their support over at #2056

Issue #1764, opened 2015-07-20
Select : class active doesn't match with the selected li


TyrionGraphiste:
Hi !

I recently discover an issue with the Select plugin.
"active" dynamic class doesn't match with the selected li element, as a matter of fact, the DOM is updated once time after we opened again the dropdown menu.

You can try yourself at the demo page : http://materializecss.com/forms.html#select and looking which li has the active class.

Can you fix this please ?

Moreover, your plugin doesn't include select with multiple options : http://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select_multiple

Can you implement it ?

Anyway thanks for everything.

carlosperate:
I am not sure I understand the first part of the question. If you are refering at chaing the select dynamically you'll have to reinitialise the materialize select or destroy it and create it again as indicated in the "Updating/Destroying Select" section from http://materializecss.com/forms.html#select .

About the second part there is already a request for mutiple selections in the select: https://github.com/Dogfalo/materialize/issues/1002

TyrionGraphiste:
I'll try to explain a little better than previously.
When you select an option, an "active" class is dynamically added. But the class isn't applied on the correct item after you selected it. There is a "lag".

For the multiple select I finally created a a plugin for materialize, is there a render screen :
![multiple_select](https://cloud.githubusercontent.com/assets/10064887/8972631/1994b424-365c-11e5-981d-abbd96cc7b1c.PNG)

Dogfalo:
You should be getting the selected value like this: http://www.w3schools.com/jsref/prop_select_value.asp

TyrionGraphiste:
Hi Dogfalo,

I have no problems regarding the recovery of the option selected, the only one I encounter concerns the position of the active class on the selected item.

What you have linked is a way to set the select with a new option, not really get it.
For getting the selected option we can do like this :

var select = $("#select"),
      selectedOpt = select.find("option:selected");

So we can do like this for getting the option, it's right, but in this case what is the point to placing an active class ? Especially if it is not on the right.

tejovanthn:
I added the following code near `select.siblings` in the `toggleEntryFromArray` function to update the select fields too.
```
      var child = select.children('option').eq(entryIndex);
      if(child.attr("selected")){
          child.removeAttr("selected");
      }
      else{
          child.attr("selected","selected");
      }
```

TyrionGraphiste:
You should create your own PR and submit it.

Issue #1002, opened 2015-03-25
Request: <select multiple>


zion822:
I would like to use ```<select multiple>``` similar to Chosen (http://harvesthq.github.io/chosen/). I managed to style Chosen somewhat like materialize but it still looks off. 

Any chance you will including the ```<select multiple>``` to the framework?

ianvieira:
I created a similar solution using selectize.js, I don't know if I can do a pull request for it or should I create a new project just with it.

crokobit:
@ianvieira  May you shared your solution ? I find everywhere but can not find any good solution yet :cry: .

allthesignals:
I'm interested as well. I mostly just like how Chosen provides a nice built-in typeahead.

DmitrySharikov:
hope, you guys know how to solve it 😓

lmerotta:
Still nothing going on with this issue ? It would be great if this could be done !

Issue #540, opened 2015-01-27
Select multiple 


ffscalco:
Is there any support for select with `multiple` tag?

I've tried to use `$('select').material_select()` in `<select multiple>` but it's render the single select type.

acburst:
We do not support that at the moment. You could use checkboxes?

ffscalco:
Sure, It's pretty amazing =D
Thanks for the answer!

ricardodovalle:
Do you know another stylized component that does the same?
I need to use select multiples because I have a large collection data.
Thanks

PR #2041, opened 2015-09-15
Add support for materialize multiple select. Fix bug with active class.

TyrionGraphiste:
This PR adds the support for multiple select which wasn't still supported by materialize, you just need to add the attribute "multiple" to the select element.
This fix the active class, so we can recover all selected items with `li.active` or with getting the selected options from the native select, both are working.

Here there is a [demo](https://jsfiddle.net/TyrionGraphiste/ocowsehq/2/embedded/result/).
I hope it will fix some issues, it's working on all browsers on my side, take time to test it on old browsers if you can.

Here there is all the issues fixed with the PR :
- [#1002](https://github.com/Dogfalo/materialize/issues/1002)
- [#1764](https://github.com/Dogfalo/materialize/issues/1764)
- [#540](https://github.com/Dogfalo/materialize/issues/540)

Then, don't hesitate to change the code ! I'm not a developer and tried to follow your code style.

PS : Sorry for typos ! I'm not **en** !

### DIFF
--- a/css/ghpages-materialize.css
+++ b/css/ghpages-materialize.css
@@ -5919,13 +5919,19 @@ button.btn-floating {
     width: 100%;
     text-align: left;
     text-transform: none; }
-    .dropdown-content li:hover, .dropdown-content li.active {
+    .dropdown-content li:hover, .dropdown-content li.active, .dropdown-content li.selected {
       background-color: #eee; }
+    .dropdown-content li.active.selected {
+      background-color: #e1e1e1;
+    }
     .dropdown-content li > a, .dropdown-content li > span {
       font-size: 1.2rem;
       color: #26a69a;
       display: block;
       padding: 1rem 1rem; }
+    .dropdown-content li > a, .dropdown-content li > span > label {
+      top: 0.6rem;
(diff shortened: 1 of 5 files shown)


Dogfalo:
Amazing, we'll look into integrating this for next release!

TyrionGraphiste:
Nice ! Thanks :)

acburst:
Thanks! I had to revert some of your dropdown code because it didn't pass some of our test cases, but besides that it's working great

TyrionGraphiste:
Hi ! That's great to have merging this :)
Unfortunately after your merge, select doesn't show up anymore on TAB key.

Anyway, real thanks it's really nice !

acburst:
Fixed in bee1b20

sml01:
Hi,

great to see all issues fixed. but not sure if my scenario below is fixed.

Let's say I have a select with multiple and there are 4 options. I select 1st, 2nd, 3rd options, then un-select 2nd, and 1st, so, we can see that there is only 3rd option is selected. but when we submit this, the selected value are 1,2, and 3.

PR #2042, opened 2015-09-15
Fixes Dropdown Alignment

### LINKED ISSUE #2039: Dropdown offset seems strange...
Not sure if this is a bug in the scripting itself or in my implementation, but I'd appreciate the help either way. I've created a fiddle of the basic layout of this part of my page. Ever since upgrading to 0.97.1, the dropdowns appear in weird places. They seem to be close to where they should  be when the viewport is small and the columns containing them are s12, but when it goes up to m6 or something similar, the position is way left of where it should be. Any idea what the issue is? 

http://jsfiddle.net/5LawL4ez/2/

Thanks!

msolters:
Reworked dropdown's `placeDropdown()` method.  2 changes:

*  Alignment is done by reference to the `origin.position()` rather than the `origin.offset()`.  `origin.offset()` worked approximately when the parent container of the dropdown was comparable to the entire window width (i.e. small screens), but no longer true when the parent container was much smaller than the window.  The position of the dropdown list should be absolute wrt the parent container, not the window, so `origin.position()` is the correct quantity to use here.  This change fixes #2039.
*  When the dropdown went off the left or right edge of the screen, the `options.alignment` was permanently changed, which resulted in incorrect alignment after the user resizes the screen.  This is no longer true, by using a local `var alignment` on each call to `placeDropdown()` that will defer to the user's specified option whenever possible without *overwriting* that option.


msolters:
Accidentally closed this.  Let's see if I can reopen.

acburst:
Fixed the positioning bug here 3869f12dacd8b3e413f554b97dfb1fd1933db8ae. There have been changes with the dropdown alignment since you created this PR so I think those resolved those problems. Let me know if they persist

msolters:
Have you actually tried [`3869f12`](https://github.com/Dogfalo/materialize/commit/3869f12dacd8b3e413f554b97dfb1fd1933db8ae)?  It does not fix either #2039 or #2052.  Here, try [this example](https://jsfiddle.net/r1c5ywL1/).  I've copy-pasted `3869f12`'s `dropdown.js` into it.  The right-aligned dropdown is in the wrong place.  (Make sure to let the window be large enough that it does not get force-aligned.)

In addition, it is not responsive -- if the dropdown moves within the page due to resizing (try shrinking the fiddle window so the cards are stacked vertically, then open the dropdown) the edge detection alignment remains locked in because you are overwriting `options.alignment` instead of using a locally scoped variable.

This PR resolves *both* of these problems.  [Here is proof](https://jsfiddle.net/oa181eks/).  Please reconsider accepting it.

acburst:
Added in 561ee50eddfdecbdd0ddf971f360e4547a89d23e. 

Sorry for not originally using your code. It hadn't passed some of our dropdown test cases which means I usually don't look too closely at the code if needed. It should be working now. Thanks!

msolters:
Yes, `561ee50` appears to work as expected now.  Awesome!  Thanks.

msolters:
I think you can close #2039 now.

PR #2044, opened 2015-09-16
Fixed materialize.js path in template files.

ManasN:
Upon downloading the template from website, one is left with a broken view. That is, without any images.
This happens due to src of materialize.js pointing to a non-existant folder. 
This commit fixes it.


carlosperate:
The problem with this fix, is that then the demo from http://materializecss.com/templates/starter-template/preview.html and the parallax one will not load the js correctly either. This is an issue that needs to be looked into keeping that in mind.

PR #2045, opened 2015-09-16
fix chips margin issue on mobile view

rasanga:
## Before fixing the isse

Before fixing the isse
## After fixing the isse

After fixing the isse


rafaelcavalcante:
Nice! I was looking forkward to create a pull request for this problem. Nice work, @rasanga!

rasanga:
Aren't we merging this PR?

PR #2056, opened 2015-09-20
Closes #2043

grsabreu:
You can not define Materialize with window.Materialize because Meteor package export won't be able to export the variable.
Tell me if I have to do some changes before merging :).

PR #2057, opened 2015-09-20
Revert breaking change in Materialize init

iliasch:
Materialize = {} and window.Materialize = {} should do the same thing, but the window object is not accessible in node environments, so this change breaks compatibility with Meteor.
This change has been tested in both a dev and live environment and seems to fix the reported problems.


carlosperate:
Covered in https://github.com/Dogfalo/materialize/pull/2056

Issue #2077, opened 2015-09-23
Issue with Meteor 1.2 & Materialize v0.97.1


docgecko:
I just upgraded to the latest version of Meteor and this Materialize package causes errors.  If I remove the package, my app works fine again.  Here are the errors that show up in the console:

Uncaught TypeError: Cannot set property 'guid' of undefined(anonymous function) @ materialize_materialize.js:10(anonymous function) @ materialize_materialize.js:15(anonymous function) @ materialize_materialize.js?852eb8ab99e3508edd310071537cb33698a58a2b:47(anonymous function) @ materialize_materialize.js?852eb8ab99e3508edd310071537cb33698a58a2b:56
global-imports.js?42ba686d5ec5de5008618b324d34de30700ad7e2:44

Uncaught TypeError: Cannot read property 'Materialize' of undefined(anonymous function) @ global-imports.js?42ba686d5ec5de5008618b324d34de30700ad7e2:44

Many thanks for any help.  Cheers,

D.

Crenshinibon:
Same here!

mbanting:
Same here. Downgrading to 0.97.0 works.
Thanks!

docgecko:
Yeap, 0.97.0 works fine, thanks!

SoilChang:
same here. the whole app broke.

carlosperate:
Duplicate of https://github.com/Dogfalo/materialize/issues/2037.

PR #2078, opened 2015-09-23
Fix for issue 2077

li8:
Fix or issue: https://github.com/Dogfalo/materialize/issues/2077

The previous version 0.97.1 had initialized globally Materialize. 
But it was not accessible in the scope of the function in global.js


carlosperate:
Covered in https://github.com/Dogfalo/materialize/pull/2056

PR #2097, opened 2015-09-29
showStaggeredList works on table tags

davidbielik:
`Materialize.showStaggeredList` does not work on `<table>`' rows:`<tr>`

This pull request fixes this by checking the selector type and choosing the correct children element to animate. By default it will look for `<li>` tags. If the selector is a `<table>` it will look for `<tr>` tags instead.

Working example: http://codepen.io/anon/pen/meOYWd


Dogfalo:
staggered list has been removed

PR #2108, opened 2015-10-01
Check if password field has been autofilled

goodnickoff:
In chrome (45.0.2454.101 m (64-bit)) if the password field has been pre-filled `change` event will not be triggered. So this event listener:
https://github.com/Dogfalo/materialize/blob/master/js/forms.js#L26
will not be executed.

Also value of such field is empty string. So, the `active` class will not be added here:
https://github.com/Dogfalo/materialize/blob/master/js/forms.js#L9

So, after page is loaded the form looks like this:
password

I have added checking if the field has been pre-filled to add `active` class to it:

```
$(this).is('input:-webkit-autofill')
```

Also i have added 10ms timeout to `Materialize.updateTextFields()` method invocation because fields are not pre-filled by browser right after document is ready. Looks not very beautiful, but I didn't find another way to handle this behavior.


Dogfalo:
Should be working now

tomscholz:
@Dogfalo Related pr https://github.com/Dogfalo/materialize/pull/1936

PR #2125, opened 2015-10-05
Toast: Line Breaking and Improved Spacing

msolters:
If toast content is too long for the screen, it currently does not line break properly because toast height is fixed and line-height is the same as container height.  In addition, long toast messages at large screen widths can force the far side of the toast container to be flush with the edge of the window, which is inconsistent with the 5% or 7% the user expects.

This PR introduces two fixes:

*  Multiline toast content now line breaks appropriately, and toast height will expand or reduce to fit content.  Default (single-line) toast height will always revert to the value specified by `$toast-height`.
*  Toast container at medium and large widths sets a 5% or 7% margin on *both* sides of the toast, which produces a more consistent user experience.  This matches how Google renders toast, for example, on the Google Store.

### DIFF
--- a/sass/components/_toast.scss
+++ b/sass/components/_toast.scss
@@ -10,12 +10,14 @@
   @media #{$medium-only} {
     min-width: 30%;
     left: 5%;
+    right: 5%;
     bottom: 7%;
   }
   @media #{$large-and-up} {
     min-width: 8%;
     top: 10%;
     right: 7%;
+    left: 7%;
   }
 }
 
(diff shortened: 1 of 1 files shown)


dcoferraz:
:+1:

msolters:
I'd like to point out that technically, according to the Google Material guidelines these toasts -- or as they call them, snackbars -- should not be more than one line.

In reality I think a lot of people leverage these tools to go above and beyond Google's prescriptions, and this CSS fix is something I find myself doing *every god damn time I implement Materialize* in a new project, so I'm very biased towards having toasts responsive out-of-the-box.  Still, an interesting read: https://www.google.com/design/spec/components/snackbars-toasts.html

acburst:
Looks good, thanks!

PR #2128, opened 2015-10-06
Collapsible elements not in a list

johngeorgewright:
Any chance you'd consider changing the collapsible functionality so it doesn't require being in a list?


PeggsDev:
:+1:

tomscholz:
Make sure to check out the [CONTRIBUTING](https://github.com/Dogfalo/materialize/blob/master/CONTRIBUTING.md) document inside the root folder!

jeroenqui:
Will this be accepted when implemented properly?

tomscholz:
I see nothing that would speak against it

PR #2130, opened 2015-10-06
Small screen icon fix

w01fgang:
On small screens (mobile phone) the icons on navigation and search bar
was misaligned


alijaffar:
I've had a similar problem, I fixed it by adding this straight to the CSS:
```
@media only screen and (max-width : 601px){nav .material-icons { line-height: inherit; } /* fix */}
```

tomscholz:
Still an issue?

w01fgang:
It seems fixed in the latest version.

Issue #1537, opened 2015-06-17
Enable html in tooltip content


Dahkon:
Hi,

I would really appreciate if you could add html content on tooltip, at least for <br>.
I use tooltips for forms online help and a one line tooltip is not enough.

My easy fix is to replace .text with .html on line 1195 and 1222 of materalize.js but that would be better to implement this optionnaly with something like :

data-enable-html="true"

Thanks,

Bound3R:
would be great!, recently I needed to add a line break or an icon but got to resign and used plaintext :(

Rvanlaak:
:+1:

albertpak:
:+1:

blinchi:
![](https://assets-cdn.github.com/images/icons/emoji/unicode/1f44d.png)

We need this!

rikato:
Jup waiting for this aswell

TyrionGraphiste:
Hi, as said in [#2064](https://github.com/Dogfalo/materialize/issues/2064)

You just have to change the line 59 into the tooltip.js file :

    newTooltip.children('span').text(origin.attr('data-tooltip'));

As :

    newTooltip.children('span').html(origin.attr('data-tooltip'));

dcworldwide:
Really need this. I have two issues, 1) i need html in tooltips and 2) if you have a lot of text in the tooltip, you need to be able to control its width.

NeXTs:
:+1:

wdupuy:
As I don't want to modify the materialize JS file, I use this code
```
$(document).ready(function() {
	$('.tooltipped').each(function(index, element) {
		var span = $('#' + $(element).attr('data-tooltip-id') + '>span:first-child');
		span.before($(element).attr('data-tooltip'));
		span.remove();
	});
});
```

akelmj:
set data-tooltip using jquery not woking:

javascript:
    $('.tooltipped').tooltip({delay: 50,tooltip:"I am tooltip"});

html:
  <a class="btn tooltipped" data-position="bottom" data-delay="50"   >Hover me!</a>

tomscholz:
PR: https://github.com/Dogfalo/materialize/pull/2079

PR #2168, opened 2015-10-17
Support Line Break('\A') to Tooltip message

chi-bd:
This patch makes it possible for ToolTip to display multi-Line message.
ex.) &lt;a class="tooltipped" data-tooltip="first line\Asecond line"&gt;


chi-bd:
Related #1537

dcworldwide:
Really need this feature. Also want to be able to use html. I.e. divs, br etc

Dogfalo:
This has been added

PR #2198, opened 2015-10-23
Feature requests: Images in select

SuperDJ:
As seen here: https://www.google.com/design/spec/components/text-fields.html#text-fields-auto-complete-text-field a select could have icons.

I came up with the following for select:

Still todo:
- Fix a bug when option is selected in select label isn't activated

What I have changed is also in the README.md


acburst:
With a pull request of this size it would be very helpful to have some code examples or docs to go along with this. Once these are made I can take a closer look at it and the code.

SuperDJ:
@acburst I've updated forms.html so it has more documentation and example code. There are still some bugs

acburst:
This is looking good! I can try looking into some of the bugs and help polish some of the styles for the eventual merge into master.

acburst:
Looking at this PR again. Because it is rather large, it would be better if you separated this out into 2 pull requests. 1 For images in select, and 1 for autocomplete. This way bugs for either feature won't hold up the submission of the other.

SuperDJ:
@acburst Removed autocomplete from master and created a new branch for it because it has more bugs than select with images

acburst:
thanks!

PR #2199, opened 2015-10-23
Fix font-family names

dduqueti:
Hello! Tried to use different Roboto font types and there weren't working properly. Noticed that it was just a minimal typo error and all names stayed as 'Roboto' and not the particular ones. I changed the names so they are usable. Thanks for working on this btw!


Dogfalo:
Hey @dduqueti,
I don't think there is an error with how we do it. Google fonts declares font weights in a similar fashion.

https://fonts.googleapis.com/css?family=Roboto

PR #2209, opened 2015-10-26
Fix #1958: add documentation for preloader spinner sizes

### LINKED ISSUE #1958: Documentation of the avaible options
There is a place where i can see what options i can use, for example for buttons what classes i can apply to sizes? big, small ? colors?
There is great examples in your site but to see how i can use certain things i need to inspect element.

kaymmm:


acburst:
Thanks!

PR #2219, opened 2015-10-28
Fix noUiSlider style when it is not connected

### LINKED ISSUE #2216: Style bug when nouislider is not connect
Hi before all, thanks for all the contribution with materialize.

I am trying to use a form -> range -> noUiSlider (3rd party), but I need that the two selectors must not be connected between them. This is my code:

var slider = document.getElementById('test5');
  noUiSlider.create(slider, {
   start: [20, 80],
   step: 1,
   range: {
     'min': 0,
     'max': 100
   },
   format: wNumb({
     decimals: 0
   })
  });

mccraveiro:
Fix #2216

### DIFF
--- a/extras/noUiSlider/nouislider.css
+++ b/extras/noUiSlider/nouislider.css
@@ -31,7 +31,7 @@
 .noUi-origin {
 	position: absolute;
 	right: 0;
-	top: 0;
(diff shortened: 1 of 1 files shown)


acburst:
thanks!

PR #2234, opened 2015-11-01
fix .active class to :active state

jamesgiroux:
under navbar links, &hover, &.active has active set as a class rather than a state.  this is an update to fix that.


acburst:
This is not a mistake. The active class is for denoting the current page your are on in the navbar.

PR #2246, opened 2015-11-04
[btn-floating] ligature alignment

m0x72:
On mobile safari, inline elements do not seem to be aligned to top by default. 
Fixes vertical alignment of <i> element / font icons.


tomscholz:
Is this still a problem?

PR #2252, opened 2015-11-05
Updated velocity to version 1.2.3

finetype:
Was having problems using newest features in velocity with a site I was working on that I also was using Materialize on. Realized velocity was included in Materialize, and was out of date.


finetype:
With permission, I'd also like to include a line that is essentially

if (jQuery.Velocity) return;

to prevent materialize from overwriting a velocity object that exists in the window already. If someone is using the UI pack and/or has already loaded velocity, we shouldn't reload it or overwrite it. Might add a console.log notification accordingly, though, for users of materialize who don't realize velocity comes as part of the package.

acburst:
Feel free to add that to this PR

finetype:
* Updated minified Velocity.js 1.2.2 present in materialize.min.js and materialize.js to Velocity 1.2.3. 
* Added conditional to the previous two files to confirm that `jQuery.Velocity` does not exist before defining it, to prevent accidentally overwriting a Velocity object otherwise declared by user (potentially a Velocity object extended with UI pack, or a newer version). 
* Added a console.log statement to notify user of this behavior, so they can stop accidentally importing Velocity twice if they were not aware Materialize included Velocity. 
* Finally, reminified velocity.min.js with the included conditional check, so that it is still present when compiled with included grunt/gulp script.

Side note: added a comment showing making clear the presence of the Hammer library (which was previously appended, in Materialize.js, to the minified Velocity codeblock, which was very confusing at first).

acburst:
Thanks! Really appreciate the detailed writeup.

smeijer:
Has this change been rolled back? I depend on some functionality from Velocity 1.2.3, but it's not present in the current release of materialize.js.

When comparing source code, you can see that the string `stop|finish|finishAll` in velocity (introduced in `1.2.3`, is still `stop|finish` in materialize.js bundle (`1.2.2` and older).

finetype:
easy way to check is to diff this PR against master (no time to do myself, sorry)

PR #2271, opened 2015-11-09
Use Grunt-postcss for vendor prefixes

obouchari:
Use PostCss (with the Autoprefixer plugin) to add vendor prefixes and remove the outdated one.

Note: grunt-autoprefixer has been deprecated in favour of grunt-postcss.


Dogfalo:
Awesome, thanks a lot @obouchari. We've been wanting to add this for a while but haven't gotten around to it.

PR #2272, opened 2015-11-10
fixed the font paths in meteor package.js

s7dhansh:
The current version causes console errors - font not found - due to wrong file paths. Fixed it.


Crenshinibon:
Thank you. Please push a new version to Atmosphere.

PR #2291, opened 2015-11-13
Handle autofocus, fix #869

### LINKED ISSUE #869: Input Forms Don't Move the Title with 'autofocus'
The title doesn't leave the text box when you start typing on a preselected input.
```html
<input id="start-color" type="text" autofocus></input>
<label for="start-color">Your Color</label>
```
`Your Color` in this case will be stuck as the placeholder and doesn't animate to a title.

leomao:
It should check `autofocus` attribute after `Materialize.updateTextFileds()`.

There is not need to create a function `Materialize.updateTextFields`.
Just do it inside `$('document').ready` before checking autofocus.

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -1,3 +1,16 @@
+// Function to update labels of text fields
+Materialize.updateTextFields = function() {
+  var input_selector = 'input[type=text], input[type=password], input[type=email], input[type=url], input[type=tel], input[type=number], input[type=search], textarea';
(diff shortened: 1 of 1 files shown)


Dogfalo:
The update field fiction is used for angular and meteor sites

leomao:
I see.  I add it back in the following commit.
You may want to squash these commit before merging.

PR #2294, opened 2015-11-14
Fix select meteor

s7dhansh:
add the ul after newSelect; instead of appending it to body. meteor does not have a well defined body


acburst:
Can you remove the select fix? Ill merge that one line in manually. It's better to keep Pull Requests that refer to one feature or fix so that they can be merged independently of each other.

acburst:
Actually nvm this is small enough to just merge. Thanks for the fix! I know we're generally slow with meteor issues as we are not very familiar with it.

PR #2300, opened 2015-11-15
Added options for scrollspy.

trusta:
Added options : 
- **activeClass :** permits to change the class name added.
- **getActiveElement :** permits to change the element where add class.

For example, these options are useful to make works the scrollspy with a navbar. In this case, the active class is added to the _li_ but not to the _a_ element. To do this, the _getActiveElement_ function must return the _base.parent()_ object.


Dogfalo:
scrollspy has been rewritten

Issue #1422, opened 2015-05-25
npm require misses hammerjs, then picker.js, etc ...


joergd:
When I use the npm module,

a) I cannot require it directly as referenced in my package.json file - I need to use a path to the node_modules/materialize-css/bin folder.

b) then when I require the materialize.js file from that node_modules location, it misses hammerjs

c) then I get hammerjs, then it is missing picker.js etc ...

I'm not sure where I am going wrong here, but I am finding it impossible to require('materialize-css') in my browserify app ... :(

Am I doing something wrong?

ktmn:
Same here.

    import Materialize from '../node_modules/materialize-css/bin/materialize.js';

or

    let Materialize = require('../node_modules/materialize-css/bin/materialize.js');

produces 

    [16:17:47] gulp-notify: [Error running Gulp] Cannot find module './picker.js' from 'C:\...\node_modules\materialize-css\bin'
    [16:17:48] gulp-notify: [Error running Gulp] Cannot find module 'hammerjs' from 'C:\...\node_modules\materialize-css\bin'

Riuujin:
same here

aplcat:
I have same issue!

Nxt3:
I can't even do `var materialize = require('materialize-css');` in my node.js app. Does anyone have a work around? I ran `npm install materialize-css --save` and it doesn't work. (I also tried a direct path to the bin in the module, but that didn't work either).

choonchernlim:
Agreed. Is there a npm module for picker.js?

danazkari:
`npm install materialize-css -D` works but `require('materialize-css')` doesn't.... this is a bummer! How are we supposed to use the npm version of it?

ndesorden:
For hammerjs you can npm install hammerjs --save
For picker you can npm install pickadate --save and change in materialize.js
"require('./picker.js')" to require(''../'../pickadate/lib/picker.js')

     ;/*!
     * Date picker for pickadate.js v3.5.0
     * http://amsul.github.io/pickadate.js/date.htm
     */

     (function ( factory ) {

    // AMD.
    if ( typeof define == 'function' && define.amd )
        define( ['picker', 'jquery'], factory )

    // Node.js/browserify.
    else if ( typeof exports == 'object' )
        module.exports = factory( require('./picker.js'), require('jquery') )

    // Browser globals.
    else factory( Picker, jQuery )

    }(function( Picker, $ ) {

Yes, not very elegant :(

[...]

PR #2302, opened 2015-11-15
1422: Make materialize work with requireJS. Added hammerjs and jquery…

llhupp:
… to package.json. Set bin/materialize.js as main entry point.

#1422


acburst:
You shouldn't edit bin/materialize.js directly because this is a generated, compiled file. You could probably find a better place to put it in global.js or something and then compile it into bin.

llhupp:
I only changed `js/jquery.easing.1.3.js` and recompiled with `grunt monitor`... am I not supposed to commit the updated `bin/materialize.js`?

acburst:
Oh, I see. Yeh, adding it to jquery.easing isn't the best idea because if we want to update it it will be lost. But looking at our compilation order, there really is no good place to put it atm. We'll have to think about this a little more I think.

llhupp:
Ah... My mistake, I totally glossed over that that was a jQuery submodule... I thought it was just some hacked together implementation of jQuery. You're right, I shouldn't update that.. Maybe I should add that global.js like you were talking about..

Dogfalo:
Closed since this was incorporated into #2381 unless I'm mistaken.

PR #2317, opened 2015-11-17
Fixes for select (and multiple)

TyrionGraphiste:
This PR should fix some bugs on select and multiple select :

- resize select width on window resize.
- scroll to the selected option when keyboard touch is pressed.
- some cleanup code.
- autoselect a selected option.


TyrionGraphiste:
Too much useless changes, I'll do a clean PR.

Issue #2274, opened 2015-11-10
<select multiple> broken


MeMyselfI:
Hi.

Well, where to start... 
1. I think it doesn't handle < option selected >
2. It doesn't support icons
3. val(["value1","value2"]) isn't reflected in the popup

Because of this programatically changing the selection isn't possible - even with material_select('destroy') and material_select() afterwards.

MeMyselfI:
Additionally, val() doesn't deliver the actual selection... :-(

TyrionGraphiste:
Hi,

Support for icons has been added after the select multiple support. So they just choose to not support icons for multiple select.

For select items two ways :
- With the selected options : $("select option:selected:not(:disabled)");
- With active class in the dropdown : $("ul#id li.active");

For selected option it's working but not really as intented :
- no checked checkbox
- no active class on the active option (li)

acburst:
1. Tested locally, adding selected on a single option in a multiple select works, Adding multiple selected does not. (Will look into other issues mentioned by TyrionGraphiste) TODO

2. Images in select were just added and we wanted to release as soon as possible so it will be done in the future. TODO

3. Tested locally, if you do a .val() on the actual original select element, it will give you the correct value. You can also set the value on the original select element. Just make sure you are setting to the correct value. On our examples, our values are [1,2,3].

rodofatos:
Hello, @acburst any idea when the fix for issue 1 will be fixed ? It's really somewhat critical at this moment to have an option to define what are selected options and have them checked in the dropdown. Thanks.

rodofatos:
Also referring issue 3, currently in 97.3 the $(selector).val() returns an array of values that contains all the values that were checked at least once(even though now they are unchecked). 
![select values](https://cloud.githubusercontent.com/assets/12541574/11181667/d243f596-8c6c-11e5-9a85-9b90658a4be2.jpg)

TyrionGraphiste:
I'm working on a fix which be pushed as a PR soon.

Odonno:
I fixed it. I'll push it on a PR asap.

Sabartius:
Another thing is the shown text, after you select several options. The optionstexts are ordered they way you clicked them. If you check first option5, then option3 and then option4 the diplayed text is "option5, option3, option4". A better way would be to preserve the order of my options, so that the text would be "option3, option4, option5".

TyrionGraphiste:
This is not important.

Odonno:
[...]

PR #2321, opened 2015-11-17
Update correctly selected value of checkbox

Odonno:
Based on issue #2274, I correctly reset `selected` value of each checkbox (unique or multiple).


acburst:
Thanks!

PR #2322, opened 2015-11-17
Fixes for select (and multiple)

TyrionGraphiste:
This PR should fix some bugs on select and multiple select :

- resize dropdown width on window resize.
- scroll to the selected option when keyboard touch is pressed.
- some cleanup code.
- autoselect a selected option.

This PR also takes into account the [PR #2321 ](https://github.com/Dogfalo/materialize/pull/2321) from Odonno.

I think the better way to resize dropdown width should be to add the part code of resize into dropdown.js. What do you think @acburst  ?

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -294,11 +294,10 @@
       $select.data('select-id', uniqueID);
       var wrapper = $('<div class="select-wrapper"></div>');
       wrapper.addClass($select.attr('class'));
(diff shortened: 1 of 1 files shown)


TyrionGraphiste:
I think the better way to resize dropdown width should be to add the part code of resize into dropdown.js.

acburst:
I removed the resize dropdown width because I think there should be a cleaner way to add that to dropdown later.

PR #2337, opened 2015-11-20
offset issue - fix css priority

sebastienva:
Hello,

I recently updated materialize and I got issues with offsets.

The fix on master  https://github.com/Dogfalo/materialize/commit/d5db2a3290ab2848f4482aae2933756b8e357346 doesn’t really work.
If we stack multiple sizes the problem is still here.

For example :
class="col m8 offset-m2 l4”

On desktop, offset-m2 is still active because l4 doesn’t override it with a margin-left:0px.

The original problem with offsets comes from this commit https://github.com/Dogfalo/materialize/commit/511043f50a41366c5a4f384f7baa0e30d1702066 which breaks css priority.

I hope my explanation is clear, I just put back margin-left and an additional loop to keep the old css priority.  

Regards.

### DIFF
--- a/bin/materialize.css
+++ b/bin/materialize.css
@@ -5090,95 +5090,107 @@ nav ul a span.badge {
     .row .col[class*="push-"], .row .col[class*="pull-"] {
       position: relative; }
     .row .col.s1 {
-      width: 8.33333%; }
+      width: 8.33333%;
+      margin-left: 0; }
+    .row .col.s2 {
+      width: 16.66667%;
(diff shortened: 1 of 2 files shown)


acburst:
Ahh this makes sense. I was wondering why I had margin-left on the original col sizes. I'll write some tests for grid soon. But this looks good. Though have you checked that it all works well with push and pull?

sebastienva:
I only checked the example from the documentation.

PR #2344, opened 2015-11-21
Material box undefined error fix

rthomson83:
The ancestorsChanged object can end up being undefined. The .css() method
was being called on an undefined object causing an error to be thrown.


Dogfalo:
Already fixed.

PR #2345, opened 2015-11-21
Add localization document with a sample for Japanese.

chi-bd:


tomscholz:
The idea in general is great, but I think Japanese is the wrong language for demo purposes.

Issue #901, opened 2015-03-12
Select list dropdown scrollbar closes the dropdown.


MattgyverHub:
I have a select list of US states which requires a scrollbar to limit the items shown to 7 or so. However, using the scrollbar immediately closes the droplist. The folks at the Angular/material git have a fix, but I don't understand how to implement it for the select list dropdown. Thanks in advance! https://github.com/angular/material/issues/1739  The following is exactly what's happening:
![9202f166-c03f-11e4-9404-e85da611d0b5](https://cloud.githubusercontent.com/assets/10520782/6622648/458afb76-c8b4-11e4-9a60-493a4bbc5f1a.gif)

Dogfalo:
The scrollbar on dropdowns worked for me on Chrome + OSX. What browser/OS are you on?

MattgyverHub:
Chrome on Windows 7 with the latest release build of Materialize v0.95.3.

Dahkon:
Same problem for me.
It happend on all browser.

The problem is located here :
      $newSelect.on('blur', function(){
        $(this).trigger('close');
      });

The blur event is fired once we click on the scrollbar.

emurselovic:
Is this issue resolved? 

If it is resolved where i can find solution.

Pat-Paran:
I did this and it works Fine

 // $newSelect.on('blur', function(){
      // // $(this).trigger('close');
      // 
      // });
      $('.select-dropdown').find('span').on('click',function(){
      		$newSelect.trigger('close');
      });

Comment out the on blur after that every click of options will get the value and close the dropdown.

Pat-Paran:
if you want to exit the dropdown when you click outside  Use this.
$(document).mouseup(function (e)
		{
		    var container = $(".select-dropdown");

		    if (!container.is(e.target) // if the target of the click isn't the container...
		        && container.has(e.target).length === 0) // ... nor a descendant of the container
		    {
		       $newSelect.trigger('close');
		    }
		});

emurselovic:
Thanks very much. It is working as expected.

webmandman:
I know this was fixed for 0.8.3, but yesterday I updated to 0.11.0 and it stopped working for IE 11 edge, still works fine in Chrome. 

Still the same issue as MattgyverHub, but this can't be about the fix not being commited or applied. The autocompleteController.js file has been updated over 40 times and in 0.11.0 it still has the same logic. Some where along it stopped working for IE. 
[...]

PR #2346, opened 2015-11-21
Fixed a problem that would be closed by clicking the scrollbar(especially IE).

chi-bd:
Please refer to #901 .

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -517,6 +517,16 @@
           };
 
       $newSelect.on('keydown', onKeyDown);
+
+      var onMouseDown = function(e) {
+        // preventing the default still allows the scroll, but blocks the blur.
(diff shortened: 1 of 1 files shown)


Dogfalo:
This fixes the problem, but introduces another because now the select no longer closes when you normally click on an item.

chi-bd:
>but introduces another because now the select no longer closes when you normally click on an item. 

Really?
I tested InternetExplorer 11.0.9600.17937 and Chrome 47.0.2526.73m on Windows8.1, and does not occur the problem that you say.
When I click on an item normally, close the selectbox automatically.

m00k1e:
Hello!

 

New message, please read <http://reaprich.com/enthusiasm.php?3hrx>

 

hello@mookiecreative.com

PR #2351, opened 2015-11-22
scrollFire: callback function, repeat fire option

kielni:
Currently, the callback option of scrollFire has to be a string, which gets called like this:

```
var callbackFunc = new Function(callback);
callbackFunc();
```

This means the function must be global, and there's no way to pass it any context. 7861638 checks if the callback is a function, and then calls it.  

I wanted to use th scrollFire for infinite scroll.  I put my trigger element at the bottom of the list, and my callback renders more content when the trigger element comes into view.  Currently, there's no way get it to invoke the callback more than once.  d7999ea adds `repeat` option, which will invoke the callback each time the trigger element comes into view.  

Here's an example of how I used a callback function and the `repeat` option to set up infinite scroll in a React component:

```
showMore: function() {
    this.setState({to: Math.min(this.state.items.length, this.state.to+pageSize)});
},
componentDidMount() {
    // ... load content
    Materialize.scrollFire([
        { selector: '#moreMarker', offset: 0, callback: this.showMore, repeat: true }
    ]);
},
render() {
    return (
        <div>
            <ItemList items={this.state.items.slice(0, this.state.to)} />
            <div id="moreMarker" className="chip">More...</div>
        </div>
    );
}
```

### DIFF
--- a/jade/page-contents/scrollfire_content.html
+++ b/jade/page-contents/scrollfire_content.html
@@ -14,7 +14,7 @@
         <pre><code class="language-javascript col s12">
   var options = [
     {selector: '.class', offset: 200, callback: 'globalFunction()' },
-    {selector: '.other-class', offset: 200, callback: 'globalFunction()' },
+    {selector: '.other-class', offset: 200, callback: this.onOtherClassVisible },
   ];
   Materialize.scrollFire(options);
         </code></pre>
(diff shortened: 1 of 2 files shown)


Redmega:
Is there a reason this was not approved from last year? It's a good option to have.

ThibauldNuyten:
How do I use the repeat function of scrollfire? Adding repeat:true did nothing.

acburst:
scrollFire removed

PR #2356, opened 2015-11-23
Select tests

acburst:
Basic jasmine tests for select.

Still many things that need to be tested, but first we need to fix some select issues so our build still passes.


Dogfalo:
Plans to add keyboard testing soon?
LGTM

acburst:
Once Select has more bugs cleaned up i'll add the more complex tests. The latest PR should have done most of it but just want to give it a while to make sure.

PR #2372, opened 2015-11-25
Chips style edited: large chips + custom close class

let-aurn:
- Ability to add `large` class to chips to get them bigger.
- Ability to use different close button component (such as font awesome) by just assigning to it `close` class.

### DIFF
--- a/sass/components/_chips.scss
+++ b/sass/components/_chips.scss
@@ -8,6 +8,18 @@
   padding: 0 12px;
   border-radius: 16px;
   background-color: $chip-bg-color;
+  
+  &.large {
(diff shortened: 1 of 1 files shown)


sass/components/_chips.scss
```diff
@@ -17,11 +29,11 @@
     border-radius: 50%;
   }
 
-  i.material-icons {
+  .close {
```
giubueno: I don't follow the logic here. Why did you change the name of the class?

giubueno:
This is a really old pull request. Could consider closing it, please? Thanks.

PR #2374, opened 2015-11-25
Fix autofocus bug

mtuck063:
Your checking for the autofocus attribute, then setting it active, then immediately updating the textfields which nullifies the change of adding the 'active' class.

I wrote a quick fix for the problem.


trostli:
Ive noticed this bug too. Would be great to get this bug fixed for forms that are autofocused on!

jameskerr:
Yes please merge this!

acburst:
Thanks for bringing this to my attention guys. Thanks @mtuck063 for the fix.

PR #2383, opened 2015-11-27
Fixed un-hovering animation

Zazo032:
Previously, when "un-hovering" an element, it had no animation. Now has a fade-out.


Dogfalo:
Can you modify the correct .scss file instead? If not we can't merge this. Thanks!

PR #2388, opened 2015-11-28
Fixed un-hovering animation

Zazo032:
Previously, when "un-hovering" an element, it had no animation. Now has a fade-out.


Dogfalo:
Thanks!

tomscholz:
I know I am a little bit late to the party, but `box-shadow: 0;` is not a valid value.

A better approach could be:
```scss
.hoverable {
  transition: box-shadow .25s;

  &:hover {
    box-shadow: 0 8px 17px 0 rgba(0, 0, 0, 0.2), 0 6px 20px 0 rgba(0, 0, 0, 0.19);
  }
}
```

PR #2393, opened 2015-11-28
Added tooltip tests and fixed tooltip delay bug

Dogfalo:


acburst:
LGTM, minor issue with constant mouseover and mouseouts.

PR #2394, opened 2015-11-28
Added basic tabs test for active tab, clicking tab and scrollable tabs

acburst:


tests/spec/tabs/tabsSpec.js
```diff
@@ -0,0 +1,76 @@
+jasmine.getFixtures().fixturesPath = 'tests/spec';
```
Dogfalo: No need to set the fixtures path.

Dogfalo:
LGTM, see comments

PR #2420, opened 2015-12-04
Update modals.html

juampi:
Remove extra spaces.


Dogfalo:
`modals.html` is compiled from jade. We can only merge changes made to those source files.

PR #2427, opened 2015-12-05
Fix #2425 "Select without options shows undefined?"

### LINKED ISSUE #2425: Select without options shows undefined?
:(

http://codepen.io/anon/pen/GoRBjG

inventti-leonardo:
Works for me :)

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -299,11 +299,7 @@
           valuesSelected = [],
           optionsHover = false;
 
-      if ($select.find('option:selected').length > 0) {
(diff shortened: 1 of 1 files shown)


TyrionGraphiste:
I don't think display nothing on the dropdown is a good idea.

inventti-leonardo:
Maybe, but the behavior of select without options does not display "undefined" using native html. :(

acburst:
I think this is fine. It does mimic what the browser default would do.

PR #2428, opened 2015-12-06
Slider re-init and destroy

devgrok:
These modifications allow the component to be re initialised with different options.
- Use case: Meteor detects changes in options and expects to be able to pass them on to the component without regenerating the html (from it's perspective it hasn't changed).
- If you want to change the speed at which the images transition etc

Also added a method to allow the component to be destroyed and for the image src url to revert back to it's state pre component init.
- Use case: I used this together with print detection. I used css to layout the slides to fill the space but with the image set as a background image trying to get the content to flow around the images was problematic.


acburst:
In v1, with the class rewrite this has been made much easier to destroy and reinit.

PR #2429, opened 2015-12-06
Modal - on close use same options open

devgrok:
To allow a close callback to be defined by the function that opens the modal and called when closed by a generic close trigger, this change stores the options used to open the modal and automatically uses them in close() if they have not been passed in.

### DIFF
--- a/js/leanModal.js
+++ b/js/leanModal.js
@@ -33,6 +33,7 @@
 
       // Override defaults
       options = $.extend(defaults, options);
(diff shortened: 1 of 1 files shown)


inventti-pedro:
Don't get why this branch isn't on master. Fix the annoying behaviour of losing options when only calling closeModal().

devgrok:
This has been fixed as part of the rewrite down in #3600

PR #2438, opened 2015-12-07
Cover case where input tag has no type specified

timtribers:
An input tag with no type attribute is considered to be of type="text",
but was not being picked up by Materialize.


Dogfalo:
Thank you @timtribers .

PR #2446, opened 2015-12-08
Update _toast.scss

fabyeah:
Ensure that the maximum width of a toast leaves the same space on both sides.


Dogfalo:
Looks good thanks!

PR #2450, opened 2015-12-09
Fix Dependency 'jquery' exists in both dependencies

listepo:


tomscholz:
Your pull request body is empty. 
Please add more information in order to allow us to undestand why you did the changes.

DanielRuf:
@tomscholz it is about one of the warnings of npm. jQuery is in the dev and normal dependencies, but should be not in the dev dependencies.

tomscholz:
^^

DanielRuf:
For example https://travis-ci.org/Dogfalo/materialize/builds/224550673#L153 ;-)

We should cleanup and update the dependencies in the near future and move to node.js 6.x LTS =)

tomscholz:
See https://github.com/Dogfalo/materialize/pull/4587

PR #2466, opened 2015-12-12
Fix navbar button

Zevran:
Resolve wrong behavior for button component inside navigation bar component.

Hover animation is now on the link instead of parent without override of button style.
Add margin left and right for correct display.
Keep actual behavior for standard links.


acburst:
Merged in dc6284aca06b1b0718f5d68862fa641544f2e84e.

I rebased master branch so it wont show as a merge here, my bad.

PR #2470, opened 2015-12-13
Added option to fix SideNav on Large Screens

florian-senn:
Useful when closeOnClick is preferred only on medium and small screens


TyrionGraphiste:
**options.closeOnLarge === true** or **options.closeOnLarge** is the same.

Dogfalo:
Fixed via #4299

PR #2473, opened 2015-12-14
no need to wait for document to be ready to extend the pushpin function

aguegu:
Just as the way other plugins load.


nerdyglasses:
I wonder why that one won't get processed. It seems like it would be important.

PR #2478, opened 2015-12-14
Feat/accessibility

mbrehin:
This is a major change for breadcrumbs.

Change html structure/tags in order to make breadcrumbs more accessible.
The aria-label is optional.

There are many more enhancements to provide to materialize in order to make it fully accessible.
This could be a starting point, or at least it would raise awareness among developers using materialize.

See https://developers.google.com/structured-data/breadcrumbs
Also https://github.com/paypal/bootstrap-accessibility-plugin


tomscholz:
Hey there 👋 
```
/materialize/sass/
/materialize/js/ 
```

mbrehin:
Hi there,

I'm sorry but I don't use materialize anymore and can't take the time to move my contributions to the right files. I hope you'll find people to help you enhance accessibility to your framework.

Cheers

PR #2487, opened 2015-12-15
Prefix icon misalignment with select

nathankoop:
Add documentation for select icon prefix


acburst:
Should be working on the latest version

PR #2503, opened 2015-12-19
fixed typo in scroll navigation on table doc

SimulatedGREG:
Not sure why but this really annoys me every time I see it, lol


kevgathuku:
@SimulatedGREG I think the correct file to edit would be https://github.com/Dogfalo/materialize/blob/master/jade/page-contents/table_content.html

Dogfalo:
Fixed in 0f87817

PR #2504, opened 2015-12-19
Chips js basic

alejandrovasta:


alejandrovasta:
@acburst I checked locally that the tests passed well without errors. I had used the chips branch previous to this PR.

acburst:
Because of the size of this PR it might take a while for us to properly merge this in. Also the holiday season is coming up for us which will also cause some delays.

alejandrovasta:
No problem, when you can @acburst. I tried to match the material basic functionality. The idea here is to finish it with the dropdown option for autocompleting and other cool stuff

acburst:
Finally merged this in. Thanks so much for your help. I cleaned up the code a bit and simplified things a little.

PR #2510, opened 2015-12-20
HTML in tooltip content

ebrian:
This adds a data-html attribute, defaulting to false, that, if set to true, will allow the user to have HTML in their data-tooltip attribute.


acburst:
Thanks! This was refactored from multiple PRs into one.

PR #2526, opened 2015-12-22
Fix custom error/success messages

### LINKED ISSUE #2525: [Bug] Form custom error/success messages
Hi,

Custom error or success messages replace spaces in string with line return ! Is it a correct behavior ?

screen shot 2015-12-22 at 12 24 30

Thanks

Zevran:
Wrong behavior with messages containing spaces
Resolve #2525

### DIFF
--- a/sass/components/_form.scss
+++ b/sass/components/_form.scss
@@ -94,6 +94,7 @@ textarea.materialize-textarea {
     content: attr(data-success);
     color: $input-success-color;
     opacity: 1;
+    white-space: nowrap;
   }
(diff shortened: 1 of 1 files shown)


remzr7:
+1

Dogfalo:
Hmm `nowrap` may not be the best solution. If you had a long error message, it would not line wrap properly. I think the better change is to set the label width to 100%

Gskartwii:
+1

Dogfalo:
This has been fixed

PR #2528, opened 2015-12-22
[Hotfix] Sidenav button

Zevran:
Buttons are displayed correctly on hover in sidenav.
Prevent buttons styles from being override by inherent styles like sidenav.
Add button in sidenav example.


acburst:
Thanks!

PR #2553, opened 2015-12-28
dropdown close on select option

baraamashaal:
Useful for a checkboxs or radio buttons dropdown something like multiselect but on a dropdown


Dogfalo:
The way to do this is just to have a stop prop if you do not want a certain element to close the dropdown

PR #2566, opened 2015-12-31
Error message didn't fit into label space

mateuszgachowski:
1. Error message was only fitting into label space [width 100% helps]
2. The other problem was with clicking directly to the label (it didn't animate, just select label text) [pointer events solved the problem]


Dogfalo:
Thanks for making the PR, however I didn't want to put pointer-events none on all labels so I modified what you wrote slightly so it only applies on inputs with the .validate class. See here 4327bb599a0b065a5fed3de2499ffce68c103ed5

:smile:

PR #2572, opened 2015-12-31
added basic click through to tooltips

simplydallas:
Added "pointer-events: none;" to .material-tooltip class so that supporting browsers allow click through and tooltips do not interrupt the intended function of the element they cover while animating away.

Change only added to SASS component in this pull request and not the compiled CSS.

### DIFF
--- a/sass/components/_tooltip.scss
+++ b/sass/components/_tooltip.scss
@@ -15,6 +15,7 @@
     overflow: hidden;
     left:0;
     top:0;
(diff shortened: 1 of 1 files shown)


Dogfalo:
I think this is a fairly safe change and can make the tooltips a bit more robust. The only downside is that people won't be able to copy text from the tooltip (if the developer put a longer timeout delay on it). Overall I think that's a small use case and this is a good change.

PR #2587, opened 2016-01-03
Added Sass variables for the button components

soniabrami:
- I tried to make the existing variables more consistent as well for example we had:
$button-floating-size but $button-color-raised.
I made them all $button - <button type> - <button css property>.

- for the background-color property I got inspired by  foundation and replaced:
$button-bg-color by $button-background
Background is one of the only css property we were shortening to bg. 

Please let me know what you think of the changes in general.
This way we can agree on some general principles before I start adding variables to the other components.

### DIFF
--- a/sass/components/_buttons.scss
+++ b/sass/components/_buttons.scss
@@ -1,46 +1,57 @@
 // shared styles
-.btn, .btn-flat {
-  border: none;
-  border-radius: 2px;
+.btn,
+.btn-flat {
+  border: $button-border;
+  border-radius: $button-radius;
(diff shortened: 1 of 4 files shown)


Dogfalo:
The changes look good, I still need to check out the changes to make sure nothing was accidentally changed.

I like the naming convention you used where you have `color` at the end, but you had it reversed in a few cases,
```Scss
$button-background-disabled: #DFDFDF !default;
$button-color-disabled
```

Dogfalo:
Looks great and thorough. Thanks!

PR #2603, opened 2016-01-06
npmignore update

capaj:
you've got a lot of files published on npm which aren't supposed to be there


TyrionGraphiste:
Hi, a title isn't a description.

Dogfalo:
Thanks!

PR #2606, opened 2016-01-06
Input error message width fix

JesseRWeigel:
When adding a custom error message to a form input that is longer than the input label, the error message text stacks vertically (see screenshot below). This fixes that by making the label fullwidth when it is active.
screen shot 2016-01-06 at 10 38 46 am

After the fix:

screen shot 2016-01-06 at 10 52 04 am

### DIFF
--- a/sass/components/_form.scss
+++ b/sass/components/_form.scss
@@ -137,6 +137,8 @@ textarea.materialize-textarea {
     font-size: 1rem;
     cursor: text;
     transition: .2s ease-out;
+    width: 100%;
+    text-align: left;
   }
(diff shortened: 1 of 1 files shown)


acburst:
Hmm i'm not sure how to reproduce this. We already have a css rule that makes the label 100% width to fix this issue. Do you have an example or issue that you can reference?

Dogfalo:
This is fixed but not realeasd

JesseRWeigel:
Awesome, thanks. You guys are doing a great job with Materialize. Keep up the good work!

PR #2608, opened 2016-01-06
Fix #2601

### LINKED ISSUE #2601: It's not scolling to item after type in select
In version 0.97.1 it's working
http://codepen.io/anon/pen/LGWbaZ

But in newer versions it's not working :(
http://codepen.io/anon/pen/obZzVR

inventti-leonardo:
Out of context variable was applying the behavior only in the last select :(

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -468,7 +468,7 @@
       }
 
       // Make option as selected and scroll to selected position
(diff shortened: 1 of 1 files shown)


acburst:
Thanks!

PR #2611, opened 2016-01-06
Modal enhancement to reset scroll position on exit

rthomson83:
When a modal has dynamic content that can change with every load of the
modal (my use case is with different forms in angularjs) the scroll
position persists between loads. Added an option that is true by default
that will reset the modal content scroll position x,y to 0,0 when the
modal window is closed.


iktakahiro:
+1

PR #2615, opened 2016-01-07
Modal button margin fix

rthomson83:
Adds some left right margin to visually separate the footer buttons when
there is more than one.


Dogfalo:
I think it is fine as is since modal buttons in material design should be flat anyways

Issue #473, opened 2015-01-17
Addition of time picker


newmanw:
Not sure if this makes sense within existing date picker, or a separate element.  However it would be nice to be able to pick a date and time.

syedsuhail:
It would be great if there was option for both a separate element and along with the existing date picker. Will this be implemented in the near future?

CoDEmanX:
In Android 5, there's a separate dialog for time picking. It's very well designed, it looks like a clock with two rings of labels (1-12 and 13-24), or, if your locale uses 12h format, one label ring and two buttons for AM and PM:

![screenshot_2015-01-21-13-48-51](https://cloud.githubusercontent.com/assets/1309249/5836391/689c0ab2-a175-11e4-940b-e71b19c1c220.png)

![android_new_time_picker_1](https://cloud.githubusercontent.com/assets/1309249/5836343/8c3647cc-a174-11e4-9fef-460c33fefdc8.png)

After picking the hour, there's a transition of the label ring(s) to a label ring for the minutes, in 5 minute steps. You can either tap on one of them, or drag to enter a precise minute. A confirmation is required after picking the minute to close the dialog.

If you click on the hour at the top, it will return to the hour picking mode (also with a nice animation).

syedsuhail:
This is great for mobile but there is no material design Date time picker for the web.

serkandurusoy:
The original pickadate.js which materialize's is based on does have a time picker. So perhaps, that could be extended similarly.

Dogfalo:
Echoing what @syedsuhail said, these time pickers are nice for mobile devices, but they do not work so well on desktop devices. More thought will have to go in to these.

bimmlerd:
I'd love to see some sort of timepicker as well :+1: 

For the moment if would be great to add a little bit of styling to the time input, as it is currently completely unstyled.

boustanihani:
:+1:

carlosperate:
@Dogfalo It's been quite a while since this was suggested, are there any plans to implement some kind of time picker?

larikitty:
@Dogfalo, I'd love to see some kind of time picker, as @carlosperate mentioned earlier.

ABeltramo:
Heavily inspired from http://simeonc.github.io/md-date-time/ I develop this very simple time picker for my project. I'll try to keep it as simple as possible.
Require Materialize and jquery.
 
http://jsfiddle.net/kakamiokatsu/o3osy173/

Any feedback is welcome!

[...]

PR #2622, opened 2016-01-09
Enhancement to add a date time picker

rthomson83:
I needed a datetime picker so decided to port one myself for using with an angularjs project and thought I would share it. 
I ended up using bootstrap-material-datetimepicker.js over at https://github.com/T00rk/bootstrap-material-datetimepicker

Not sure if this is the way you want to go and there's some enhancements like adding a today button etc that would be worthwhile.

image


rthomson83:
Should satisfy issue #473.

xxryan1234:
+1

cDima:
Wow, thanks @rthomson83 just what I needed.

rthomson83:
Glad it helped @cDima!

tomscholz:
See: https://github.com/Dogfalo/materialize/pull/4445

Issue #2413, opened 2015-12-03
touchStart event closes dropdown menu without triggering action


theberengersniper:
Hi guys,

There seems to be an issue that was introduced in 0.97.2 where the touchStart event interferes with dropdown menus.

Observed behaviour:
Dropdown menus are dismissed without triggering an action when the user taps a menu item.

Expected behaviour:
Dropdown menus are dismissed along with an action being triggered.

I appreciate this is likely a duplicate of #2212, however that issue hadn't received any comments from development. This is turning out to be something of a show-stopper in my project.

bmcbride:
:+1: This appears to be most noticeable on iOS Safari and doesn't seem to be much of an issue on Android/Chrome.

niwasmala:
Currently I'm working on hybrid apps using cordova and materializecss on iOS. I'm having similar problem for dropdown input, on touchStart. Is there any way to prevent dropdown closed? I already try attach touchStart and prevent default, but it failed, dropdown still closed. Any hints would be helpful. Thank you.

future-boy:
Hi everyone,

I ran into this recently on devices and after a little investigation the problem seems to be the event bindings on the dropdown. The closing of dropdowns were previously bound to the 'click' event as you would expect, but in 0.97.2 it appears the 'touchstart' event was added to this binding and I'm not sure of the reasoning behind that, because devices and their browsers do actually fire the 'click' event for taps in all the cases I have tried (so the dropdown would still close, albeit a litte later in the event heirarchy).

So to give an example, let's say we click a given node which opens the dropdown (the opening ONLY has the 'click' event bound by the way). Then the dropdown is shown and the close events are bound (both 'click' and 'touchstart' this time). Now if I'm on a touch device, as soon as I try to tap any type of our buttons inside the dropdown, the 'touchstart' fires first, the dropdown closes and (assuming you have your own actions bound to clicks) the button never gets the click event because technically you're no longer on the button for the 'click' event (the dropdown has already closed as of the 'touchstart').

The best fix for this is probably a revert of commit 2456c2e9047b9c3b724913af93603c8238d58c3e, although there may be reasons behind that addition that I'm not seeing here, but that's the commit where the issues started.

Hope this helps, thanks.

samba:
I'm seeing some trouble with it in Android on iOS as well, version 48.0.2564.104.

zhaoyao91:
[...]

PR #2632, opened 2016-01-10
fix: cannot click menu item on IOS

zhaoyao91:
By my debug, on ios, it will trigger touchstart event, which will close the menu before the click event, so the menu item won't be clicked.

### DIFF
--- a/js/dropdown.js
+++ b/js/dropdown.js
@@ -1,236 +1,240 @@
 (function ($) {
 
-  // Add posibility to scroll to selected option
-  // usefull for select for example
-  $.fn.scrollTo = function(elem) {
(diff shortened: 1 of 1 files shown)


samba:
Hi, this appears related to #2413 as well. Would love to see this merged (or otherwise mitigated) soon, as developers (such as myself) integrating this on new projects need to support iOS sensibly. Any workarounds in the mean time?

zhaoyao91:
this pr has a problem: on ios, although it will trigger action, but it cannot be closed by clicking on other place of screen

yzargari:
Anyone knows when this would get merged? I'm still having this issue in version 0.97.6 (+zhaoyao91 - your fix worked perfect for me)

PR #2658, opened 2016-01-15
remove old drag-target

zhaoyao91:
remove old drag-target to prevent a lot of drat-target after re-init several times.

### DIFF
--- a/js/sideNav.js
+++ b/js/sideNav.js
@@ -18,6 +18,9 @@
           menu_id.css('width', options.menuWidth);
         }
 
+        // Clear old Touch Area
+        $('div.drag-target').remove();
+
         // Add Touch Area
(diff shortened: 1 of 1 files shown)


zhaoyao91:
remove old drag-target to prevent a lot of drat-target after re-init several times.

tomscholz:
@DanielRuf This looks good to me, but @zhaoyao91 Is this an issue that happens a lot?

DanielRuf:
Could be simplified. We could check with `.lenth` if the element already exists and append it if not.

But this here works too and also looks good to me.

acburst:
This has been taken care of in the sidenav rewrite for v1

DanielRuf:
@acburst can you link the commit which fixes it?

PR #2668, opened 2016-01-16
Fixed jQuery Compatability Bug with Scrollspy

### LINKED ISSUE #2647: Scrollspy Compatibility
Your scrollspy plugin only supports jQuery 2.1.x as far as I've tried. jQuery 1.x is a no go, same with 2.2.x.
I would assume it was something specific to jQuery 2.1.x.

Technically, the bug resides in the linking, that jQuery does not recognize how you through the linking at it - a[href=#step] - in any other version that not 2.1.x so maybe something they changed only there, and then reverted the change on 2.2.x.

tcyrus:
Should fix #2647. Changed all jQuery selectors from `$('a[href=#id]')` to `$('a[href="#id"]')`.

### DIFF
--- a/js/scrollspy.js
+++ b/js/scrollspy.js
@@ -71,7 +71,7 @@
 			bottom = top + jWindow.height();
 
 		// determine which elements are in view
-//        + 60 accounts for fixed nav
+		// + 60 accounts for fixed nav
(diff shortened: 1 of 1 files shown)


Dogfalo:
This has been fixed in 5c792c0966a36594bd4e71570ec9a34381613f30.

PR #2672, opened 2016-01-18
Fix bug when calling material_select() on a class selector multiple times

markibanez:
Encountered a bug when calling material_select() on a class selector when the said function has already been previously called. 

Example: 
$('.mselect').material_select();

I observed that the generated div container also gets the original classes of the element, so when the function is called again, generated elements are cloned.

This change just checks if the element is a select element and if it has a class initialized.


acburst:
All plugins should now properly destroy itself when called multiple times, in v1.

PR #2696, opened 2016-01-23
renamed jQuery import to match module name.

gazzer82:
Prevents incompatibilities with web pack and other packagers.


Dogfalo:
Please modify the uncompiled file if you want this to be mergable.

Dogfalo:
You didn't modify the source file so this can't be merged. However this is now fixed so don't worry about it.

PR #2704, opened 2016-01-24
Removed bad showcase link

SwarShah:
Removed becuase of domain has expired.


tomscholz:
If you want to submit your project to our showcase, drop us a mail

PR #2707, opened 2016-01-25
fix: Widen .picker__select--year.browser-default

charlesponti:
Widen .picker__select--year.browser-default to 26% from 25%. At 25%, the '6' in 2016 is cut in half. At 26%, the '6' is show in full with equal padding on both sides of '2016'


Dogfalo:
Thanks! I think it was an OS issue since not all scrollbars are the same width across different platforms

Issue #2614, opened 2016-01-07
On mobile phone screen, div with class 'drag-target' close right part of 'slide-nav' menu


asbelov:
Hello,

On screens with small resolution (mobile phone screens), when showing menu 'slide-nav', you set width of div with class 'drag-target' to 50% 
f.e. in js file: dragTarget.css({width: '50%', right: '', left: 0});

In this case, div with class 'drag-target' close right part of menu 'slide-nav' and I can't use elements, located on a right part of 'slide-nav' menu.
Please, correct this, it is very important for me.

Thank you for you work

fega:
Close in favor of #2715

PR #2715, opened 2016-01-29
set drag target width so it doesn't overlap side nav

kielni:
On small screens, the side nav may be more than 50% of the screen.  In this case, the drag target overlaps the side nav and is on top of it, so clicking anywhere in this (invisible) area closes the nav.

Example: 
for screen width = 360px with side nav = 240px
right side side nav extends from 120px to 360px
drag area is 180px (50% of 360px); it covers the left-most 60px of the side nav

screen shot 2016-01-27 at 3 28 40 pm

related issue: https://github.com/Dogfalo/materialize/issues/2614


alejo-moreno:
What does this invisible drag-target element do anyway? I'm having the same issue and i could remove it or put it behind with z-index. What would you recommend?

hernanvicente:
+1 To merge and fix this problem, thanks @kielni

alejo-moreno:
how come in the sideNav demo this doesn't happen? the drag-target still is 50% and everything is the same (i compared both css files, demo's and mine), but in my app i have the problem, the side nav closes but in the demo the side-nav remains as if it was behind

Sandbakk:
This issue is still there. I use a direct link to https://cdnjs.cloudflare.com/ajax/libs/materialize/0.97.7/css/materialize.min.css and https://cdnjs.cloudflare.com/ajax/libs/materialize/0.97.7/js/materialize.min.js.

drag-target-wrong

weinde:
Has the issue been resolved? I'm also experiencing this issue on my project...

cactushead:
Had the same problem with the drag-target overlapping the right hand side of the slide-nav menu despite the z-index being apparently correct.  The demo website also seems to have this problem.

Decided to find a work around as setting the width to 50% and right aligning it seemed to be the culprit.

On about line 2672 of materialize.js you will see:-

`if (options.edge === 'left') {`
`   $dragTarget.css({width: '50%', right: 0, left: ''});`

 I changed this to:-

`if (options.edge === 'left') {`      
`$dragTarget.css({width: '100%', right: 0, left: options.menuWidth ,'background-color': 'rgba(100, 255, 0, 0.3)'});`

The only changes I needed to make was setting the width to 100% and the left position to move over the width of the side menu.
I added a background-color so I could easily see it working - but I will obviously remove this for the final website.

PR #2718, opened 2016-01-29
Making tabs ignore invisible li

Moykn:
When an li is hidden, and is the first item on ul, the indicator size is miscalculated, leading to a weird behaviour. A hidden li also mess with the $index value that is used to calculate indicator's edge.


tomscholz:
Your pull request is very messy. Please open a new one and please try not to commit compiled files like materialize.js/css and html files

PR #2721, opened 2016-01-29
scrollspy selector issue

vrtt:
I was getting following error with scrollspy and this fixed the issue.

Error: Syntax error, unrecognized expression: a[href=#id-here]


Dogfalo:
Merged a similar fix here: #2764.

PR #2729, opened 2016-02-01
support fullscreen modal

zhaoyao91:
support full screen modal


zhaoyao91:
it seems will not clear the properties properly. I will consider how to implement it better

Issue #2530, opened 2015-12-22
Icons Broken After Upgrade


douggard:
I upgraded to 97.4 to fix the materialboxed problems. This ended up breaking the icons. Particularly the mdi-social-person. You can see the problem at the bottom of this page: http://materializecss.link/icons.html

The code previously used was: 

    <i class="mdi-social-person"></i>

The fix was including the Google icon font replacing the icon code with something like this:

    <i class="material-icons">person</i>

Not sure if this is something you want to support.

Dogfalo:
The Google icon font had been the official way we include fonts for several
releases now

IPWright83:
I've noticed this too - having re-downloaded them from Google they're fixed, but we're currently including materialize in our project using Bower. This means that whenever we produce a build, we're going to have broken icons :-1:

Dogfalo:
The old icons have been removed now.

PR #2731, opened 2016-02-01
Fixed material icons

IPWright83:
The material icons weren't working - so I've replaced the font files which seems to fix the problem. Unfortunately because I'm using Bower I can't simply replace the font's on my system, because they keep getting overwritten by the broken ones.


IPWright83:
I believe this fixes https://github.com/Dogfalo/materialize/issues/2530 - If there's a different approach for including the material icons maybe instead they can be excluded from materialize?

Dogfalo:
The old font files are deprecated and will be removed in the next version

PR #2736, opened 2016-02-02
use input and propertychange events

zhaoyao91:
to be more compatible, especially for ios 8.


Dogfalo:
Please explain better what this does / fixes

Issue #2118, opened 2015-10-04
Change chip code to allow custom close buttons


k4kfh:
I think it would be useful to have the option of using any HTML element as a close button for a chip. With the current implementation, any of the Materialize icons that you put inside a chip automatically become a close button. This is handy if that's all you want to do, but if you plan on having an icon in there for another reason, it's highly annoying to have to modify things every time you want to do that. Also, you might want to have another element (an image, for example) close the tag. I have both of these use cases in my project, so I made a small modification to `materialize.js` to make it possible.

On line 3284, I changed it so that the line is this:

`$(document).on('click.chip', '.chip .close-chip-btn', function (e) {`

Rather than the original. This worked well for me (despite my limited knowledge of jQuery) so I figured I'd see if anyone else thought it was a good idea. If I'm not contributing this suggestion "properly" you'll have to pardon me, I'm relatively new to GitHub so I'm "flying blind" here. Correct me if I did anything stupid. :stuck_out_tongue_closed_eyes:

Sajonara:
+1

Could you explain what this change does and if one could override the normal behavior when I'd add this to the footer of a document?

k4kfh:
If you don't know, the default behavior is for any button (anything of the .material-icons class) to function as a close button. You can call a function with ``onclick`` with the button, but regardless of what the function does the chip will disappear. I needed the chip to only vanish in certain instances, so this wouldn't work.

The change I outlined above makes it so that any element inside the chip that belongs to ``.close-chip-btn`` will function as a close button, but anything that doesn't will function just like it would anywhere else. If you need more information on the default behavior, see [the MaterializeCSS documentation on Chips.](http://materializecss.com/chips.html)

For example, if I used this markup:

```html
<div class="chip">
    Some Text
    <i class="material-icons">close</i>
</div>
```
With the default behavior, the button would make the chip vanish. With my behavior, this would do nothing. If that seems useless, take the example below:

```html
<div class="chip">
    This is not a closable chip
    <!---This icon is an "add" icon from google's material design icons-->
    <i class="material-icons" onclick="someJSfunction()">add</i>
</div>
```

With the default behavior, it would close the chip (just like the example above) because anything inside it that belongs to ``.material-icons`` closes the chip. With my behavior, however, this would not close the chip. It would execute ``someJSfunction()``, but the chip would remain on the page.

Now take this example:

```html
<div class="chip">
    <i class="material-icons left" onclick="someJSfunction()">add</i>
[...]

PR #2761, opened 2016-02-08
Close chips only if icon is 'close'

fabienb4:
Prevent closing a chip on icon click if icon isn't a 'close' icon.

This allow the use of icons instead of images without them removing chips on click.

See #2118


acburst:
I merged in a large chips PR so this is now solved.

PR #2764, opened 2016-02-08
Fix selector of href.

Rompei:
I fixed the bug of selectors of href in jQuery used in `scrollspy.js`.  
From jQuery 1.5, "a" selector must have " or ' (ref: http://stackoverflow.com/questions/5818555/uncaught-exception-syntax-error-unrecognized-expression-href-example-html), but the code didn't have these, so this error was occurred.

```
Uncaught Error: Syntax error, unrecognized expression: a[href=#software]
```

### DIFF
--- a/js/scrollspy.js
+++ b/js/scrollspy.js
@@ -176,7 +176,7 @@
 			elements.push($(element));
 			$(element).data("scrollSpy:id", i);
 			// Smooth scroll to section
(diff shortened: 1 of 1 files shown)


themakshter:
Would there be any way to add any sorts of tests for this?

Dogfalo:
LGTM , thanks!

isaachinman:
@Dogfalo When will the npm package be updated with this fix?

PR #2766, opened 2016-02-08
Dropdown - Stop Event Propagation Support

rthomson83:
Added support to stop event propagation on origin element click handler.

My use for this was adding a context button (with dropdown) to an accordion header. Without this it triggers the click handler on the accordion header.


Dogfalo:
Thanks @rthomson83 !

PR #2768, opened 2016-02-09
Split forms sass partial in subpartials

soniabrami:
- checkboxes, file-input, input-fields, radio-buttons, range, select and switches


Dogfalo:
Awesome, thanks!

PR #2770, opened 2016-02-10
Added cursor:pointer to card-action > a

ranneyd:
The 'a' tags in the card-action should have cursor: pointer because they
are meant to function as buttons.


Dogfalo:
`<a>` tags are by default cursor: pointer.

ranneyd:
That's not true. `<a>` tags have `cursor: pointer` IF they have an href. If they don't have an href, they have the default pointer.

PR #2796, opened 2016-02-15
Add trigger for tab click client event

f1ana:
Add a trigger for the tab click event.  This is useful in cases where elements within the tab div need redrawing or resizing after the display attribute has been changed.

### DIFF
--- a/js/tabs.js
+++ b/js/tabs.js
@@ -101,6 +101,9 @@
           $indicator.velocity({"right": $tabs_width - (($index + 1) * $tab_width)}, {duration: 300, queue: false, easing: 'easeOutQuad', delay: 90});
         }
 
+        // Trigger for any post-tab change handling
+        $('ul.tabs').trigger('tab_change', $active);
+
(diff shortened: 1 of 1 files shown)


chriskobar:
How would this be implemented in markup? Say, for example, I want to re-render a calendar (e.g. fullCalendar) using a "render" method. How would I write that using your trigger?

f1ana:
This should do the trick:

```javascript      
$('ul.tabs').bind('tab_change', function(event, content) {
  if (content && content.hash) {
    console.log('now on tab ' + content.hash);
  }
});
```

f1ana:
With the placement of the trigger in tabs.js, your code has the guarantee that it will execute after the tab section itself has finished coming into view.

cigzigwon:
I'm not using jquery at all but would like to be able to supply an onClick option to M.Tabs.init. i'm new to this library but it doesn't make much sense to put a handler onShow because do things upon pre-render and I think this would be much more helpful when working in React. Any suggestions?

PR #2797, opened 2016-02-15
Add '-webkit-transform' property for safari

abouthiroppy:
Hello :)

`transform` did not work in safari(verion: 8.0.4)


Dogfalo:
This isn't necessary, the prefixes are already being added.
https://github.com/Dogfalo/materialize/blob/master/bin/materialize.css#L9389

PR #2798, opened 2016-02-15
Thumbnail support on "materialbox" image.

paulux84:
Added support to "data-original" on img element to use thumbnail.
If for example you add an image element like this ``
the src attribute is used as thumbnail source and data-original as original image url.
On click "src" become "data-original"

### DIFF
--- a/js/materialbox.js
+++ b/js/materialbox.js
@@ -16,14 +16,53 @@
       var outDuration = 200;
       var origin = $(this);
       var placeholder = $('<div></div>').addClass('material-placeholder');
+      placeholder.css("display","table");
+      var progressWrapper = $("<div style='position: relative;bottom: 12px;'></div>");
+      var progress = $('<div></div>').addClass('progress');
+      var indeterminateProgress = $('<div></div>').addClass('indeterminate');
+      progress.append(indeterminateProgress);
+      progressWrapper.append(progress);
+
       var originalWidth = 0;
       var originalHeight = 0;
       var ancestorsChanged;
+      var imageChanged = false;
       var ancestor;
       origin.wrap(placeholder);
 
 
       origin.on('click', function(){
+        if (!imageChanged && origin.data('original') !== ""){
+          progressWrapper.insertAfter(origin);
+          origin.load(function(){
+            onImageClick();
+            progressWrapper.remove();
+          });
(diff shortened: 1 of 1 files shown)


js/materialbox.js
```diff
@@ -16,14 +16,53 @@
       var outDuration = 200;
       var origin = $(this);
       var placeholder = $('<div></div>').addClass('material-placeholder');
+      placeholder.css("display","table");
```
paulux84: This is necessary to ensure that progress bar have the same width of image

Habeeb-rahman: Haii

ivellios:
Is this one going to the version or there's another solution? I am trying to achieve something similar and do not know how.

Habeeb-rahman:
May I know who are you?

jam1e:
Any answer for @ivellios? As I am also trying to achieve the same thing.

Habeeb-rahman:
Who are you?

ivellios:
I tried to make a workaround with a current version by adding hidden materialboxed elements and triggering click manually when someone clicks thumbnail. Problem is - I see no option of overriding materialbox hiding to hide the image again (once enlarged image is clicked/page scrolled/background clicked it should be resized, but original image has to be display:none again so only thumbnail is left). 

I looked into source and there seems to be no event callback one could hook into and override close behavior anyway.

ivellios:
I tried to use this file again today, but...
1) There seems to be a bug - src of the `img` element is changed to original on click, but never changed back to thumbnail image. As a result, the image shrinks down and replaces thumbnail with respect to scale only. 
2) You have a bug with ancestorchanged variable - you change it's css, but when it stays undefined (in case while loop never finds it) it throws an error - this probably will be fixed simply by `if (ancestorchanged !== undefined)` but I am not sure what this one is for.

The animations with velocity seem odd here and I am not quite sure how it works. Tried to add `src attr` change back to thumbnail, but whatever I tried, it ended with odd behaviour.

Anyone else tried to fix this one? I will keep on trying since it is important feature for me.

ivellios:
@jam1e, @paulux84 - I managed to fix some issues and create code up to date with current version of materialize. I mostly used this solution, but with little changes. My PR is under: https://github.com/Dogfalo/materialize/pull/4085

Hope that helps you :)

PR #2803, opened 2016-02-16
sass selector syntax fix

a17:
Default IDE inspector shows error with this sass syntax.


Dogfalo:
Thanks for the catch.

Issue #2635, opened 2016-01-11
about tabs adjust to  small resolution ratio(320px) the header scroll bar


UniqueCrownClown:
About components Tabs,in 320px screen,the header scroll bar appear,feels it is not nice,how to full of the screen

Frzy:
I've noticed this too.  Not just on 320px screen but as I grow or shrink the window I noticed a horizontal scrollbar flicker on the tab headers. See screenshot below.

**Browser**
![screen shot 2016-01-22 at 11 32 46 am](https://cloud.githubusercontent.com/assets/1053425/12519528/b09a6ece-c0fc-11e5-8e63-ef1efeb06705.png)

**Element**
![screen shot 2016-01-22 at 11 33 08 am](https://cloud.githubusercontent.com/assets/1053425/12519538/b824775c-c0fc-11e5-81ee-3702b6046341.png)

When Resizing the window a little scroll goes away

**Browser**
![screen shot 2016-01-22 at 11 33 27 am](https://cloud.githubusercontent.com/assets/1053425/12519550/cdf93fea-c0fc-11e5-9004-b3a927df1bd1.png)

**Element**
![screen shot 2016-01-22 at 11 33 35 am](https://cloud.githubusercontent.com/assets/1053425/12519554/d06d5554-c0fc-11e5-85a0-134c1c376d0d.png)


2 work-arounds.

To never show scroll bar on tabs add the following in _tabs.scss from
```css
.tabs {
    ...
    overflow: hidden;
}
```

or update the following lines in the tabs.js file

```js
// Line 41
$indicator.css({"right": $tabs_width - (($index + 1) * $tab_width)});
// Line 51
$indicator.css({"right": $tabs_width - (($index + 1) * $tab_width)});
// Line 95
$indicator.velocity({"right": $tabs_width - (($index + 1) * $tab_width)}, { duration: 300, queue: false, easing: 'easeOutQuad'});
// Line 101
$indicator.velocity({"right": $tabs_width - (($index + 1) * $tab_width)}, { duration: 300, queue: false, easing: 'easeOutQuad'});
```

to

```js
// Line 41
$indicator.css({"right": Math.max(0, $tabs_width - (($index + 1) * $tab_width))});
// Line 51
$indicator.css({"right": Math.max(0, $tabs_width - (($index + 1) * $tab_width))});
// Line 95
[...]

PR #2809, opened 2016-02-17
Changed tabs and tab width calculation from methods that cause rounding.

mnannola:
#2635 is being caused by the tab indicator sometimes having a `right:-1px` being set.  This happens when the width of the tabs container has a decimal.  Because the width() and outerWidth() jquery methods round the return value, `$indicator.css({"right": $tabs_width - (($index + 1) * $tab_width)})` would sometimes result in `-1`, causing a horizontal scroll bar to appear.  

Replacing those methods with getBoundingClientRect() fixes this issue, as that method will always return exact pixel values.

### DIFF
--- a/js/tabs.js
+++ b/js/tabs.js
@@ -11,8 +11,8 @@
 
       $this.width('100%');
       var $active, $content, $links = $this.find('li.tab a'),
-          $tabs_width = $this.width(),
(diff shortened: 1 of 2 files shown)


kboucher:
+1

PR #2815, opened 2016-02-19
Fix typos

sajadtorkamani:


Dogfalo:
The change you made in ` jade/page-contents/media_content.html` is not correct. There is supposed to be a comma there. Please revert that change and I will merge this. Thanks.

Dogfalo:
Thanks!

PR #2820, opened 2016-02-19
Fix leak from P.$root reference on date picker

gsilis:
During teardown, close() is called which schedules a function to hide
the date picker. Close accepts a teardown argument so that it can
nullify the reference it has to its root dom node.

Master version:
screen shot 2016-02-19 at 16 00 24

Fix branch version:
screen shot 2016-02-19 at 16 00 37


Dogfalo:
Have you reported this upstream? https://github.com/amsul/pickadate.js/blob/master/lib/picker.js#L194

gsilis:
I have not yet, but I will! Thanks for pointing me there.

PR #2825, opened 2016-02-21
Prevent Multiple Character Counters on Same Element

sosoHero:
This prevents multiple character counters from being created for the same element if the $.characterCounter method is called against an element that already has a character counter.


acburst:
Thanks!

PR #2861, opened 2016-03-01
Fix invalid value for letter-spacing in .picker__day--infocus class

kboucher:
`letter-spacing: -3;` is invalid CSS. There should be a unit specified:

https://developer.mozilla.org/en-US/docs/Web/CSS/letter-spacing
https://developer.mozilla.org/en-US/docs/Web/CSS/length

`px` seems to have the desired affect on the layout.


tomscholz:
Fixed in https://github.com/Dogfalo/materialize/commit/2e9e473a3a829cfa9fd3185253c4f65716a246a4

Issue #1702, opened 2015-07-09
Tooltip close delay option


rvanlaak:
We would like to use the tooltips for interaction, because a dialog is too much for the interactions we have.

Unfortunately the tooltip does not stay open on `mouseleave`. We currently use the following jQuery plugin to archieve that: https://github.com/do-m-en/jquery-ui-mouse-over-tooltip

The following improvements would be nice:
- [x] allow html
- [ ] option to set the close delay
- [ ] do not hide the tooltip on hover

toandocoffee:
Hi,

How about this function?

Thank you

tomscholz:
Related #2864

Dogfalo:
in v1-dev tooltips can be opened and closed programmatically

PR #2864, opened 2016-03-02
Implement a tooltip timer (usefull for mobile devices)

yaennuuh:
On iPhone and iPad the tooltip didn't disappear after opening.
Therefore I've developed a fork containing a tooltip timer.
This timer is optional.

### DIFF
--- a/dialogs.html
+++ b/dialogs.html
@@ -171,14 +171,15 @@
         <pre><code class="language-markup">
   &lt;!-- data-position can be : bottom, top, left, or right -->
   &lt;!-- data-delay controls delay before tooltip shows (in milliseconds)-->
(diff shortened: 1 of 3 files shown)


pwittke:
Any news on this PR? Would be verry helpful to be able to define a delay that removes a tooltip.

yaennuuh:
> Any news on this PR? Would be verry helpful to be able to define a delay that removes a tooltip.

I think it won't be merged since the whole tooltip.js changed.
Check this @pwittke : https://github.com/Dogfalo/materialize/issues/1702#issuecomment-318844098

behuti:
This would be nice on tooltip options!!!! I need this on responsive devices.

behuti:
@Dogfalo what do you think about this? I tried to delete the tooltips "programatically", but no solutions like this at the moment.

Please consider this improve.

PR #2868, opened 2016-03-02
Resolved Google Drive Picker API conflict

AkshayChordiya:
Hey,

The `.picker` class conflicted with the Google Drive Picker API which caused the Picker Dialog to not load and open and this problem  #is really hard to debug. Hence I fixed this issue by modifying the `materialize.min.css` on my side and tested by replacing `.picker` to `.picker-material` which worked great.

### DIFF
--- a/sass/components/date_picker/_default.date.scss
+++ b/sass/components/date_picker/_default.date.scss
@@ -1,16 +1,16 @@
 /* ==========================================================================
-   $BASE-DATE-PICKER
+   $BASE-DATE-picker-material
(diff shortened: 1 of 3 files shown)


prachi708:
How did you edit materialize.min.css file when I am trying to do same it is showing a non editable editable file. 
Any suggestion will be greatly helpfull.

prachi708:
I tried to override it by creating another css file in client/style which override default.date.scss, but still it is not working.
Please suggest something

PR #2870, opened 2016-03-02
Correct jquery require

danny-andrews:
The jquery require should be lowercase. The way it is, it cases a warning in when [compiling with webpack](https://github.com/webpack/webpack/issues/382).


Dogfalo:
Looks good, thanks!

PR #2885, opened 2016-03-07
Check element exist before appending

Polorouge:
For both leanModal and sideNav I had a trouble with my backbone project. It appends elements everytime it's called ... So before appending I add a check to see if it exists. If it exists then use it instead of appending.


kosssi:
👍 
I have the same error

tomscholz:
leanmodal has been deprecated

PR #2887, opened 2016-03-07
Fix for indicator with dynamic tab sizes

WilbertE:
This update makes the indicator (active tab underline) work when tabs have different widths.

The fix measures the width of the active tab and the offset and applies it to the indicator


nathankoop:
Thanks @WilbertE

Dogfalo:
Fixed in 8be8783

Issue #2688, opened 2016-01-21
datepicker is not working with jquery v2.2.0


kamrankr:

samrm111:
+1

niwasmala:
+1

This is annoying when you don't know what could be wrong, just doing bower update and voila, datepicker not working anymore. Currently I'm using jquery v2.1.4 so datepicker still works.

throrin19:
+1, I test with the official pickadate library and it works fine but i have not the materialize theme :(

vhaarr:
+1

Good thing I had the wits to check the issue tracker on github or I would have gone crazy trying to find the problem.

Would be nice if materializecss updated to at least pickadate.js 3.5.5 regardless, so that we get the "closeOnSelect" option.

Dogfalo:
Fixed by #2889

PR #2889, opened 2016-03-08
use $.fn.focus instead of HTMLElement.prototype.focus

ramybenaroya:
This is a fix for https://github.com/Dogfalo/materialize/issues/2688


Dogfalo:
Thanks for tracking down the root of the issue!

PR #2891, opened 2016-03-09
Fix issue with the dropdown on mobile

kevinbrink:
- In some circumstances, attempting to scroll within the dropdown on mobile
  would cause the dropdown to disappear immediately, making it impossible to use
  properly. In my particular case, I was re-styling the dropdowns inside a date picker (http://amsul.ca/pickadate.js/, because it offers a few required features that the materialize date picker doesn't) in order to use the materialize dropdowns.

### DIFF
--- a/js/dropdown.js
+++ b/js/dropdown.js
@@ -201,7 +201,7 @@
           // If menu open, add click close handler to document
           if (activates.hasClass('active')) {
             $(document).bind('click.'+ activates.attr('id') + ' touchstart.' + activates.attr('id'), function (e) {
-              if (!activates.is(e.target) && !origin.is(e.target) && (!origin.find(e.target).length) ) {
+              if (!activates.is(e.target) && (!activates.find(e.target).length) && !origin.is(e.target) && (!origin.find(e.target).length) ) {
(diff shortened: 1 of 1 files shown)


neFAST:
This works but won't close the dropdown.
I think you need to add this:
https://github.com/Dogfalo/materialize/pull/2953

kevinbrink:
@neFAST : Mind expounding a bit? In my tests, it seems to work fine, and closes the dropdown when I need it to. I see the commit you linked to, as well as the additional code added to that. 

When do we **want** the dropdown to close that my commit doesn't allow for?

neFAST:
Well, if you look at the manual example from a desktop computer, the dropdown will close whenever you click an item: http://materializecss.com/dropdown.html
With your patch it will break this behavio on both desktop and mobile (and fix the scrolling issue on mobile which is nice :+1: )

kevinbrink:
Ah, ok, I see what you mean now. Our particular application only encountered that particular issue in one area of the app. I see it now though!

soryy708:
@kevinbrink There are merge conflicts

PR #2900, opened 2016-03-11
Test with current LTS and stable node versions

revington:
Current LTS node is 4.4 and stable is 5.8. `node` is a pointer to current stable version.
node-archiver, a dependency, has not been updated in 2 years. When I install materialize in my computer I get this warning from npm 

```
npm WARN deprecated graceful-fs@3.0.8: graceful-fs version 3 and before will fail on newer node releases. Please update to graceful-fs@^4.0.0 as soon as possible.
```

Also, some dev dependencies are pretty old.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
   - "0.12"
-  - "4.3"
+  - "4.4"
   - "node"
 before_install:
   - npm install -g grunt-cli


alqu:
Still throws a warning on install:
> deprecated graceful-fs@3.0.8: graceful-fs v3.0.0 and before will fail on node releases >= v7.0. Please update to graceful-fs@^4.0.0 as soon as possible.

I tested it with the current version 0.97.6

alqu:
Hm. Looks as if the dependency is coming from node-archiver - sorry, will file a bug there.

tomscholz:
See: https://github.com/Dogfalo/materialize/pull/4634

PR #2906, opened 2016-03-13
Propose more files to sass

carusogabriel:
Pull requesting some new divisions on sass folder, making easy for developers comment that css proprieties they doesn't need


NitroBAY:
I agree

tomscholz:
https://github.com/Dogfalo/materialize/pull/4006

PR #2909, opened 2016-03-14
Fix and test for issue #2907. 

### LINKED ISSUE #2907: Options after optgroup looks like they are in optgroup
materialize-optgroups

There is no such effect in browser select control.

There is a codepen to compare behavior of Materialize and browser select: [http://codepen.io/anon/pen/jqMzEa](http://codepen.io/anon/pen/jqMzEa)

SeriousDron:
Fixes #2907. <option> element after <optgroup> should not looks like it's in <optgroup>
Solved by adding class to options which are members of optgroup. Indent is added only to options with that class.


acburst:
Thanks!

PR #2919, opened 2016-03-16
Character Counter: allow and prefer to use data-maxlength attr instead of length

Anubisss:
The length input attribute is invalid because it doesn't exist.
http://www.w3schools.com/tags/tag_input.asp

I think data-maxlength (or data-length?) should be used (at least preferred to use).


acburst:
We now use data-length

PR #2920, opened 2016-03-17
Fix malformed bower.json

mike-north:
Bower throws an `EMALFORMED` when attempting to install recent materialize SHAs (i.e., `ffb9783b388912fbe1245a1362f0b1ace03d46bd`).


Dogfalo:
Thanks @mike-north

PR #2931, opened 2016-03-20
Fix pagination links to fill entire li tags

shinbuntu:
The pagination links don't fill entire li tags. Although the waves effect event is fired on click, the link is not.


acburst:
Thanks!

PR #2932, opened 2016-03-20
scrollfire now accepts standard callbacks

francesco-strazzullo:
Scrollfire can now accepts also standard callback and not only in the string format


acburst:
Fixed

Issue #2916, opened 2016-03-15
¿Bug in slider?


superbeast-dev:
Hi, I think I've found a bug in the slider. When you create a slider, and just put an image shown right. But if you use a mobile device, and that one slide you try to pass the finger as if they had more... the slider is lost and have a nice white background.

acburst:
Merged in PR

PR #2942, opened 2016-03-23
Fix for #2916

paulina-pacheco:
If there is a single slide in slider and swipe left or right, return to original position.
fix for #2916


acburst:
Thanks!

PR #2947, opened 2016-03-23
Update picker.js

demonio:
Now... when you close datepicker, it STOP scroll to bottom.


Dogfalo:
I haven't seen any scrolling behavior. So I don't think this fix is necessary. Can you explain when this is happening to you?

demonio:
This occurs when scrolling in the browser because the content occupies more than one page and using the datepicker in a form located above, to close the datepicker this makes it Toggles between down completely, the document.

Dogfalo:
This isn't reproducible by me and nobody else has complained of this issue.

PR #2953, opened 2016-03-24
fix bug: cannot click item on ios devices

zhaoyao91:
1. only close dropdown if click or touch is not on activates
2. only close dropdown with click on activates(ensure the click on activates will work)


4tron:
This works for me, Have not tested it on an actual device but the issue was the same in the chrome developer tools. Should be added to main thanks @zhaoyao91

OG84:
+1

tomscholz:
I don't have an ios device. Can anybody confirm this was/is an issue?

PR #2957, opened 2016-03-25
Event on collection item dissmissed

david-campos:
Simply added a trigger of an "dissmissing" event on JQuery when a dismissable item of a collection is about to be dismissed.
I think many people will appreciate this.


Dogfalo:
transitions.js has been removed.

PR #2958, opened 2016-03-25
moved removeMenu() to .sideNav('hide')

subins2000:
Currently `sideNav("hide")` will trigger a click on overlay which don't work if the user wants to disable the overlay. There is a removeMenu() function inside `init`. Why not move it to `hide` and call it inside `init`.


acburst:
Plugin rewritten

PR #2959, opened 2016-03-25
Fixed Textarea Resizing Bug

subins2000:
Currently, Materialize resizes textarea when typing is started. What if I want to set a default height to textarea and when the text overflows that height, textarea should resize. Also, when the content becomes null again, textarea should resize to the default height instead.

Instead of setting the height of textarea default on typing start, Materialize should store the height and then resize textarea according to the value length and the original height of the textarea.

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -144,14 +144,31 @@
         hiddenDiv.css('width', $(window).width()/2);
       }
 
-      $textarea.css('height', hiddenDiv.height());
+      /**
+       * Resize if the new height is greater than the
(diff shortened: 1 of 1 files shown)


acburst:
Do you have an issue or an example of the bug youre describing?

subins2000:
Bug : https://jsfiddle.net/subins2000/2qad9oyg/

AmanGupta-2210:
+1

fega:
+1

acburst:
Thanks!

PR #2961, opened 2016-03-25
require 'jquery' lowercase

mischkl:
This seems to have been fixed in the `initial.js` source but not in `dist/js/materialize.js` - jquery is required with improper casing, causing a warning when resolving the package: "There is another module with an equal name when case is ignored. This can lead to unexpected behavior when compiling on a filesystem with other case-semantic."


Dogfalo:
Fixed already in 27c19140ef984e3b07384520e5f0aa93d3c26fae

PR #2975, opened 2016-03-30
Adds an override flag to allow for evaluation of items on page load

marioharper:
Currently if a trigger point is within view on page load it will not execute until the user scrolls. Adding this flag allows the user to call scrollfire like:
`Materialize.scrollFire(options, true);`

 This way if they know a trigger point may possibly appear above the fold it will still get fired.


rubendinho:
+1

PR #3004, opened 2016-04-05
Added javascript-defined tooltip positioning

Roundaround:
Remake of Pull Request #1628 from a new fork/branch that is mergable with Dogfalo's master.

Original Comment:
Added default tooltip position and linked position to look at the javascript parameters when not defined in the attribute.

The general reason for this change is to reduce unnecessary bloat in the DOM when trying to use the Materialize tooltips. Having to specify a class and at least 2 data-attributes makes an originally small button definition to need line wrapping in some IDEs and code styles. This change makes the data-position attribute optional in the same fashion as the data-delay attribute to possibly reduce this clutter.

### DIFF
--- a/js/tooltip.js
+++ b/js/tooltip.js
@@ -5,7 +5,8 @@
 
       // Defaults
       var defaults = {
-        delay: 350
+        delay: 350,
+        position: 'bottom'
       };
 
       // Remove tooltip from the activator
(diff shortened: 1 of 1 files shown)


acburst:
Thanks! This was refactored from multiple PRs into one.

Issue #2971, opened 2016-03-29
Css search bug


Josepdal:
https://i.gyazo.com/c8bfe30a9424cd779ed2c61dfe934d33.png

acburst:
can you post some code and flesh this out more. These types of issues are very unhelpful because I have no way of diagnosing the problem to fix it.

Josepdal:
@acburst you can watch code here: http://materializecss.com/navbar.html

is wrong in your guide too

Thanood:
@Josepdal Which browser is it that you use?

Josepdal:
@Thanood Chrome, in os x and windows does not working fine

NitroBAY:
Hello Josepdal, sorry for your problem, I had the same on IE (#2903). I sent a pull request, I don't know if Materialize's team will accept it but as I already patch my website you could test if my pull request fix your problem here : https://pharmaciesousprefecture.fr/ :D

AlvaroLQueiroz:
I have the same problem in linux chrome v50. I fixed using:

`nav .nav-wrapper form, nav .nav-wrapper form .input-field{
height: 100%;
}`

acburst:
Merged c129dc1

DatoBula:
@AlvaroLQueiroz thanks, I had same problem and your solution worked for me.

Issue #2903, opened 2016-03-12
[URGENT] search navbar in IE 11 bug


NitroBAY:
In IE 11 the navbar search is totaly buggy (with all zoom). The input field is only in very top of the bar... http://imgur.com/BGo6LdL please note that the bug is present when I use the official website and also when I try the code on my side... I'm waiting the bug to be fix before launching my project in prod.

NitroBAY:
Up

NitroBAY:
I also have this bug in chrome on android ...

NitroBAY:
Still not solved

AlvaroLQueiroz:
I have the same problem in linux chrome v50. I fixed using:

nav .nav-wrapper form, nav .nav-wrapper form .input-field{
   height: 100%;
}

acburst:
Merged c129dc1

PR #3013, opened 2016-04-08
Fix navbar search who wasn't taking full height in all browser

NitroBAY:
Hello, it's a fix to the bug reported by #2971 and #2903. I tested it on IE 11

### DIFF
--- a/sass/components/_navbar.scss
+++ b/sass/components/_navbar.scss
@@ -117,6 +117,8 @@ nav {
   }
 
   // Navbar Search Form
+  form { height: 100%; }
+
   .input-field {
     margin: 0;
     height: 100%;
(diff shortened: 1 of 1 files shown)


Josepdal:
Hey, i tried this, but is necessary to put the form height up to 100% too.

Check it here: http://taligram.org/en/channels so many thanks :smiley:

NitroBAY:
You welcome, as I didn't use form in my search I didn't saw that, I'll commit soon then :D

NitroBAY:
Actually even with form tag surrounding search navbar as proposed by doc it's good for me in IE 11, but I can reproduce the issue of your website as you use search from navbar in a non convential way by removing height 100% from form thus I'll include your modify

Josepdal:
Nice job :) :+1:

acburst:
Thanks!

PR #3025, opened 2016-04-10
Removing width change onModalOpen

carusogabriel:
Removing width change onModalOpen, providing a better user experience on how navbar looks like.
sem titulo


Dogfalo:
The width change is actually keeping the width the same because of the removal of the scroll bar. I think the way it is now is actually good as it will prevent any janky page reflows and won't shift page content.

PR #3031, opened 2016-04-12
Fix sideNav 'initialization' bug

stonedleaf:
Fixed the following bug:
*sideNav runs an 'initialization' event when click is made on the side of the web page. This 'event' is when the sideNav abruptly appears and disappears. This only happens on the first click

**To reproduce the bug:
1) Have a normal sideNav (take note on what edge it is set, for this case, the default edge, 'left', will do)
2) Click on the edge of the window, in this case, the leftmost part of the page.
3) You'll notice a very fast appearance and disappearance of the sideNav.

Note: this is very noticeable in mobile / small screen users

Cause(s):
a) The events for handling touch aren't setting menuOut to its proper value
b) the click event for the overlay doesn't check if there is really an overlay and/or the menuOut variable is true.


acburst:
Thanks for finding and fixing this!

K1N5L4Y3R:
+1

PR #3087, opened 2016-04-26
Make bower.json package name lower-case

jacobq:
Please consider following the convention of using lower-case for package names in bower, npm, etc. I lost a significant amount of time today debugging a strange problem that was ultimately caused by this package using a capitalized name but being referenced in the code by a lower-case one on a  case-sensitive file system.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,5 +1,5 @@
 {
-  "name": "Materialize",
+  "name": "materialize",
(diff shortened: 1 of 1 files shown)


kevgathuku:
This was fixed in #3212
Mind closing this out @jacobq?

Issue #3108, opened 2016-04-30
JS characterCounter() function uses incorrect attribute


Nukeface:
`characterCounter()` function searches `<input>` elements for `length` attribute. However, though it technically 'could' be added to these elements, the [HTML5 docs](https://www.w3.org/TR/html-markup/input.text.html#input.text) state the usage of `maxlength` attribute instead. Could be solved in either or both manners of: set function to default search for `maxlength` or add option to pass function attribute name to search for (e.g. `data-length`)

Dogfalo:
This is done to leave the choice for the developer to enforce a hard limit during typing or to allow typing of the limit while warning the user.

PR #3112, opened 2016-05-02
Add support for HTML5 attribute maxlength for max character count on inputs and textareas

derekrliang:
Adds enhancement for https://github.com/Dogfalo/materialize/issues/3108.

This adds character count support for HTML5 attribute "maxlength", meaning you don't have to have the "length" attribute in order for character count to work. However, if you have both, it will choose "length" over "maxlength".

Also note that "maxlength" is a hard limit on character count so the form doesn't let you enter more characters after that length, so having both "length" and "maxlength" at the same length or "length" more than "maxlength" would be meaningless. However, "length" less than "maxlength" would validate the input still but have a hard limit of "maxlength".

I was thinking about changing "length" to something like "data-max-valid-length", but it would break backwards compatibility with people using "length"...

### DIFF
--- a/forms.html
+++ b/forms.html
@@ -881,6 +881,13 @@
                 <label for="textarea1">Textarea</label>
               </div>
             </div>
+            <br/>
+            <div class="row">
+              <div class="input-field col s12">
+                <textarea id="textarea2" class="materialize-textarea" maxlength="150"></textarea>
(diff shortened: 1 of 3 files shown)


Dogfalo:
See my comment here: #3108

PR #3113, opened 2016-05-02
Fix search .input-field height on Chrome

victorpre:
I was trying to use the Search Bar inside the Navbar component and I was getting the following on Chrome Version 50.0.2661.86, it is even not showing up properly on [Materializecss.com](http://materializecss.com/navbar.html):

screen shot 2016-05-02 at 12 17 06 am

### DIFF
--- a/bin/materialize.css
+++ b/bin/materialize.css
@@ -4184,12 +4184,12 @@ nav ul.left {
 }
 
 nav form {
-  height: 100%;
+  height: inherit;
 }
 
 nav .input-field {
   margin: 0;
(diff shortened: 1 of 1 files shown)


NitroBAY:
Not fixed by #3013 ??

victorpre:
It seems to still be broken on Chrome Version 50.0.2661.86 .. Can anybody update their Chrome and take a look [here](materializecss.com/navbar.html/)?

tomscholz:
Works for me

Issue #3103, opened 2016-04-29
Material Design Icons


n1k1c4:
Arrrgh… T_T

I spent on the search for solutions 3 hours of your life!

Your font icons are not working.

I had to keep the font of the icons locally, and I finished writing styles and took the font from your repository. But nothing worked. Until I took the original font files from the Google repository.

Dogfalo:
The font documentation says this at the top of the page.

On Friday, April 29, 2016, Nikita <notifications@github.com> wrote:

> Arrrgh… T_T
>
> I spent on the search for solutions 3 hours of your life!
>
> Your font icons are not working.
>
> I had to keep the font of the icons locally, and I finished writing styles
> and took the font from your repository. But nothing worked. Until I took
> the original font files from the Google repository.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/Dogfalo/materialize/issues/3103>
>


-- 
Doggy sends his greetings from Mars.

Issue #3033, opened 2016-04-12
CORS Issue with loading WebFont


WillsB3:
Hi,

I've been trying to get Materialize working in my Web App however the WebFont doesn't seem to be loading from the CDN.

For example: http://output.jsbin.com/xulohapoxe/1/?output. As you can see in the console there are errors relating to CORS.

Have I missed something?

Thanks

uok:
seems like duplicate of #2989

Dogfalo:
This problem is resolved now as the font is being hosted on CDNJS at the right path

Issue #2989, opened 2016-04-03
Missing font files on cdnjs (v0.97.6)


uok:
Using the [getting started](http://materializecss.com/getting-started.html) template with cdnjs links causes multiple 404 errors for font files. 
the path should be `/font` and not `/fonts`
e.g. cdnjs.cloudflare.com/ajax/libs/materialize/0.97.6/font**s**/roboto/Roboto-Regular.woff2

aquatix:
Same here (hate to do a +1 comment, but it bites me :) )

lebart:
In the changelog of April the 1st they mentioned the changement in font path. I think they just missed the cdn.

aquatix:
Indeed, but it's still a bug in the hosted version :) Lets hope they can fix it soon, I like to be up-to-date, especially with this nice project.

lebart:
I totally agree with you. But lately the issues fixing and pull merging aren't ... let says regular ;)
That's sad because the community is growing fast and the project is really nice!

Dogfalo:
This issue has been resolved as they are now hosted at the right path on CDNJS :smile:

Issue #2185, opened 2015-10-20
Not all icons are listed


mitar:
For example, [cards page](http://materializecss.com/cards.html) uses `<i class="material-icons right">more_vert</i>` icon, which is not listed [among icons available](http://materializecss.com/icons.html).

It seems this is a much better list (you have to adapt names a bit): https://s3.amazonaws.com/kantan-dresssed-demos/demos/ives/blue_preview/007_icons@fa-lightbulb-o%252F004_mdi_icons.html

SuperDJ:
Maybe you can find some more here: http://www.google.com/design/icons/

mitar:
But naming is unclear from there.

SuperDJ:
If you want to use: `<i class="material-icons right">more_vert</i>` add `<link href="https://fonts.googleapis.com/icon?family=Material+Icons" rel="stylesheet">` in your `head` if you haven't already done so. Then each icon which has a space requires the use of `_`.  If there is something else unclear please let us know

mitar:
Ah, http://www.google.com/design/icons/ dynamically renders the page. This is why searching for "more very" didn't find anything. It does not render the whole page at once. What a bad approach for a reference material.

obouchari:
You can use https://materialdesignicons.com/ instead.

mitar:
Great site. But materialdesignicons.com seems to include icons which do not exist in what Materialize provide? Even if I list just Google icons there are missing icons. At least under the names there. Like `car_wash` is on the list, but it is not available.

obouchari:
MaterializeCss also seems to have duplication in the documentation you need to include the link to MDI from google fonts, but there is already a list of font files included in the framework! sass/components/_icons-material-design.scss

timtribers:
@mitar on Chrome you can use the dev panel to search the dynamic page source at https://www.google.com/design/icons/ - just press ctrl-f (or mac equiv) and search for your keyword.  If you find an element with that name, you can right-click the source and use 'scroll into view' to show it.  But expect you knew that already.

tomscholz:
Please close

PR #3118, opened 2016-05-03
independent support of official google material icons

tnga:
-because application ( for mobile, tablet, desktop) built with web technologies generally not need internet connection to load UI.
-because for development, tests generally are localy run.
-because some independant projects need self-hosting features.

#2989 , #3033 , #3103 , #2185


Dogfalo:
I really don't think this is necessary. Seeing as we do not modify the fonts in any way, they are being updated independently of us, and the fact that Materialize does not require the use of a specific icon font, it makes sense to keep them separate. Developers who require local hosting can download the fonts themselves from Google.

PR #3121, opened 2016-05-03
Add option in ScrollSpy to set the vertical offset

Chrismuzikos:
A fixed offset of 200px is set in ScrollSpy by default and it should be changed by any user according to its website design.

We can now call :

```
$('.scrollspy').scrollSpy({ scrollOffsetFix : 55 });
```

By default, `200px` will be used as hardcoded.


ktamiola:
Excellent! Works like a charm! Thank you!

acburst:
Thanks for this!

PR #3153, opened 2016-05-12
update version of waves.js to 0.7.5

PiotrKlys:
Changed version of wave.js (according to https://github.com/fians/Waves) to the 0.7.5 (newest one).
In my case version 0.7.5 solves problem with unwanted waves during scrolling on mobile devices.


mindr242:
Please apply this, the newer version of waves fixes many problems, one is the above and the other is here #3226.

Issue #3161, opened 2016-05-13
Search bar issue on Chrome


gempain:
Hello,

With Materialize v0.97.2, on Gooogle Chrome version 50.0.2661.102 (64-bit) for Mac OS X Yosemite 10.10.5, the search bar does not render correctly (screenshot taken straight from the official website):

![Image of Yaktocat](http://s32.postimg.org/7h72vdbph/Screen_Shot_2016_05_13_at_12_59_03.png)

Also, notice that on Safari 9.1, the search bar behaves strangely with the carret. When first focused, the carret fills 100% of the height, while when typing it fills half of the height. I think this comes from some `line-height` property somewhere.

First focus:

![First focus](http://s32.postimg.org/bo4re1ww5/Screen_Shot_2016_05_13_at_17_44_30.png)

While typing:

![While typing](http://s32.postimg.org/atj4urmx1/Screen_Shot_2016_05_13_at_17_44_41.png)

This is not new. I had the same problem with Materialize v0.97.0.

If I succeed finding a workaround, I'll let you know.

All the best, and thanks for the amazing work !

wedneyyuri:
I can confirm this. It's working for firefox v37 but not working for Chrome v50

![image](https://cloud.githubusercontent.com/assets/7511692/15249141/7a4d2a8c-18f4-11e6-8d2a-fdbc3eff3ed6.png)

gempain:
@wedneyyuri Thanks for confirming. I just updated my post to add another issue I have found (it's not as bad as the other one, but worth mentioning I think).

acburst:
This issue was fixed and merged c129dc1.

It will be public in the next release or you could download our latest version

wedneyyuri:
@acburst Do you have plans about the next release date?

gempain:
Awesome :) Thanks for the great work !

giuliolastra:
As a temporary fix you can add this style:
`.nav-wrapper, .nav-wrapper form, .nav-wrapper form .input-field{
	height: 100%;
}
`

raphaelsoul:
the demo on the official website is not yet fixed.
![searchbox](https://cloud.githubusercontent.com/assets/8615787/15614019/53d0f978-2471-11e6-9dfc-823b7df8deb2.gif)

gempain:
I think it won't be until the next release.

fega:
Working in the latest version

BjarnBronsveld:
[...]

PR #3165, opened 2016-05-14
Fixed search bar issue on Chrome and Safari

dotjamie:
Referencing https://github.com/Dogfalo/materialize/issues/3161 

This fixes the bug in Chrome and Safari (previous commit https://github.com/Dogfalo/materialize/commit/c129dc17f76b799d7dc68a550db04c8a53bcdae1 solves only Chrome issue)

### DIFF
--- a/sass/components/_navbar.scss
+++ b/sass/components/_navbar.scss
@@ -21,6 +21,12 @@ nav {
     position: relative;
     height: 100%;
   }
+    form {
+      height: 100%;
+    }
+      .input-field {
+        height: 100%;
+      }
 
   @media #{$large-and-up} {
     a.button-collapse { display: none; }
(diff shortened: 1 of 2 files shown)


sass/components/_navbar.scss
```diff
@@ -21,6 +21,12 @@ nav {
     position: relative;
     height: 100%;
   }
+    form {
```
gugerard: Indentation ?

gugerard:
Thanks, it works well for me.

keyscores:
+1 can we merge this?

gugerard:
Actually this issue might be duplicated with : https://github.com/Dogfalo/materialize/issues/3161 which is already merged (more code is changed as part of this PR though).

dotjamie:
@gugerard The original issue mentions a cursor bug in Safari and this addresses that as well as the Chrome fix

acburst:
There is duplicated code in this pull request. Take a look at the whole file, your form and input-field styles are duplicates

PR #3172, opened 2016-05-17
Added 14 Cursor Style Classes

angrytoad:
Added 14 classes that will change the style of the cursor depending on the action you would like to do.

I wanted to add these as I always end up having to write them into projects that are very UI heavy. Seeing as materialize is focused on feedback and material design, It seems natural to include classes which modify the cursor.


angrytoad:
Does anybody else have thoughts on this being added?

quentinHydrobat:
You should add it too docs ^^'
But yeah, great improvement, I'm doing this myself, dunno why i've never done this pull request before.

acburst:
This is not the proper way to alter CSS in this project. Are you familiar with Sass? All our CSS is compiled from sass files.

Also if you are adding something it is usually helpful to make some doc changes as well.

PR #3178, opened 2016-05-19
Dropdown respects disabled attribute

rusekr:
Hello!

This change makes dropdowns respect disabled attribute and dont fire open while clicking or hovering.

Thanks!


Dogfalo:
Disabled is not a valid attribute for anchor tags, you'll find however that it works if you added to a button tag

PR #3187, opened 2016-05-21
New helper

SeyedAliR:
insert helper classes for change direction (ltr & rtl) for make this framework mote usable for all languages


tomscholz:
https://github.com/Dogfalo/materialize/pull/4391

PR #3196, opened 2016-05-23
Fix .browser-default on <ul>

### LINKED ISSUE #2582: Unordered lists show up without any styling
The following means that by default all unordered lists appear without styling:

``` css

ul {
  list-style-type: none; }

```

Perhaps this property should have a narrower focus?

sfcgeorge:
The `.browser-default` class on `ul` didn't work as the list-style-type was being removed on the more specific `ul li` level on line#51

Fixes #2582

### DIFF
--- a/sass/components/_global.scss
+++ b/sass/components/_global.scss
@@ -18,7 +18,7 @@ main {
 }
 
 ul {
-  &.browser-default {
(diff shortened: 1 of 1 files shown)


acburst:
Merged

PR #3203, opened 2016-05-24
adjust the rendering of card's contents

tnga:
Here is a pen to see what have been done: http://codepen.io/tnga/full/NNQmGq/


quentinHydrobat:
+1, but some of your suggestion aren't compliant with material design's spec : https://www.google.com/design/spec/components/cards.html#cards-behavior

tnga:
@quentinHydrobat considering i have understood suggestion you are talking about, it isn't support as "feature" but it's just in case to show we can have fluid behavior with a different structure as example. Structure which only depend on  *html* not on *materialize*. I took in mind the *material design's spec* when i added or modified code. Suggestion like **how about reverse structure** is just a *html* approach to show flexibility, kept with the fix :wink: .

tnga:
@Dogfalo , @acburst maybe there was already discussion about it but since it still a problem, due to the number of waiting pull request here and  the fact that they become old with time, without an assigned label or message to tell if they will be merged or not;  i thing this can discourage contributions to this project.

Hope you will find time and solution. That said, can you please check this pull request so that i can have an idea of it status ? (and maybe others interesting fix and adjustments)

tomscholz:
@DanielRuf Can you check?

Dogfalo:
We'd prefer to keep our cards implementation closer to spec

DanielRuf:
I agree with @Dogfalo, the components should follow the specs.

PR #3223, opened 2016-06-02
Optimisation of the code for less DOM polution

DaedalusDev:
Added "var" in front of $overlay and overlayID to reduce var scope to the function


acburst:
thanks

PR #3242, opened 2016-06-11
Simplify and improve "active" input styling

redbmk:
1) BUG: Prefix looks focused whenever there is input
2) BUG: Labels overlap placeholder/value briefly before JS init
3) FEATURE: Materialize.updateTextFields no longer needed
  --add deprecation warning (eventually remove code, but don't
    break sites that are still using it)
4) FEATURE: Label should remain a placeholder until text is input
5) FEATURE: Validation should happen on input rather than on change


redbmk:
Sorry if this is too much for one pull request, but they were all related changes and seemed easier to do it all in one sweep. If you need, I can separate the changes into multiple Commits or PRs.

I removed the part of the documentation regarding Materialize.updateTextFields as it shouldn't be required anymore after this change. However, to avoid breaking code for others who are already using this I kept the function and just displayed a deprecation warning in the console. This should let devs know to remove the code before a future version removes the function altogether. Also, adding the "active" class is no longer necessary.

redbmk:
It looks like this isn't quite ready actually. It works for pre-filling an input in the DOM, but not when modifying the input using javascript (e.g. `document.getElementById("first_name").value = "John";`). I'm looking into a fix for this that ideally wouldn't require using `updateTextFields`, but I'm actually not sure the performance impact of updating the DOM "value" attribute on input changes (for example React might have other ways of doing this and could cause conflicts).

redbmk:
I also found out this breaks textareas. I'm going to close this pull request for now and possibly come back to this as a series of issues/pull requests to avoid too many changes at once.

PR #3244, opened 2016-06-11
BUG: prefix looks focused whenever there is input

redbmk:
Other elements on the input are the default color when not focused
Modified some samples in test/html/forms.html to play around with


acburst:
Good catch. Thanks!

PR #3245, opened 2016-06-12
BUG: textarea padding

redbmk:
When not using a prefix icon, there is too much padding on the top. This is noticeable when two textareas are side-by-side.

1) With no text, the placeholders line up, but the lines are offset.
2) With text, the labels match, but the written text and the lines are offset.

Here's a code-pen with v0.97.6 showing the bug: http://codepen.io/anon/pen/zBqGPB

I've also included the text-areas in the sample test/html/forms.html for easier testing


acburst:
Thanks!

PR #3246, opened 2016-06-12
FEATURE REQUEST: Leave label as a placeholder when text is empty

redbmk:
The [Material Design spec](https://material.google.com/components/text-fields.html#text-fields-labels) seems open to interpretation about _when_ to float the label:

> When _the user **engages** with the text input field_, the floating inline labels move to float above the field.

I could see "engages" meaning clicking on the text field, or typing something into it. This is the only place I can find in the spec that indicates when to float the text. Google's official libraries seem split on the subject as well. [Polymer's paper-input](https://elements.polymer-project.org/elements/paper-input?view=demo:demo/index.html&active=paper-input) does it on text input, but [Material Design Lite](https://getmdl.io/components/index.html#textfields-section) and [Angular Material](https://material.angularjs.org/latest/demo/input) do it on focus.

Personally, I think it makes more sense to do it on text input. A normal input with a placeholder will show the placeholder up until you start typing, then it disappears. In the case of a floating input it would float instead of disappearing. Having it disappear before you start typing seems inconsistent with standard placeholders.


redbmk:
P.S. introducing these code changes made the bug in #3245 much easier to notice. While they don't technically depend on each other, merging this PR and not the bugfix could raise some issues. The reason I made two separate PRs is because if this one isn't accepted the padding bug could still be fixed.

tomscholz:
Hey there 👋 
```
/materialize/sass/
/materialize/js/ 
```

redbmk:
@Dogfalo was this closed because it needed cleaning up, or because you decided against the idea?

I updated and cleaned up the code in [redbmk:placeholder](https://github.com/redbmk/materialize/tree/placeholder) just in case you want to take a look at it again.

PR #3247, opened 2016-06-12
FEATURE REQUEST: Validate on input and init

redbmk:
Validates now as user is typing
Validates prefilled content as well
Contains minor performance improvements:
- creating a jQuery object once and using that, instead of creating a new one each time
- checking for the `validate` class before making other calculations


tomscholz:
Hey there 👋 
```
/materialize/sass/
/materialize/js/ 
```

Dogfalo:
Validation shouldn't be checked while typing. This would give a bad experience in many cases such as typing in an email. You would see errors while typing in a valid email address

PR #3280, opened 2016-06-19
Add icons-material-design to package.js (meteor) 

NitroBAY:
5afb6000 was said to add support to sass for meteor, which is a great thing but it throw an error immediatly when the materialize.scss was loaded on client because package.js wasn't loading icons-material-design so this fix add it.


Dogfalo:
Thanks!

NitroBAY:
Thanks to you for you awesome framework and to add the sass support on meteor as some asked :D

PR #3282, opened 2016-06-19
Always prevent the default behaviour on clickable dropdowns

shoetten:
Currently preventDefault() is only called when opening the dropdown menu. But unfortunately this is not of much use, because the default (aka visiting the link) _is_ executed, when the dropdown menu is closed again, resulting in an ugly extra history state.

This pull requests always prevents the default behaviour, when using a clickable dropdown menu.


Dogfalo:
Fixed in 1.0 alpha

PR #3283, opened 2016-06-20
BUG : Added missing quote to the slider image

JimmyBillan:
Without quotes in the background-image property, images with parentheses in the url are not displayed.
The style is not valid .


tomscholz:
@DanielRuf @acburst Looks good to me. What do you think?

Dogfalo:
Looks good, but I would prefer this style: `$(this).css('background-image', 'url("' + $(this).attr('src') + '")' );`

tomscholz:
Fixed in https://github.com/Dogfalo/materialize/commit/b1491c37885fc947dd88890dd4d7185619823412

Issue #3286, opened 2016-06-21
Align data-success to the right


fupuchu:
Hi there, is it possible to shift data-error and data-success messages to the right of the label?

Currently it looks like this 
![screen shot 2016-06-21 at 12 43 21 pm](https://cloud.githubusercontent.com/assets/12230197/16218269/0eb73f6e-37ae-11e6-99e9-d2b90091cf4c.png)

Do I add a text-align: right here?

> textarea.materialize-textarea:focus.valid + label:after {
>   content: attr(data-success);
>   color: #4CAF50;
>   opacity: 1;
> }

fega:
solved, also the position is accordingly with the material design patterns https://material.io/guidelines/patterns/errors.html

PR #3287, opened 2016-06-21
Fix text label and error label overlap

rahulcomp24:
Fixed issue #3286 without shifting data-error to right.

Earlier when we have same row and two data fields are coming vertically as shown in issue #3286.

Before fixing 

image

After fixing it looks like this

image


rahulcomp24:
I modified the min file directly. Will make changes to CSS file in bin folder and request soon.

rahulcomp24:
I modified the min file directly. Will make changes to CSS file in bin folder and request soon.

PR #3288, opened 2016-06-21
Fix text label and error label overlap

rahulcomp24:
Fix text label and error label overlap when two textfields are in same row and vertically stacked.

Fixed issue #3286 without shifting data-error to right.

Earlier when we have same row and two data fields are coming vertically as shown in issue #3286.

Before fixing
image

After fixing
image


acburst:
Thanks!

PR #3318, opened 2016-06-26
add hover effect to button backgrounds

idontbyte:
added a 'color'-btn class to the colors file so that you can change the background of buttons whilst maintaining the hover effect of lightening the background 5%.

Example usage - http://jotmagic-pl.azurewebsites.net/Basics/Buttons
(Note I use 12% lighten here)


spaceneenja:
Good idea, but shouldn't this be a part of the button module?

tomscholz:
Your link isn't working :=)

PR #3326, opened 2016-06-27
Add functions to allow programmatically adding of tooltip and position…

Tempus35:
I took two PR's that have not been merged #3004 and #2510 yet and added them and added ability to also send in tooltip text programmatically.

### DIFF
--- a/js/tooltip.js~
+++ b/js/tooltip.js~
@@ -5,9 +5,7 @@
 
       // Defaults
       var defaults = {
-        delay: 350,
-        tooltip: '',
-        position: 'bottom'
+        delay: 350
       };
 
       // Remove tooltip from the activator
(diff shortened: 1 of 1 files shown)


acburst:
Thanks!

wzup:
And how does it work? How to use? Were docs adjusted accordingly?

Tempus35:
I do not have access to docs but it works as you had in you issues posts. You can now do:

$('#something').tooltip({delay:50,position:'top',tooltip:'Hello World'});

Instead of the data-tooltip etc on the element, also think there is no need for tooltipped class if you init yourself.

GorgonsMaze:
@Tempus35 Do you know if you can change the text and background color using this? I'm trying  to change multiple tooltips throughout to different colors?

Tempus35:
If you enable html via data-html="true" or html:true then you can do something like <p style="color:#000000">Something</p> for the text, at least that should work. As for background, I do not think that is possible at the moment. Might be good to make a new param that can take a custom class or classes.

js/tooltip.js~
```diff
@@ -0,0 +1,203 @@
+(function ($) {
```
annismckenzie: What's this file doing in this PR? Looks like a backup file of some sort.

Tempus35: Its likely something my IDE made, likely a backup that I forgot to exclude.

PR #3341, opened 2016-07-01
fix markup mistake for optgroups in forms.html

MuffN:
The Documentation for Forms shows a single select Optgroup Form (See Line 413 in forms.html) but the markup presented later uses `<select multiple>` thus not matching the example shown.


MuffN:
I should have modified the jade file... New pull request here: https://github.com/Dogfalo/materialize/pull/3342

Issue #210, opened 2014-12-09
Parallax not working, as directed


akihitoalextsuboi:
Other js components working.
When appended class='parallax', image will disappear.
My slim file is like this.

``` top.html.slim
.parallax-container
  .parallax
    = image_tag 'top_hawaii.jpg', width: '100%'
```

cchandurkar:
I am using 0.95.3 from cdnjs.cloudflare.com. It's still not working for me.

allthesignals:
I'm still seeing this issue. I have materialize-sass 0.96.1. It works on first load, but navigating to new pages and returning to the page with the parallax effect disappears.

iesen:
i am seeing this issue in 0.96.1. Image is hidden and not displayed..

nosajnosliw:
I'm also seeing this and have been I was using .95 and now am using .96.1. any clues why this is happening? I need to launch my site today.

allthesignals:
My issue was actually related to [turbolinks](http://stackoverflow.com/questions/18770517/rails-4-how-to-use-document-ready-with-turbo-links). If you're using Rails, double-check this.

andersdn11:
I also can't get the parallax to work. 

Using CDN 0.97.0 

Any solutions to this?

DmitrySharikov:
Same problem with 0.97

DmitrySharikov:
I think I found solution for this problem.
@andersdn11 just delete any size settings for image

andersdn11:
I dont have any size settings, still won't work. Could it be because i'm working localhost?

DmitrySharikov:
I don't think so. I'm also working in localhost






__Best regards



Dmitry SharikovRuby on Rails developer





http://dmitrysharikov.github.io

DavidCWebs:
I had this problem with v0.96.0 - fixed it by adding
~~~
(function($){
  $(function(){
    $('.parallax').parallax();
  });
})(jQuery);
~~~
...in the head rather than in the footer. Images are now visible, parallax working as expected.

nosajnosliw:
I swapped out out the .95 js code for dropdowns into my .96 js and all works fine now.  Hope the new build supports parallax and dropdowns.

andersdn11:
so i got it working. My problem seemed to be the z-index: -1 on parallax. class

Fix:
.parallax{ z-index: 0; }

Sajgoniarz:
[...]

PR #3344, opened 2016-07-02
Fix for issue #210

jtkDvlp:
Fix for issue #210 that works for me. 

`$(this).load` doesn't work without args!

It´s not necessary to call `load` because either it is already loaded or it will be and the pre given handler will work.

Hope it helps.

### DIFF
--- a/js/parallax.js
+++ b/js/parallax.js
@@ -39,7 +39,7 @@
         $this.children("img").one("load", function() {
           updateParallax(true);
         }).each(function() {
-          if(this.complete) updateParallax(true);
+          if(this.complete) $(this).trigger("load");
         });
 
         $(window).scroll(function() {
--- a/js/parallax.js
+++ b/js/parallax.js
@@ -39,7 +39,7 @@
         $this.children("img").one("load", function() {
           updateParallax(true);
         }).each(function() {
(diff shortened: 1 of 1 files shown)


jtkDvlp:
Help please, I dont know why it has failed.

js/parallax.js
```diff
@@ -39,7 +39,7 @@
         $this.children("img").one("load", function() {
           updateParallax(true);
         }).each(function() {
-          if(this.complete) $(this).load();
+          if(this.complete) updateParallax(true);
```
cerebroso: what if you change 
```
trigger('load');
```
to
```
trigger('load')
```
?
That's what the jquery 3 migration guide recommends

https://jquery.com/upgrade-guide/3.0/#breaking-change-load-unload-and-error-removed

NitroBAY:
any news ?

js/parallax.js
jtkDvlp: Sry, I missed the comment. I will test it myself and will give feedback and accordingly a commit.

jtkDvlp:
Do I have to reslove the merge conflict? oO

tomscholz:
This is fixed

Issue #3139, opened 2016-05-09
Cannot find module './picker.js'


denzBrade:
Discovered this issue then researched it. The OP can be found here by numerous other people:
https://github.com/Dogfalo/materialize/issues/1422

This is a **big issue** for those of us trying to either require or import the NPM version.
Please could this be addressed.

To clarify when running the project the picker.js file is missing from the '/materialize-css/bin' folder.

criticalmaas:
I was able to bypass this by simply copying picker.js into the bin directory using npm postinstall command. **_It's a temporary solution only._**

```"postinstall": "cp node_modules/materialize-css/js/date_picker/picker.js www/lib"```

**Comment by denz-brade-2015** [comment]
@criticalmaas nice one & thanks for the trick, could prove useful in the interim. 
Still I hope they can address this issue its an extremely popular issue which would deter people from using this nice framework.

**Comment by actuallymentor** [comment]
Still an issue. I applied a temporary fix that still allows for require('materialize-css') in a browserify/gulp setting.

```  "postinstall": "cp node_modules/materialize-css/js/date_picker/picker.js node_modules/materialize-css/bin" ```

**Comment by aeksco** [comment]
Also seeing this issue with Browserify.

**Comment by nmors** [comment]
i'm getting this issue with browserify

**Comment by fernandogmar** [comment]
I am still getting this issue, 2017!! :sweat_smile: 

Thanks!

**Comment by fernandogmar** [comment]
Ok, I am using jspm to install it so this is what I finally did:

```
jspm install materialize-css=npm:materialize-css -o '{format: "amd", "main": "bin/materialize.js", shim: { "bin/materialize": ["jquery"] }}'
```

just in case someone got the same problem ;)

Jujunol:
Being such a big issue I'm surprised that this isn't patched yet.

denzBrade:
Seeing all these recent messages :( 
Sorry guys the point of my post was for the issue to be rectified asap, 
hope you all have a temporary work around in the meantime.

NitroBAY:
@Dogfalo why don't you fix that ?

christopher-s:
This is a pretty severe little bug. Can we get it in the next release right away?

madsem:
ridiculous that since and related issue are open since 2015. Jeez

Flubbex:
[...]

PR #3355, opened 2016-07-07
copy tasks to put picker where its required

nicosommi:
A temporal fix for
https://github.com/Dogfalo/materialize/issues/3139
and
https://github.com/Dogfalo/materialize/issues/1422

The thing is that when you run for example your tests you get 'Cannot find the module ./picker.js'. That's because the concat task on the Gruntfile does not do anything with the paths within require calls.
It is just two lines to copy the two extra files on those paths so the require works on node.js, the other differences are just because consequence of running the release tasks.

This is a temporal fix because it may be better to remove or take a different solution for this problem and I don't have the proper context to clearly see that. But this is literally making users choose a different tool.


nicosommi:
Recents pull requests are failing all because of the same thing: a failing test regarding to a tooltip test

acburst:
Yep, that is my bad. I just fixed this, you can sync and try again

nicosommi:
all right now it's in sync thanks @acburst

june07:
When is this going to be merged?  It seems the issue remains.

june07:
Actually, not sure why all those other files were changed?!  Please see the PR I just submitted: https://github.com/Dogfalo/materialize/pull/4158.  Thanks.

nicosommi:
@june07 all right then I will close this one. Yes I forgot to stash some things I've done test the buils and the changes.

PR #3357, opened 2016-07-07
Adds support for external links in tab module

dmitov:
That should've been submitted already, but unfortunately it's not in current version of the lib.

Whenever a link among the tabs has either `target="_blank"` or `target="_self"` this should bypass the tabs behavior and take the user to the `href` location.

PS: Apologize for the strage `git diff` but **Atom** is doing things because of... well... reasons.


tomscholz:
See: http://materializecss.com/tabs.html#external

PR #3364, opened 2016-07-11
Update _input-fields.scss to fix problem with input labels when using sass version

omensinger:
"-webkit-transform" is included in distributed (minified) css for labels of input fields. But it gets lost when compiling from scss. The animation of the label will work no longer when using the sass version. Add "-webkit-transform: translateY(-140%);"to "label.active" in _input-field.scss to solve this problem!

Or you could think of something like http://bourbon.io/ for any transition/transform (I found some more, in _navbar.scsc for example).

### DIFF
--- a/sass/components/forms/_input-fields.scss
+++ b/sass/components/forms/_input-fields.scss
@@ -150,6 +150,7 @@ textarea.materialize-textarea {
   label.active {
     font-size: $label-font-size;
     transform: translateY(-140%);
+    -webkit-transform: translateY(-140%);
   }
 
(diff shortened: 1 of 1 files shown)


kmmbvnr:
To get browser dependent css from sass you need to add `autoprefixer` to your build script.

`Autoprefixer` is much better solution, b/c it works automatically

See

https://github.com/Dogfalo/materialize/blob/master/Gruntfile.js#L78

omensinger:
Thank you! Please forget about this pull request!

PR #3371, opened 2016-07-14
Fix mistake in Grid system $large-and-up value

ershov-ilya:


kmmbvnr:
This change makes `medium-and-up` definition  inconsistent with `large-and-up`

Actually you define not `large-and-up` but `upper-than-large`

PR #2124, opened 2015-10-05
Tabs: External Links (optional)

msolters:
Specifying the optional `target` property of a tab hyperlink will allow it to behave as a normal hyperlink.  Does not affect any other tabs, which continue to work normally.  External tab links do not have any internal tab content.  Resolves #2121.

I have included documentation w/ example in `tabs.html` as well as a test tab inside `test/html/tabs.html`.

PR #3379, opened 2016-07-17
Allow External Links From Tab

rickr:
This is a cherry-pick of bc3837f334b68a9bc914f0855f620717d367896b which seems to have been lost?


stap123:
bump, @acburst any movement on this being merged?

Dogfalo:
To me, this seems like an undesirable behavior. Tabs should switch panes within its associated view, not link you to some external place.

stap123:
Not sure I agree/understand. This was merged in previously by @acburst in #2124 but was lost. Why the change of heart? Seemed to get a lot of positive response before.

PR #2847, opened 2016-02-27
make sidenav usable for webapps - added destroy

makeros:
sidenav.js small refactor

PR #3419, opened 2016-07-27
Add 'destroy' method to sideNav API

benelliott:
This PR is inspired by #2847 but reimplemented as a simpler fix so that it can be merged more easily. It adds a 'destroy' method to the sideNav API so that any existing instances can be removed from the DOM.


acburst:
Thanks for this.

I fixed some issues with variables and scope that were causing issues with multiple side-navs being initialized

PR #3451, opened 2016-08-02
Correction autocomplete documentation

vpiassa29:
In autocomplete documentation we have an id attribute that actually should be class attribute. This confuses beginners.

### DIFF
--- a/forms.html
+++ b/forms.html
@@ -940,7 +940,7 @@
       &lt;div class="row">
         &lt;div class="input-field col s12">
           &lt;i class="material-icons prefix">textsms&lt;/i>
-          &lt;input type="text" id="autocomplete">
+          &lt;input type="text" class="autocomplete">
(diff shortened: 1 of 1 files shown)


acburst:
The Id's are actually intentional. It is to allow our labels to use the for attribute: http://www.w3schools.com/tags/att_label_for.asp

vpiassa29:
However if you follow the documentation you can not operate the autocomplete.

Thanood:
@vpiassa29 is right, the initialization example states:

```javascript
    data: {
      "Apple": null,
      "Microsoft": null,
      "Google": 'http://placehold.it/250x250'
    }
  });
```

This part: `$('input.autocomplete')` initializes an input with *class* `autocomplete` instead of an input with id `autocomplete`.

acburst:
Fixed in 086ba34. Thanks for pointing this out!

Issue #2452, opened 2015-12-10
Srcset support in lightbox ?


SheepDomination:
Does this Lightbox support Srcset ?

ITfee:
Will srcset be implemented into Materialize Material Box in the foreseeable future?

emxjay:
+1

lucasmoyano:
some news? :C

PR #3468, opened 2016-08-05
Support srcset and sizes attributes for materialbox.js

bcole808:
Responsive images with the `srcset` and `sizes` property set allow the browser to dynamically download a larger source file as needed when the browser or image size changes. 

This pull request adds support for images that use these properties:

``` html
<img 
    class="materialboxed" 
    width="650" 
    src="sample.jpg" 
    srcset="sample-thumbnail.jpg 200w, sample-large.jpg 800w" 
    sizes="(max-width: 200px) 100vw, 200px" />
```

When the image is expanded, materialbox.js will now update the `sizes` attribute (if present) to match the larger size that the image is about to be displayed at. 

It will also reset the `sizes` attribute to its original value when `returnToOriginal()` is called.  

For sites that use these properties along with materialboxed, this pull request enables the thumbnails to be displayed with small source files, and the browser will download the appropriate larger file when the image is expanded. 
#2452

### DIFF
--- a/js/materialbox.js
+++ b/js/materialbox.js
@@ -121,6 +121,12 @@
           newHeight = windowHeight * 0.9;
         }
 
+        // Update responsive image sizes if present
+        if(origin[0].hasAttribute('sizes')) {
+          origin.data('sizes', origin.attr('sizes'));
+          origin.attr('sizes', Math.round(newWidth) + 'px');
+        }
+
         // Animate image + set z-index
         if(origin.hasClass('responsive-img')) {
           origin.velocity({'max-width': newWidth, 'width': originalWidth}, {duration: 0, queue: false,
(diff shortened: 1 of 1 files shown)


ITfee:
@bcole808 Tested this in three browsers:

Firefox = OK
Chrome = for the thumbnail image the large version is loaded instead of thumbnail version
Huawei Browser = when touching thumbnail, the thumbnail image is maximized instead of loading the large version of it

tomscholz:
Can somebody also test this in ie and edge?

bugsoftware:
When will this merge happen?

Dogfalo:
materialbox was rewritten so this can't be merged

ITfee:
@Dogfalo So srcset support in Materialize will never happen?

Artem-Schander:
Hi everyone, i thought i share my workaround:
```javascript
M.Materialbox.init(document.querySelectorAll('.materialboxed'), {
    onOpenStart: el => {
        if (el.hasAttribute('data-large-src')) {
            const largeSrc = el.getAttribute('data-large-src');
            el.setAttribute('src', largeSrc);
        }
    },
    onCloseEnd: el => {
        if (el.hasAttribute('data-small-src')) {
            const smallSrc = el.getAttribute('data-small-src');
            el.setAttribute('src', smallSrc);
        }
    }
});
```

Issue #3492, opened 2016-08-10
Accessibility of select lists with images


btopro:
Images in the select lists complain about not having alt info because none is set. This is an easy one, it the alt info is set to null (as they are purely decorative -- http://webaim.org/techniques/alttext/#decorative ) then it'll be skipped by screen readers / be happy.
`alt=""` will fix this issue

btopro:
Associated pull request #3493

btopro:
PR was accepted so I'll close out :)

PR #3493, opened 2016-08-10
Accessibility of select lists with images #3492

btopro:
This PR adds null alt="" attributes to icons that appear within the fancy select lists that have images in them. This helps improve accessibility of these items because these images are purely for decoration and setting a null alt attribute tells screen readers to ignore them.


acburst:
Thanks! We definitely can approve our accessibility all around. Let me know if you find any other areas of improvement

PR #3525, opened 2016-08-16
Input field's label made not interactive.

jesion:
This PR contains a fix for input field's label. A label could be improved to be not interactive. Having label responsive to selection and mouse events introduces UX issues in Angular 2 apps using materializecss project for styling.

### DIFF
--- a/sass/components/_mixins.scss
+++ b/sass/components/_mixins.scss
@@ -3,3 +3,16 @@
 //     -moz-box-shadow: $args1, $args2;
 //     box-shadow: $args1, $args2;
 // }
+
+@mixin not-selectable {
+	-webkit-touch-callout: none;
+    -webkit-user-select: none;
+    -khtml-user-select: none;
+    -moz-user-select: none;
+    -ms-user-select: none;
+    user-select: none;
+}
+
+@mixin not-clickable {
+	pointer-events: none;
+}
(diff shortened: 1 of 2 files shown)


kmmbvnr:
Yeah, that also sometimes annoys me in a usual project w/o angular js. On some assident click, you can get a label selected instead of entering into a field.

sass/components/_mixins.scss
```diff
@@ -2,4 +2,17 @@
 //     -webkit-box-shadow: $args1, $args2;
 //     -moz-box-shadow: $args1, $args2;
 //     box-shadow: $args1, $args2;
-// }
\ No newline at end of file
+// }
+
+@mixin not-selectable {
+	-webkit-touch-callout: none;
```
kmmbvnr: Are browser specific prefixes required here? MateralizeCSS uses autoprefixer to add browser-dependent CSS https://github.com/Dogfalo/materialize/blob/master/Gruntfile.js#L81

You can try to remove it, build and see what prefixes would be added automatically.

jesion: Those browser specific prefixes were removed. Also I have renamed mixins to be more inline with your naming conventions.

tomscholz:
https://github.com/Dogfalo/materialize/pull/4242#issuecomment-305600520

Issue #3490, opened 2016-08-10
Accessibility of select lists


btopro:
When applying the select style to select lists that have labels it breaks the convention of having the label right next to the item its tied to. It goes from
```
<label for="stuff"></label>
<select id="staff">....</select>
```
to this
```
<label for="stuff"></label>
<input .... />
<select id="staff">....</select>
```
The input field should have an ID so that there's a label for it. Possible way to do this:
- generating a hashed id for the input field
- changing the for attribute on the label to the hash

The select is hidden so I think this would resolve the accessibility flag seen below.

![2016-08-10_08-32-05](https://cloud.githubusercontent.com/assets/329735/17553649/f700cb30-5ed4-11e6-9703-4b07b7e0d052.png)

btopro:
in testing, the label swap fixes the issue but generates a new one as there is no label associated with the original select. This most likely doesn't matter because the original select field is hidden and hopefully would be skipped by screen readers.

btopro:
https://github.com/Dogfalo/materialize/pull/3526 pull request to match this

PR #3526, opened 2016-08-16
improved accessibility of select forms #3490

btopro:
This improves the accessibility of select lists both on the demonstration page itself as well as in the output after the js forms function applies the select style with dropdown list. The demo page is an easy fix and should be accepted regardless since it's a basic accessibility improvement.

The js change looks to see if the select it is working on has an `id` attribute, then it checks for a matching label via `for` attribute to match. If it finds these two things (meaning it is accessible) then it generates a clone of the `<label>` with a `for` attribute to match a new guid generated for the `input` field that was appended. It then hides the original label without removing it (which will squash accessibility strict checkers like webaim Wave plugin).

This should satisfy the accessibility of the select js plugin and also has the added benefit of now allowing users to click on the label and have it activate the dropdown so it improves usability as well.

### DIFF
--- a/forms.html
+++ b/forms.html
@@ -384,33 +384,33 @@
           <form class="col s12">
             <div class="row">
               <div class="input-field col s12 m6">
-                <select>
+                <select id="choose-options-1">
                   <option value="" disabled selected>Choose your option</option>
                   <option value="1">Option 1</option>
                   <option value="2">Option 2</option>
                   <option value="3">Option 3</option>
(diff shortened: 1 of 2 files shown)


tomscholz:
Hey there 👋 
```
/materialize/sass/
/materialize/js/ 
```

Dogfalo:
Select is now keyboard acessible in 1.0 alpha

Issue #3434, opened 2016-07-29
Turn off dragging side nav from the edge.


kuxa:
Hi,

I use two side navs, both from left side. The thing is I want the second menu to be opened **only** through the link in the first menu, so it highly bothers me that it is possible to open it also by dragging from the left edge. It would be nice if it would be possible to turn off dragging when initializing side nav, for example
`$('.side-nav-2').sideNav({
      closeOnClick: true,
      openOnDrag: false
    }
  );`

This property would of course be true by default. I believe it would not be a big problem for you to implement this and it would make a lot of lives easier.

Thanks a lot!

acburst:
Added draggable as a side-nav option in 8301f86

PR #3529, opened 2016-08-16
DragTarget on SideNav is optional

kuxa:
You can now disable DragTarget on Sidenav by including _dragTarget: false_ into options while initializing SideNav. (solves #3434)


acburst:
There were some merge errors and I tweaked the implementation.

Added draggable in 8301f86.

Thanks

Issue #3514, opened 2016-08-14
Search by typing for select doesn't work with Ukrainian and Russian


VladTishchenko:
Hi, how can I use select search by typing with Ukrainian or Russian?

Issue #3392, opened 2016-07-21
"Search by typing" in select


K1llMan:
Hi!
Wrong work "search by typing" option with different locales. "keypress" event returns correct localized keycode but "keydown" not.

PR #3530, opened 2016-08-17
Fix for selects when searching with typing in characters - Update forms.js

brgrz:
Switched to use onKeyPress instead of onKeyDown because onKeyPress correctly handles international key char codes


brgrz:
fixes issues #3514 and #3392

PR #3548, opened 2016-08-19
Use already defined `view` var instead of `$(this)`

snicky:


tomscholz:
cc @DanielRuf @fega

DanielRuf:
@snicky can you rebase it against the current master?

@tomscholz should be ok to merge after resolving the conflict. Or we create a new PR.
This PR saves an unnecessary DOM lookup, uses the cached selector and improves the performance a bit.

PR #3552, opened 2016-08-21
Update initial.js

hammadfauz:
Fix for the error `TypeError: $*.velocity is not a function` that occurs when using webpack to require('materialize-css')


echoAlexey:
Can we have this <3

tomscholz:
Related? https://github.com/Dogfalo/materialize/pull/3782

tomscholz:
Can you please test if https://github.com/Dogfalo/materialize/pull/3782 solves your problem too? If so I would close this PR

PR #3576, opened 2016-08-26
Autocomplete support for angularjs ng-model

vsodre:
When using ng-model on autocomplete fields the value captured is the typed one, not the selected on the list. This change tells angular that the value has changed.

Furthermore, this change improves autocomplete since it mocks browser's behavior of broadcast any change on input fields.


acburst:
Thanks!

Issue #3583, opened 2016-08-29
Add support to disable appending images with auto complete


mushtaqak:
I would like a default option that would let us disable appending images to autocomplete li.

P.S: I am working on PR that would allow us this option.

yanickrochon:
Currently, if the associated data value is a string, it is displayed as an image URL. Just add the feature that, if it is an object, then it is some data which is *not* an image. This way, no need for extra options, and it should be 99% BC.

For example : 

```
// standard implementation
$('input.autocomplete').autocomplete({
  data: {
    "Apple": null,
    "Microsoft": null,
    "Google": 'http://placehold.it/250x250'
  }
});

// data has data... (i.e. no images)
$('input.autocomplete').autocomplete({
  data: {
    "Apple": null,
    "Microsoft": null,
    "Google": { id: 1, url: 'http://placehold.it/250x250' }
  }
});
```

mushtaqak:
@yanickrochon Looks a good idea, let me find some time and come up with a better solution. Thanks

PR #3584, opened 2016-08-29
Add support to disable appending images with auto complete

mushtaqak:
Adds  default option that would let us disable appending images to autocomplete li. By default, it would work as it working previously.

If you want to disable adding images to autocomplete, just do 

```
    data: {
      "Apple": null,
      "Microsoft": null,
      "Google": 'http://placehold.it/250x250'
    },
   append_image: false
  });
```

**Before**
screen shot 2016-08-29 at 3 48 23 pm

**After**
screen shot 2016-08-29 at 3 48 59 pm

Described in [issue#3583](https://github.com/Dogfalo/materialize/issues/3583)

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -281,7 +281,8 @@
     $.fn.autocomplete = function (options) {
       // Defaults
       var defaults = {
-        data: {}
+        data: {},
+        append_image: true
       };
 
       options = $.extend(defaults, options);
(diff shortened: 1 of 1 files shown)


mushtaqak:
@acburst Please review. Thanks

yanickrochon:
Personally, I don't like underscores in attributes. See the relative issue for my comment on this.

mushtaqak:
Not going to work on this. SO closing this. Feel free to re-open and work if you need.

LennartOlsen:
@yanickrochon to be clear, 

The desired solution :
"if data is an object and not a string we wont append images?"

Issue #3586, opened 2016-08-29
Add support to add data attribute to autocomplete


mushtaqak:
Add a data attribute to the autocomplete input field so that the value can be used when we pass data list to autocomplete plugin.

E.g.

```
$('input.autocomplete').autocomplete({
    data: {
      "Apple": 1,
      "Microsoft": 2,
      "Google": 3
    },
    data_attr : 'data-id'
  });
```

yanickrochon:
I'd rather have a callback when a value is selected.

```
const data = {
  "Apple": 1,
  "Microsoft": 2,
  "Google": 3
}; 
$('input.autocomplete').autocomplete({
  data: data,
  select: function (target, selection) {
    var val = data[selection];
    $(input).data('id', val);
  }
});
```

mushtaqak:
@yanickrochon Looks a good idea, let me find some time and come up with a better solution. Thanks

fega:
Closed due inactivity, feel free to reopen it if it is still necessary.

PR #3589, opened 2016-08-29
Add a data attribute to the autocomplete input field so that the valu…

mushtaqak:
Add a data attribute to the autocomplete input field so that the value so that the value can be used when we pass data list to autocomplete plugin.

E.g :

```
    data: {
      "Apple": 1,
      "Microsoft": 2,
      "Google": 3
    },
    data_attr : 'data-id'
  });
```

**Before**

```
<input type="text" id="autocomplete-input" class="autocomplete">
```

**After**

```
<input type="text" id="autocomplete-input" class="autocomplete" data-id="1">
```

Addresses [Issue#3586](https://github.com/Dogfalo/materialize/issues/3586)

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -281,7 +281,8 @@
     $.fn.autocomplete = function (options) {
       // Defaults
       var defaults = {
-        data: {}
+        data: {},
+        data_attr : ''
       };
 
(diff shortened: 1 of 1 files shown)


mushtaqak:
Not working on it

PR #3605, opened 2016-09-04
add full width button

bardiarastin:
I guess it's very useful .


tomscholz:
I think this we are not looking to add this into our core framework. It is easy enough to customize yourself.

PR #3615, opened 2016-09-06
Remove hard-coded vertical align.

Munawwar:
Remove hard-coded 'top' value for vertical aligning Select's dropdown arrow. This adapts to any height of parent element.


acburst:
Thanks for this!

Munawwar:
I remembered I had to add `line-height: 1` as well, but you spotted it :) Thanks.

PR #3629, opened 2016-09-08
Fix lorempixel image not showing

GGG-KILLER:
Currently on Chrome Version 52.0.2743.116 m (64-bit), the URL `http://lorempixel.com/100/190/nature/10` returns no content and the `Content-Length` header is `0`.

I am not sure if the image on the new URL provides the desired result but it works.


acburst:
Fixed in 47b2f3f

PR #3643, opened 2016-09-12
Bump scss verion for meteor

ForsakenHarmony:
Fix for crash under windows (at least for me)


StorytellerCZ:
This can be closed as 0.97.8 uses `fourseven:scss@3.10.0`.

PR #3648, opened 2016-09-13
Variable grid number

chriscauley:
I prefer to have 4 media queries for grid cutoff points instead of 3, so I made abstracted out the media queries and put them into an `@each` loop. Then there only needs to be a $grid-sizes variable which is a list of lists containing the cutoff and the media query for that grid size. I couldn't figure out how to get it to put the smallest grid not in a media query, so I put it as `only screen and (min-width: 0)`. I couldn't see anything about testing css, but the "More Responsive Grid Examples" on grid.html works as expected.


tomscholz:
Hey there 👋 
```
/materialize/sass/
/materialize/js/ 
```

PR #3649, opened 2016-09-13
Create buttom navigation

pjdepractica:
It would be great to have a lower navigation. I'm something new with it. Thank you.


acburst:
I'm sorry but this is not how our components are structured. If you want to make a demo of what a component could look like, feel free to submit this as a codepen in an issue.

But this is not something we can merge in this state.

PR #3666, opened 2016-09-16
homogenize backdrop transition style

paulpflug:
height of 14 for left and right positioning gives no benefit, so I used height of 7 (current default for top and bottom) for all positions.

width of 14 and height of 7 are hard-coded now. before it was half hard-coded and half dynamic.

marginTop / marginLeft removed - consistently using top / bottom / left / right instead.

missing `top: 0` on bottom position can produce wrong positioning of the backdrop


Dogfalo:
Tooltip animation has been reworked

PR #3672, opened 2016-09-19
material_select fix scrolling with key navigation with multiple select boxes

guzmanfg:
When material_select widget is applied to a multiple selector (e.g. $('.select').material_select()), only latest select scrolls properly using keyboard.

This commit fixes this misbehavior.


guzmanfg:
This is a minimal change. We are highly using it on a beta product and seems it actually fixes multiple select boxes problem. Did I miss any step to get this pull request accepted?

Dogfalo:
Usually it helps a lot to provide a codepen (use the link in CONTRIBUTING.md) to show us an example of the problem.

I tried recreating the problem myself, but it seems to be working for me. http://codepen.io/anon/pen/BpBEvz.

guzmanfg:
Sorry about that. 

The problem was solved by [e3668e36ad2c82f6520406cdb1a971c1a69d4f79](https://github.com/Dogfalo/materialize/commit/e3668e36ad2c82f6520406cdb1a971c1a69d4f79). 

Thank you very much!

PR #3694, opened 2016-09-27
Autocomplete improvement - replacing data, arrow keys usage, etc

nuplo:
Hi,
aoutocomplete in official version doesn't have a lot of crucial features so I decided to add:
-up and down arrow usage
-updating data for autocomplete - for instance for AJAX request
-if see auto complete list and you will make focusout action, whole list disapper. I think in most cases we don't want to display this list if we do something else. It's rather bug than feature.
-when you write something in auto complete field and press "Esc" button whole list will disappear
-whole content after autocomplete list go down when this list appear. I changed that to absolute position and now it doesn't affect other elements
-in my opinion this list is sometimes too long - for example if we have 100 elements. Because of that, I decided to extend default object with "maxElementsAmount". If you set that for example on 5 during initializtion, you will see only 5 elements from the begining of this list. If you skip that, you will see all matching elements.

Please check that, if you have any questions or comments just write. I'm hope so we can marge that soon because I use that in my project ;)

Cheers,
Adam

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -279,85 +279,161 @@
      * Auto complete plugin  *
      *************************/
     $.fn.autocomplete = function (options) {
-      // Defaults
       var defaults = {
-        data: {}
-      };
+        data: {},
+        maxElementsAmount: null
+      },
+      currentLi = 0,
+      autocompleteOption = null,
+      currentAutocompleteVal = '';
 
       options = $.extend(defaults, options);
 
-      return this.each(function() {
-        var $input = $(this);
-        var data = options.data,
-            $inputDiv = $input.closest('.input-field'); // Div to append on
+      // Set input value
+      function setIputValue(ulElement, inputElement) {
+        ulElement.on('click', 'li', function () {
+          inputElement.val($(this).text().trim());
+          ulElement.empty();
+        });
+      }
 
-        // Check if data isn't empty
-        if (!$.isEmptyObject(data)) {
-          // Create autocomplete element
+      // highlight elements
+      function highlight(string, $el) {
+        var img = $el.find('img');
+        var matchStart = $el.text().toLowerCase().indexOf("" + string.toLowerCase() + ""),
+        matchEnd = matchStart + string.length - 1,
+        beforeMatch = $el.text().slice(0, matchStart),
+        matchText = $el.text().slice(matchStart, matchEnd + 1),
+        afterMatch = $el.text().slice(matchEnd + 1);
+        $el.html("<span>" + beforeMatch + "<span class='highlight'>" + matchText + "</span>" + afterMatch + "</span>");
+        if (img.length) {
+          $el.prepend(img);
+        }
+      }
+
+      // function on keyUp
+      function keyUp(e, data, $input, $autocomplete) {
+        var val = $input.val().toLowerCase();
+
+        // Capture Enter
+        if (e.which === 13) {
+          $autocomplete.find('li')[currentLi].click();
+          return;
+        }
+
+        // hide list on ESC button
+        if (e.which === 27) {
+          $input.blur();
+        }
+
+        // Check if the input isn't empty and it's not equal to old option
+        if (val !== '' && currentAutocompleteVal !== val) {
+          var maxElementsAmount = options.maxElementsAmount;
+          $autocomplete.empty();
+
+          for (var key in data) {
+            if (maxElementsAmount === 0) {
+              break;
+            }
+
+            if (data.hasOwnProperty(key) &&
+            key.toLowerCase().indexOf(val) !== -1 &&
+            key.toLowerCase() !== val) {
+              if (maxElementsAmount !== null) {
+                maxElementsAmount--;
+              }
+
+              autocompleteOption = $('<li></li>');
+              if (!!data[key]) {
+                autocompleteOption.append('<img src="' + data[key] + '" class="right circle"><span>' + key + '</span>');
+              } else {
+                autocompleteOption.append('<span>' + key + '</span>');
+              }
+              $autocomplete.append(autocompleteOption);
+
+              highlight(val, autocompleteOption);
+
+            }
+          }
+
+          currentAutocompleteVal = val;
+          autocompleteOption = $autocomplete.find('li');
+          currentLi = 0;
+          $(autocompleteOption[currentLi]).addClass('active');
+        }
+      }
+
+      // arrow usage - on key down
+      function arrowUsage(e, $autocomplete) {
+        // Capture up and down key
+        if (
+        e.which === 38
+        || e.which === 40
+        ) {
+          e.preventDefault();
+
+          if (currentLi > 0 && e.which === 38) {
+            currentLi--;
+          }
+
+          if (currentLi < (autocompleteOption.length - 1) && e.which === 40) {
+            currentLi++;
+          }
+
+          $autocomplete.find('.active').removeClass('active');
+          $(autocompleteOption[currentLi]).addClass('active');
+        }
+      }
+
+      // reset current element position
+      function resetCurrentElement($autocomplete) {
+        currentLi = 0;
+        $autocomplete.find('.active').removeClass('active');
+        $(autocompleteOption[currentLi]).addClass('active');
+      }
+
+
+      // initialization for each element
+      return this.each(function () {
+        var $input = $(this),
+        data = options.data,
+        $inputDiv = $input.closest('.input-field');
+
+        // Check if data isn't empty and append autocomplete element if doesn't exist yet - run only once
+        if
+        (!$.isEmptyObject(data)
+        && $inputDiv.find('.autocomplete-content').length === 0) {
           var $autocomplete = $('<ul class="autocomplete-content dropdown-content"></ul>');
 
-          // Append autocomplete element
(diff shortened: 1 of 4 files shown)


totient:
Works well; adapted this in a commercial project.

arlowhite:
Also, I wanted a 'select' event for when the user clicks an entry in the autocomplete or presses enter. ('change' fires for every key)

What are your thoughts on a 'select' event?

See #3729

js/forms.js
```diff
@@ -279,85 +279,181 @@
...
-            $input.val($(this).text().trim());
-            $input.trigger('change');
-            $autocomplete.empty();
+          // reset selected element on the list
+          $input.blur(function () {
+            resetCurrentElement($autocomplete);
           });
+
+          // updating input value
+          setIputValue($autocomplete, $input);
```
arlowhite: Typo. should be **setInputValue**

js/forms.js
```diff
@@ -279,85 +279,181 @@
...
+      currentAutocompleteVal = '';
 
       options = $.extend(defaults, options);
 
-      return this.each(function() {
-        var $input = $(this);
-        var data = options.data,
-            $inputDiv = $input.closest('.input-field'); // Div to append on
+      // Set input value
+      function setIputValue(ulElement, inputElement) {
```
arlowhite: Why **setIputValue** instead of **setInputValue**?

js/forms.js
```diff
@@ -279,85 +279,181 @@
...
+      autocompleteOption = null,
+      currentAutocompleteVal = '';
 
       options = $.extend(defaults, options);
 
-      return this.each(function() {
-        var $input = $(this);
-        var data = options.data,
-            $inputDiv = $input.closest('.input-field'); // Div to append on
+      // Set input value
```
arlowhite: The JavaDoc doesn't give more information than already hinted by the name of the function.

Is inputElement the main $input? Why is the parameter needed? Can't you just use $input captured in the closure?

Why is this function needed? It's only used once in the code. I think just remove the function and put its body where the call on line 450 is.

js/forms.js
```diff
@@ -279,85 +279,181 @@
...
+        ulElement.on('click', 'li', function () {
+          inputElement.val($(this).text().trim());
+          ulElement.empty();
+        });
+      }
+
+      // highlight elements
+      function highlight(string, $el) {
+        var img = $el.find('img');
+        var matchStart = $el.text().toLowerCase().indexOf("" + string.toLowerCase() + ""),
```
arlowhite: No reason to add empty strings, it doesn't do anything.

js/forms.js
```diff
@@ -279,85 +279,181 @@
...
+
+        // hide list on ESC button
+        if (e.which === 27) {
+          $input.blur();
+        }
+
+        // prevent refreshing list on pressing ctrl, shif, alt, arrow, f1-f12, etc
+        // helpful if we selected something using arrow and we press ctrl for example
+        if(
+        [9, 16, 17, 18, 19, 20, 33, 34, 35, 36, 37, 38, 39, 40, 45, 112, 113, 114, 115, 116, 117, 118, 119, 120, 121, 122, 123].indexOf(e.which) < 0
```
arlowhite: Alphanumeric codes are probably in the same range. So probably better to do  `if(e.which > X && e.which < Y)`

But I have no idea what X and Y should be. I started looking into the docs, but they don't mention numerical codes.
https://api.jquery.com/event.which/
https://www.w3.org/TR/uievents-code/

js/forms.js
```diff
@@ -279,85 +279,181 @@
...
+        ){
+          highlightElements(data, $input, $autocomplete);
+        }
+      }
+
+      // function to displaying elements
+      function highlightElements(data, $input, $autocomplete){
+        // Check if the input isn't empty and isn't equal to the old option
+        if (val !== '' && currentAutocompleteVal !== val) {
```
arlowhite: Use of val is confusing. Looks like it's used in the if statement before being defined on the next line.

js/forms.js
```diff
@@ -279,85 +279,181 @@
...
+          var val = $input.val().toLowerCase(),
+          maxElementsAmount = options.maxElementsAmount;
```
arlowhite: Missing **var**?

js/forms.js
```diff
@@ -279,85 +279,181 @@
...
+          for (var key in data) {
+
+            if (maxElementsAmount === 0) {
+              break;
+            }
+
+            if (data.hasOwnProperty(key) &&
+            key.toLowerCase().indexOf(val) !== -1 &&
+            key.toLowerCase() !== val) {
+              if (maxElementsAmount !== null) {
```
arlowhite: I think the code should establish that maxElementsAmount is a number before starting the for loop rather than checking if it is null here.

js/forms.js
```diff
@@ -279,85 +279,181 @@
...
+          currentLi = 0;
+          $(autocompleteOption[currentLi]).addClass('active');
+        }
+      }
+
+      // arrow usage - on key down
+      function arrowUsage(e, $autocomplete) {
+        // Capture up and down key
+        if (
+        e.which === 38
```
arlowhite: Maybe save e.which in var since it's used a few times.
`var keyCode = e.which;`

js/forms.js
```diff
@@ -279,85 +279,181 @@
...
+          autocompleteOption = $autocomplete.find('li');
+          currentLi = 0;
+          $(autocompleteOption[currentLi]).addClass('active');
+        }
+      }
+
+      // arrow usage - on key down
+      function arrowUsage(e, $autocomplete) {
+        // Capture up and down key
+        if (
```
arlowhite: if statement line breaks and indenting look weird. (Not normal JavaScript style)

js/forms.js
```diff
@@ -279,85 +279,181 @@
...
+        data = options.data,
+        $inputDiv = $input.closest('.input-field');
+
+        // remove list if it exist
+        if($inputDiv.find('.autocomplete-content').length !== 0){
+          $inputDiv.find('.autocomplete-content').remove();
+        }
+
+        // Check if data isn't empty and append autocomplete element if doesn't exist yet - run only once
+        if(
```
arlowhite: if statement line breaks and indenting look weird. (Not normal JavaScript style)

sass/components/forms/_input-fields.scss
```diff
@@ -270,9 +270,14 @@ textarea {
 /* Autocomplete */
 .autocomplete-content {
   margin-top: -15px;
-  display: block;
+  display: none;
   opacity: 1;
-  position: static;
+  position: absolute;
+
+  & ~ .autocomplete-content,
```
arlowhite: CSS changes should be reviewed, but I don't have time to look at them now.

js/forms.js
nuplo: Comment maybe doesn't bring so much, but I do that because it's easier for me to read a code. Comments have different color and for me when I looking through the code it's like headers.

I separated that intentionally. First and less important I think right now it's just more readable. In one section you define functions - in different section you fired that. 
But more important thing: if you move everything to one loop and have for example 100 autocomplete inputs, when you run this loop you will define all this function 100 times. Right now you will define functions only once and you can fire that 100 times. It's just about performance.

And you cannot move this code to line 450 because this function just will not work.

js/forms.js
nuplo: it's a legacy after old code. I changed that and I'm hope so it's not a some kind of hack for old IE or something.

js/forms.js
nuplo: I'm not so sure about that, but if you could provide solution like that and it will work we can use that.

js/forms.js
nuplo: Yup, you're right. It should be before if. Changed.

js/forms.js
nuplo: Duplicated - it's the same what you wrote before. Changed.

js/forms.js
nuplo: Hmmm... If you do that you can change:
	if (maxElementsAmount !== null)
On something like this:
	if (maxElementsAmountIsNumber === true)

I don't think it's a big change

js/forms.js
nuplo: I changed that, but all programmers has own style and there isn't "the style" which you can use in each situation. It's about preferences. If we want to have the same style in each file we should configure task in "gulp watch" which will be reformating code after save. That's the only way to have the same style in the whole project if you have more than 2 programmers ;)

js/forms.js
nuplo: look above

js/forms.js
nuplo: changed

sass/components/forms/_input-fields.scss
nuplo: I've just found one surplus line and deleted it but it doesn't change so much.

nuplo:
@arlowhite  I added callback onSelect which you described in #3729

gdoron:
📢 📢 📢  Why things in this project never get merged? 
What's the purpose of having an open source project if people's hard work is flush into the toilet every time?
@Dogfalo don't get me wrong, I really appreciate your hard work, but please let the community help you maintain this **amazing** project!

PR #3696, opened 2016-09-27
Avoid conflict between fixed and closeOnClick

FrancoisH:
Avoid the side-nav closing when the screen size is at or below 992 (when the side-nav gets collapsible)


Dogfalo:
Fixed via #4299

PR #3700, opened 2016-09-28
Add style field to package.json

RyanZim:
Users of https://github.com/postcss/postcss-import and similar tools need this. This allows us to do `@import "materialize-css";` in our css, and materialize's css will be inlined during the build step.

Bootstrap has this: https://github.com/twbs/bootstrap/blob/master/package.json#L21, materialize should too!

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "url": "http://materializecss.com/",
   "version": "0.97.7",
   "main": "bin/materialize.js",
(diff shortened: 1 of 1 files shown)


tomscholz:
Added in https://github.com/Dogfalo/materialize/commit/55ee054c54b10c2ecd8fad0ff970824d753cb70c

PR #3701, opened 2016-09-28
Added activeClass and getActiveElement options to scrollspy

TMaYaD:
This is a rewrite of #2300
Both the options have been split into separate commits and the getActiveElement is updated to take section id instead of base which makes it more flexible.

### DIFF
--- a/jade/page-contents/scrollspy_content.html
+++ b/jade/page-contents/scrollspy_content.html
@@ -67,6 +67,19 @@
               <td>activeClass</td>
               <td>Class name to be added to the active link. Default: active</td>
             </tr>
+            <tr>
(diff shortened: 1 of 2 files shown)


TMaYaD:
This PR makes it possible to use scrollspy with navbar:
```JS
    $('.scrollspy').scrollSpy(
        getActiveElement: function(id) {
            'a[href=#' + visible[0].attr('id') + ':parent]'
        }
    )
```

tomscholz:
@acburst @Dogfalo @DanielRuf @fega Looks and Sounds good to me.

DanielRuf:
@tomscholz I agree. Makes totally sense and it's well implemented.

PR #3706, opened 2016-09-30
long waves indicator

complynx:
If one has to indicate the element to handle long clicks in a different way than average ones, one might use this solution. Very handy, indeed.


Dogfalo:
Doesn't seem to be a commonly asked for feature, will consider reopening if more people request this.

PR #3726, opened 2016-10-04
Added .col .s12 div to navbar examples

ruslandzhumaev:
To fix this padding issues:


tomscholz:
Hey there 👋 
Unfortunately does your PR include compiled files. Please do not edit the .html in the root! You should edit the source files in `/materialize/jade/page-contents/` instead. Also make sure to check out our [CONTRIBUTING GUIDE](https://github.com/Dogfalo/materialize/blob/master/CONTRIBUTING.md)!

Dogfalo:
This is incorrect as it doesn't nest `col s12` inside a `.row`

PR #3729, opened 2016-10-04
Add 'select' event to autocomplete

arlowhite:
The 'change' event fires on every keystroke.

It would be nice to have another event for when the user clicks an item in in the autocomplete dropdown or hits enter on the keyboard.

What do you think?
Is "select" a good event name?

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -349,8 +349,10 @@
 
           // Set input value
           $autocomplete.on('click', 'li', function () {
-            $input.val($(this).text().trim());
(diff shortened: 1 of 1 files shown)


arlowhite:
I'm not sure why, but when you listen to _select_ you received **undefined** a few times.
```javascript
input.on("select", function(event, value) {
  // value undefined sometimes
});
```

nuplo:
added in #3694

arlowhite:
not working on this anymore

PR #3731, opened 2016-10-05
Fix/Global $ conflicts for Velocity

### LINKED ISSUE #3623: jQuery noConflict issues with velocity
If another library is loaded, like `prototype.js`, that makes use of a global `$`, [this line](https://github.com/Dogfalo/materialize/blob/4e5804ea59da9eef074f195bb9281f9d6e77f6df/js/global.js#L39) in `global.js` resloves as true and carries on using `$.velocity` even though jQuery might be in noConflict.

Bouncey:
Refactored to check for `noConflict` jQuery first.

This is to counter an issue where if a global `$` is being used by another library, Velocity breaks.

closes #3623

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -78,11 +78,12 @@ Materialize.elementOrParentIsFixed = function(element) {
 };
 
 // Velocity has conflicts when loaded with jQuery, this will check for it
+// First, check if in noConflict mode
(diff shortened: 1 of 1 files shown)


acburst:
Thanks!

PR #3734, opened 2016-10-05
Add showSingleMatch option to autocomplete.

arlowhite:
Currently, if the autocomplete text exactly matches a data key, the autocomplete is hidden.

With this option, the single match is displayed and the user will now be able to hit enter on the keyboard to fire the 'change' event for a single matching data entry.

Default is **showSingleMatch: false** so behavior will not change unless developers opt-in by adding **showSingleMatch: true**

One issue is that a single match is not highlighted, but I don't think it's worth modifying the highlighting code just to highlight the whole thing in blue.

Let me know if you like this feature and want me to update the Materialize documentation too.

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -281,7 +281,8 @@
     $.fn.autocomplete = function (options) {
       // Defaults
       var defaults = {
-        data: {}
+        data: {},
+        showSingleMatch: false
       };
 
       options = $.extend(defaults, options);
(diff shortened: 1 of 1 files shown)


arlowhite:
This may conflict or be redundant with https://github.com/Dogfalo/materialize/pull/3694

arlowhite:
not working on this anymore

PR #3735, opened 2016-10-05
Flow-text support for headers

ScienceDuck:
The changes here adapt the flow-text code to include support for header text. This enables headers to scale with screen size, and prevent large headers from looking oversized in proportion to the related page text. 

The changes proposed will adjust the headers whether the .flow-text class is added to the header or the header is nested within a .flow-text div.


ScienceDuck:
I copied the proposed changes into a Code Pen to demonstrate the effects.

[http://codepen.io/ScienceDuck/pen/XjVZPg](url)

PR #3747, opened 2016-10-07
Separate color variables and classes

### LINKED ISSUE #3743: Decouple color variables and classes
Color variables and classes are packed into single file. This creates some design issues.

One issue is palette extension. Since the classes are generated from the palette in-place and they are generated with `@each` and not mixins, the only way to add additional colors or change existing ones is to paste `@each` blocks and run them again. In the case of changing existing colors this will result in dupe classes that should be treated with CSS cleaner.

Another one is using color variables outside of the framework (Web Components, Angular components). Importing `_color.scss` alone into component SCSS file to access colors with `color('red', 'base')` will result in ~30 Kb CSS overhead (unminified).

The suggestion is to divide `_color.scss` into `color/_palette.scss` (variables, functions, mixins that produce no output on import) and `color/_classes.scss`.

For compatibility reasons `_color.scss` may be kept in place:

```
@import "color/_palette.scss";
@import "color/_classes.scss";
```

bisubus:
Closes #3743


tomscholz:
See https://github.com/Dogfalo/materialize/pull/4556/

PR #3750, opened 2016-10-08
Lean overlay missing for some reason, added it

Levitas:


acburst:
The modal plugin was rewritten recently so there is no more lean overlay

l2aelba:
@acburst  But JS still using `.lean-overlay` ??

PR #3752, opened 2016-10-08
Added feature to chips

lsxliron:
Added an option to allow duplicates in chips


JanieHuang:
Going to bump this. Any reason this isn't merged in? Would be really useful 

@Dogfalo

PR #3804, opened 2016-10-21
Allow tab navigation through the calendar days

bumbu:


tomscholz:
Can you explain a little bit more what you did?

bumbu:
Sorry, I'm not exactly sure what it does as it was a long time ago. 
But from what it seems - it adds `tabindex` to each calendar day. This way you can use tab key (on keyboard) to navigate through days and respectively when you keypress on Enter - that day will get selected. Basically this should allow selecting a day using only keyboard or voiceover commands.

If you think that it doesn't work/make sense or that's a wrong behaviour - feel free to close this PR as I'll not be able to work any further on this PR. Thanks!

Dogfalo:
Added manually in cdbb8bb4a8f5d1d040749abe0cb814775dd9098c. Sorry for not merging this earlier, I ended up just making the one line change myself due to the fact that the file has changed a lot since then.

PR #3808, opened 2016-10-21
Improve calendar accessibility

bumbu:
Remove aria-activedescendant from active day as it is used wrongly.
Should be set on a widget and should have an ID as a value 
https://www.w3.org/WAI/GL/wiki/Using_aria-activedescendant_to_allow_changes_in_focus_within_widgets_to_be_communicated_to_Assistive_Technology

Replace role `presentation` with `row` as widget with role `grid` should have nested `row`s, and elements with role `gridcell` (days) should have a `row` as a parent.


Dogfalo:
We no longer use pickadate.js as a dependency

PR #3815, opened 2016-10-24
restore images to origin state.

forehalo:
We should not just give images css attributes empty value, because there may be inline styles used.
Store all styles we have changed in data property and restore them after return to 'original' which looks the same as real origin.


Dogfalo:
Fixed in v1-dev

PR #3818, opened 2016-10-24
autocomplete event rename

adamasantares:
The "change" event is common for inputs, the better way to have special event name for "select" event.

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -350,7 +350,7 @@
           // Set input value
           $autocomplete.on('click', 'li', function () {
             $input.val($(this).text().trim());
-            $input.trigger('change');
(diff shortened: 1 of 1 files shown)


acburst:
I'm not sure I understand? We are triggering the change event because it is common. The change event is triggered when inputs are changed normally but not through javascript so we are triggering the same event to make it consistent with what you would expect.

adamasantares:
Hi, I mean, that "change" event triggers on input change, it's general behavior of text input, so, how I can detect that the "change" event fired by select in dropdown list? May be user have entered text but haven't selected.

Dogfalo:
We fire change because changing the input text using JavaScript does not fire the change event

Issue #3077, opened 2016-04-24
<input> tags rendering with hard to click <i> around it


actuallymentor:
When implementing a form (not using materialize form markup) the rendering makes the button hard to click. Specifically a wrapping <i> makes the button nonresponsive to clicking outside the actual text.

Source html:

`<input id="one" class="answer col l10 offset-l1 m12 s12 waves-effect waves-light btn" type="submit" value="I Understand and Agree"></input>`

Rendered html in chrome:

`<i class="answer col l10 offset-l1 m12 s12 waves-effect waves-light btn waves-input-wrapper" style=""><input id="one" class="waves-button-input" type="submit" value="I Understand and Agree"></i>`

Clicking the <i> gives a wave effect but does not trigger the form.

derekrliang:
See the issue here, along with a button submit:
https://codepen.io/derekrliang/pen/GZBbyw

Any particular reason you want to use a input instead of a button for form submit?

Dogfalo:
Your markup with the `i` tag wrapping seems very strange, place the waves effect on the button directly instead as @derekrliang does in his example

PR #3820, opened 2016-10-25
Correct flat-btn and wave effect combination

shawndibble:
Currently, if you have a flat-btn with the wave effect, then you must click directly on the button's words whereas the wave effect extends a bit outside the button. This can lead the user to think they clicked the button (thanks to the wave effect) when they really did not click the button at all. This change makes it so the clickable area includes the full wave effect area.

### DIFF
--- a/sass/components/_buttons.scss
+++ b/sass/components/_buttons.scss
@@ -1,6 +1,5 @@
 // shared styles
-.btn,
-.btn-flat {
+.btn {
   border: $button-border;
   border-radius: $button-radius;
   display: inline-block;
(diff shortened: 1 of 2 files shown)


acburst:
Can you first give a code example or an example issue that demonstrates this problem? The flat button example in our [docs](http://materializecss.com/buttons.html#flat) seems to be working fine

shawndibble:
The issue comes up when you do form submission input buttons (such as submit or reset).

`<input type="submit" class="btn-flat waves-effect waves-indigo">`  will turn into `<i class="btn-flat waves-effect waves-indigo waves-input-wrapper"><input class="waves-button-input" type="submit" value="Update"></i>` in the browser after all the javascript loads. 

With the current code base, only the input is clickable, but the `<i>` element gets all the padding. This results in a button that will give the illusion you clicked on it when it will do nothing unless you also clicked on the input element as well.

My changes addresses this instance and adds the padding on the input element so that if you get the wave animation on click, then the form will submit.

tuliren:
This change will be useful because the auto generated submission button in rails uses the `input` tag instead of the recommended `button` tag. This also answers the [comment](https://github.com/Dogfalo/materialize/issues/3077#issuecomment-214005119) in #3077.

PR #3821, opened 2016-10-26
Update _input-forms.scss to work on FireFox

IAMtheIAM:
On FireFox v49.0.1, the `label` gets fixed incorrectly to the right of the input field, instead of flushed with the input field at the left margin. 
Adding `left: 0` fixes this immediately, with no foreseeable consequences.

### DIFF
--- a/sass/components/forms/_input-fields.scss
+++ b/sass/components/forms/_input-fields.scss
@@ -116,6 +116,7 @@ textarea.materialize-textarea {
     content: "";
     position: absolute;
     top: 60px;
+    left: 0;
     opacity: 0;
     transition: .2s opacity ease-out, .2s color ease-out;
   }


IAMtheIAM:
http://screencast.com/t/S8ZDldGhaZ

seanwuapps:
Confirmed in firefox, a temporary solution is to use class like `input-field col s12` as the textarea's parent div (worked for me in ff).
This is a better solution to the problem.

IAMtheIAM:
@seanwuapps Your solution is fine, but it assumes that the user is using materialize's provided grid. The reason I found this bug is because I use a custom grid and don't use any of the col or s# tags. So I find that the better solution is to add float left as I showed here in this PR. In fact I don't see any disadvantages of this PR, as it allows users to use their own grid when materialize's grid is not adequate for the project needs.

seanwuapps:
@IAMtheIAM Hey I didn't make myself clear, I meat this PR is a better solution :)

PR #3824, opened 2016-10-27
Fix scrollspy watcher

### LINKED ISSUE #1080: Scrollspy target is not accurate anymore (0.96.0)
Self-explanatory, see screenshot:

image

There is a big gap when clicking on any Scrollspy link.

NitroBAY:
Scrollspy has the scrollOffset option but it only changes the place where the click leads you, for instance if you click to something with href="#test" it will lead you you `<h1 id="test">` with 0 offset, 200px before `id="test"` if you have offset of 200 etc. But the problem is that when one set scrollOffset it doesn't change the moment where the link is triggered.
So if you have, for instance, a scrollOffset of 0, if you click to #first on the table of contents the browser will scroll to id="first" but as by default Materialize add 200px to the position of element the table of contents indicates that you're on second

Exemple (I manipulate the DOM to show you as I have already patched your JS), I have a scrollOffset of 0, I click to "2eme sous chap" and so I go directly to "2eme sous chap" but on the table of contents the one which is emphasized is""Premiere partie" which is below which doesn't make any sense for user : http://img4.hostingpics.net/pics/610420materialize.png

It wil fix #1080, so people will not have to waste times any more by figuring out why there's a weird behavior with scrollspy and will stop changing source code.


acburst:
Thanks for pointing this out! I actually implemented this fix in a different way in 696bfc4.

NitroBAY:
Great minds think alike.

NitroBAY:
By the way why isn't dist/js/materialize.js updated by the commit but only bin/materialize.js is updated ?

PR #3836, opened 2016-10-31
spin-close FAB toggle

seanwuapps:
Added `spin-close` class for FAB to easily achieve the pretty animation of spin-close icon.

CodePen: http://codepen.io/deseanwu/pen/gwNjve

### DIFF
--- a/buttons.html
+++ b/buttons.html
@@ -230,6 +230,39 @@
   &lt;/div>
         </code></pre>
 
+        <br>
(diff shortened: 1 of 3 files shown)


LCallaghan84:
this would a nice addition

Dogfalo:
There are some small tweaks that I might make to this, such as making the first icon spin as well and changing the easing curve slightly, but overall this looks useful.

Issue #3835, opened 2016-10-30
link with disabled class can be clicked.


gidcs:
Is this a known issue?
The link of chevron_left is working although it has disabled class.
http://materializecss.com/pagination.html

theavinashofficial:
No man, I checked it and it is not working. 
Check my pen
http://codepen.io/avinashsingh/pen/RobYgQ
Just giving it a disabled class it becomes disables and color fades i.e. becomes light grey.

gidcs:
Yes. It becomes light grey but it could be clicked.
Try again with this.
http://codepen.io/gidcs/pen/pNzxJb

avinash97:
Ya  Man I checked it and you are correct. It is an issue.
I have corrected this here
http://codepen.io/avinashsingh/pen/VmwWXv
@Dogfalo or @acburst has to just add this line
.disabled {
pointer-events: none !important;
}
in the class and it will be resolved.

developeranirudhprabhu:
@acburst Can your merge the commit and mark this as closed?

NonameSLdev:
Close this issue please. @fega @tomscholz

tomscholz:
Okay, I will close this for now. And BTW @TheAvinashSingh you totally copied my GitHub bio ;)

TheAvinashSingh:
I liked your bio @tomscholz so I made it mine too ;-)

PR #3856, opened 2016-11-03
Added text-align justify property and removed the error of disabled links can be clicked. #3835

TheAvinashSingh:
There is no option to justify align text.
There is kinda class named "toast" which justifies text but gives a black background.
So I added a simple class named "text-justify" to align text to justify.

### DIFF
--- a/bin/materialize.css
+++ b/bin/materialize.css
@@ -2682,6 +2682,7 @@ video.responsive-video {
 .pagination li.disabled a {
   cursor: default;
   color: #999;
+  pointer-events: none !important;
 }
 
 .pagination li i {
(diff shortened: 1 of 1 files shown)


avinash97:
Will you take a look here @Dogfalo and @acburst .

tomscholz:
Have a look at our [CONTRIBUTING guide here](https://github.com/Dogfalo/materialize/blob/master/CONTRIBUTING.md) Never commit compiled files. We compile them with every version bump.

TheAvinashSingh:
Okay. I will take care of this.

tuxredux:
Definitely think this is needed... Perhaps change the class name to `justify-align` to go along with the [Text Align Helpers](http://materializecss.com/helpers.html) convention?

TheAvinashSingh:
Ya, there isn't any pre defined class available for CSS = text-align:justify;
They should add it with justify-align as you should.
@acburst Could you please check this out.

PR #3868, opened 2016-11-04
Fixes #3859

### LINKED ISSUE #3859: Responsiveness problem with the carousel
If you navigate back the carousel (no_wrap has to be false) the click event on the carousel item is being prevented. It happens becausethe center parameter is negative and it is being subtracted from here: 
```javascript
 var diff =  (center%count) - clickedIndex; //This is in the carousel's click function
```
this results in a negative diff (not equal 0) which causes the click event to be canceled. I managed to fix it by changing the line referenced above to:
```javascript
var diff = Math.abs(center % count < 0?center % count + count:center%count) - clickedIndex;
```
I don't have time to setup a developing environment and submit a pull request addressing this issue, that's why i post it here

Shivang-Bhandari:
A small change in the carousel js file as mentioned in issue #3859


tomscholz:
Hey there 👋 
```
/materialize/sass/
/materialize/js/ 
```

PR #3904, opened 2016-11-10
Enable custom checkbox fill colors using built-in color classing system

EPICmynamesBG:
This feature adds the color classing ability already set up (ex: purple lighten-1) in the framework to checkboxes so people aren't stuck with the default teal anymore.

I thought this had a fairly large use case as I frequently found myself changing the fill colors to match the theme I'd chosen. I already had the code made and figured others could benefit as well.

[CodePen Demo](http://codepen.io/mynamesbg/pen/aBdpMY)

### DIFF
--- a/sass/components/forms/_checkboxes.scss
+++ b/sass/components/forms/_checkboxes.scss
@@ -218,3 +218,23 @@ form p:last-child {
     border-color: $input-disabled-solid-color;
   }
 }
+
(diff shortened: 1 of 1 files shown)


Dogfalo:
I would suggest you use Sass or override the checkboxes with css manually as this will add a lot of css code

PR #3906, opened 2016-11-11
reIndent

lustyrain:
Hi
Please check it.

I looked editorconfig and your code
I think more miss indent.
It is not unified coding role.

Sorry.
I am not good at English.
If you are rude.
i'm so sorry.

Thanks
Lasty


Dogfalo:
Thank you!

PR #3912, opened 2016-11-15
Required Field Validation

acebruin:
- Fixed bug to catch required fields validation

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -58,7 +58,7 @@
       var $inputElement = $(this);
       var selector = ".prefix";
 
-      if ($inputElement.val().length === 0 && $inputElement[0].validity.badInput !== true && $inputElement.attr('placeholder') === undefined) {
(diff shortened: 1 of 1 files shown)


jtuberville:
This corrects an issue with the html5 "required" attribute not being recognized by the "validate" class handling logic.

Dogfalo:
I'm not entirely sure whether we should be tying the validation style to the `validate` attribute. Maybe @acburst can chime in as well on what he thinks.

I think this change might be good although it will definitely require some documentation changes if it goes through.

scote:
@acburst may you check this pull request? It would be nice to have this feature in next version.

tomscholz:
@acburst @Dogfalo Can you check on this please?

Dogfalo:
Merged similar PR here: #4883

acebruin:
#4883 doesn't have line 61 fix for the selector

Issue #3916, opened 2016-11-17
ScrollFire optimization


yanickrochon:
There are a few WTF in the scrollFire component.

1) it will run a timer indefinitely, regardless if all callbacks have been fired.
2) it will search and check the element's selector, regardless if all callbacks have been fired.
3) it will keep references to obsolete objects beyond their useful states
4) it will add duplicate window scroll listeners...

Bottom line is, this component is in dire need to be refactored. I can prepare a PR tonight.

acburst:
Do you think you could help post some code examples to test with?

yanickrochon:
You mean add tests cases? For now, my intention was to make it seamless with existing use cases, and didn't bother reading existing tests; I ran `grunt travis --filter=scrollFire` and it passed. Also note the timestamp of this issue and my PR; I basically wrote and pushed all that in 20 minutes, before leaving work, today :smile:

acburst:
Yeh no rush, Just wondering if you saved any of the code you used to discover these errors. Would be nice to have

yanickrochon:
No, actually, I was investigating another bug, and started reading the source code for scrollFire.... the bug I have has nothing to do with scrollFire, but I still investigated that avenue. I noticed that, in my code, I create a "load more" control to create an infinite scrolling list, and I was calling scrollFire many times because of that. I just wanted to prevent mem leaks and other timers being executed for no reason in the background.

yanickrochon:
I just realized that the tests do not run against the actual source, but against the compiled JavaScript?? This is just weird.... anyhow, I did all the tests (i.e. lint, compiling, running the server with ad-hoc testing, automated spec tests, etc.) and all seem to check out, now.

acburst:
Added in 0257ec7

yanickrochon:
You simply replaced setInterval with a throttled version. Many of the issues are not resolved, no.

PR #3917, opened 2016-11-17
Optimization and refactor of scrollFire

yanickrochon:
This commit solves issue #3916 and allow `Materialize.scrollFire` to be called any number of times. The component will always use a single timer and a single scroll listener. The component will also cleanup itself if it becomes inactive, and free any unused resources.

Code review welcome!

### DIFF
--- a/js/scrollFire.js
+++ b/js/scrollFire.js
@@ -1,48 +1,66 @@
 (function($) {
 
-  // Input: Array of JSON objects {selector, offset, callback}
+  var timerInterval = 100;
+  var handlers = [];
+  var listening = false;
 
-  Materialize.scrollFire = function(options) {
-
+  function startScrollListener() {
     var didScroll = false;
-
-    window.addEventListener("scroll", function() {
+    var scrollListener = function () {
       didScroll = true;
-    });
+    };
+    var timer = setInterval(funciton () {
+      if (didScroll) {
+        didScroll = false;
 
-    // Rate limit to 100ms
-    setInterval(function() {
-      if(didScroll) {
-          didScroll = false;
+        var windowScroll = window.pageYOffset + window.innerHeight;
 
-          var windowScroll = window.pageYOffset + window.innerHeight;
+        for (var i = 0, len = handlers.length; i < len; ++i) {
+          // Get options from each line
+          var value = handlers[i];
+          var selector = value.selector,
+              offset = value.offset || 0,
+              callback = value.callback;
+          var currentElement = document.querySelector(selector);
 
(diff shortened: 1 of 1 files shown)


acburst:
I decided to go in a different direction. I generalized the throttle function and used that instead of the setInterval.

Added in 0257ec7. Let me know if there are still problems with my approach

yanickrochon:
You still keep references to the options passed... This is creating memory leaks.

yanickrochon:
.. and you are still testing the element's bounding box before validating if you actually even need to test that element. I could've simply refactored my PR instead. You basically shut off all my PR for your own versions without discussing any matter. Do I really want to invest more time on this project for my PR?

kmmbvnr:
@yanickrochon Can you provide a codepen that demonstrates the issue?

yanickrochon:
@kmmbvnr I don't need to.

```js
Materialize.scrollFire({ selector: '.load-more', callback: function () { });
```

will always run in the background. *Always*. The component won't care if the event has been fire or not, [this line](https://github.com/Dogfalo/materialize/blob/master/js/scrollFire.js#L38) will always be executed for ever and ever. More over, during the lifecycle of the page, if another `.load-more` element is created (using React), then the component will *still* be called another time

```
Materialize.scrollFire({ selector: '.load-more', callback: function () { });
```
... it needs to, since the first one is flagged as [`done`](https://github.com/Dogfalo/materialize/blob/master/js/scrollFire.js#L29) (but is [still getting evaluated](https://github.com/Dogfalo/materialize/blob/master/js/scrollFire.js#L10) anyway!).

This PR, which is in conflict now because, instead of discussing, you guys decided to trash this one, addressed these issues.

kmmbvnr:
Mm, for me it looks too late to remove a scroll listener when onScroll event happened.

It's better to manage component state explicitly and remove listeners as soon as component been destroyed.

@acburst can we have an ability to destroy scrollSpy? Ex:

```javacript
  $('.selector'').scrollSpy('destroy');
```

That's like the MaterializeCSS already have for the selects `$('select').material_select('destroy');`

yanickrochon:
@kmmbvnr  sure! but we would need to create two constructors

1. `Materialize.scrollFire(options);` to support current implementations, keep BC, while implementing the fixes as I mentioned already. (Seriously, I want to contribute, but please don't go behind my PR and re-implement it... especially if the issues the PR addresses aren't resolved!) This API should warn that it is deprecated when used.
2. `$(selector).scrollFire(options)` the new API, bound to the element, where `$(selector).scrollFire('destroy');` may be called to remove all handlers and cleanup.

PR #3927, opened 2016-11-21
Adjust to hide side menu including shadow

tabass:


acburst:
Do you have any reasons for opening this?

Albeit a sloppy fix, this is to avoid having random box-shadows over your content if you have any on the side-nav

tabass:
In the current implementation, the side-nav shadow is displayed a bit on the left side of the screen when side-nav hidden.
I think that should not the shadow on the far left when side navigation is hidden.

PR #3936, opened 2016-11-24
Zombie.js headless browser support.

vbaranov:
Handling of undefined input.validity. This is the case for Zombie.js headless browser. 
Original issue: http://stackoverflow.com/questions/40634842/materialize-cannot-read-property-badinput-of-undefined.


chris-fa:
Can we merge this into the version that's previous to the one that broke the public API?

chris-fa:
bump

PR #3947, opened 2016-11-26
indent mistakes in form code examples. radio buttons and checkboxes

lsrock1:
Hello
There are some indent mistakes in radio buttons and checkboxes code examples.

So i fix it.


lsrock1:
edited wrong file

Issue #3955, opened 2016-11-29
Issue with Tabs Jumping Scroll Position


Anthall:
Hi,  I've set up some materialze tabs, when I click on each of the tabs it's jumping the scroll position of the page upwards as if it's got a page anchor on it. (See Screen Gif below). 

![tabs](https://cloud.githubusercontent.com/assets/24223210/20706728/93ef4276-b620-11e6-8d79-0d7280e2a404.gif)

Does anyone know whats happened here? This is my tab code...

```
<div class="row">
  <div class="col s12">
    <ul class="tabs bhivetabs"><li class="tab col s4"><a class="active" href="#tablets" target="_blank">Tablets and Capsules</a></li><li class="tab col s4"><a href="#drugs" target="_blank">Controlled Drug Liquids</a></li><li class="tab col s4"><a href="#amps" target="_blank">Ampoules</a></li>
    </ul>
  </div>
  <div class="col s12" id="tablets">
    <p>[ciscode|rev=1|tool=elmsmedia|item=1256|entity_type=node|render=link|field=field_image|text=Tablet Flow Chart|target=_blank|delta=0|item_type=image|alt=Tablet Flow Chart|width=933|height=489]&nbsp;
    </p>
  </div>
  <div class="col s12" id="drugs">
    <p>[ciscode|rev=1|tool=elmsmedia|item=1257|entity_type=node|render=link|field=field_image|text=Liquid Flow Chart|target=_blank|delta=0|item_type=image|alt=Liquid Flow Chart|width=933|height=489]&nbsp;
    </p>
  </div>
  <div class="col s12" id="amps">
    <p>[ciscode|rev=1|tool=elmsmedia|item=1258|entity_type=node|render=link|field=field_image|text=Ampoule Flow Chart|target=_blank|delta=0|item_type=image|alt=Ampoule Flow Chart|width=933|height=489]&nbsp;
    </p>
  </div>
</div>
```

tomscholz:
@Anthall Code you also provide a codepen? I can't reproduce your problem

Anthall:
I've just set up a codepen but it works fine here. This is the code i'm using though...

http://codepen.io/Hornetant/pen/LbzJbJ

It happens regardless of it being an image within the tab or just a line of text.

tomscholz:
@Anthall Is your website WordPress based/ are you using any anchor plugins, jquery... anything?

johann-abraham:
[...]

PR #3973, opened 2016-12-04
#3955 - Issue with Tabs Jumping Scroll Position

johann-abraham:
#3955 - Issue with Tabs Jumping Scroll Position

The reason for that was that the content for the previously active tab was being hidden before the content for the next tab was being shown, which was shortening the document.
    
As a fix the next tab content is shown first before the old tab content is hidden.
    
If the tabs are much different lengths then the document will still be shortened and there will still be some movement, but only if the bottom of the shorter tab is above the bottom of the screen


acburst:
Thanks!

PR #3979, opened 2016-12-07
Fix tooltip reposition within top/bottom position on the screen

philipsadman:
There are two mistakes you made in calculating top/bottom crossing:

– scrollTop – is function, but not a value
– you need to deduct scrollTop position to adjust new top position for element
– consider scroll left in calculation final left position for tooltip


developeranirudhprabhu:
@philipsadman I dont think materialize.js is the right file to change. There are specific files for each component. Please make appropriate changes to appropriate files. @acburst @Dogfalo please comment

philipsadman:
@developeranirudhprabhu revert changes made in bin folder

acburst:
Plugin rewritten

PR #3983, opened 2016-12-08
fix indent mistakes in form example

lsrock1:
Hello
I found some indent mistakes in form example and fix it.

to

I have edited .html page under jade/page-contents/.
Is that right?


Dogfalo:
Thanks!

PR #3999, opened 2016-12-13
Remove scrollTo selected element in multiple select

wackychocolatefactory:
Automatically scrolling to a selected element inside a multiple select causes a frustrating user experience if there are many elements. The current design assumes that a user will make selections from top to bottom.

enbvxch5tb


kmmbvnr:
yep, very annoying

smeijer:
Any idea on when the new release will be published? Can't wait for this to be fixed.

PR #4009, opened 2016-12-17
Made `new` badges do not append ` new` to the text

handicraftsman:
I think it would be better if it will not be auto-appended. For example, this will be great if it will be used for showing total count of X and changing color (style `new`) when there's new X

### DIFF
--- a/sass/components/_badges.scss
+++ b/sass/components/_badges.scss
@@ -18,9 +18,6 @@ span.badge {
     background-color: $badge-bg-color;
     border-radius: 2px;
   }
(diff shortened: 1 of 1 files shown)


handicraftsman:
@acburst -> ping

Kithio:
@handicraftsman So... You aren't really making any case about why it's better to do it the way you're suggesting here...

I see no reason to remove the appended "new" from the end since that's the purpose of that badge extension class.... You can easily remove it on your own local hosted version of it or just override it using your own css code

PR #4010, opened 2016-12-17
FAB buttons with behavior similar to IONIC 2

Cmte-Kirk:
My suggestion is to make the FAB buttons with the same behavior and appearance as the FAB buttons of IONIC 2.
Vertically up and down, horizontally to left or right, as well as the centered and positioned buttons.
An example can be seen in the IONIC 2 documentation.
[](https://ionicframework.com/docs/v2/components/#fabs)
[](https://github.com/driftyco/ionic-preview-app/blob/master/src/pages/fabs/basic/template.html)

Thanks


Dogfalo:
This PR is not very clean so I cannot easily see what changes you are making. Please reopen a separate pull request with only one commit

PR #4047, opened 2017-01-01
Fixed Permissions

ankitpati:
Many files have incorrect permissions. More specifically, they have the executable bits set. This is unnecessary, as none of them are executable. This PR unsets the execute bits.


Dogfalo:
Looks good, not sure how only some of those had improper permissions, but thanks for fixing it.

ankitpati:
You’re welcome. Thank you for the merge.

Chances are, those files were authored on a system that does not understand Unix permissions. Like Windows.

PR #4050, opened 2017-01-02
Parallax speed option

aleksander351:
Adding a parallax speed option.
Default value is `100`.

If you want only half of the speed of your parallax effect, write 50 as a parameter into the parallax initialization. 
Example:
```
$(document).ready(function(){
  $('.parallax').parallax(50);
}); 
```

If you want double the speed of your parallax effect, write 200 as a parameter into the parallax initialization. 
Example:
```
$(document).ready(function(){
  $('.parallax').parallax(200);
});
```

### DIFF
--- a/js/parallax.js
+++ b/js/parallax.js
@@ -1,6 +1,14 @@
 (function ($) {
 
-  $.fn.parallax = function () {
+  $.fn.parallax = function (speed) {
+    var parallaxSpeed = function (speed) {
+      if(isNaN(speed)){
+	      return 1;
+      }else{
+	      return (speed / 100);
+      }
+    };
+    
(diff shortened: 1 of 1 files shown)


acburst:
This doesn't really work that well because what our parallax does is make sure that the image is always shown. I think with this, you could have it at a speed where the image will move too quickly and a blank background will show up.

aleksander351:
I thought about this myself and I agree with you.
Either we make scroll speed available with a warning, that changing the speed might have negative consequences with regards to image visibility, or we add a remark to the documentation, that if someone wants to adjust the speed of the parallax, he/she should manipulate the size of the image, which automatically adjusts the speed of the parallax effect.

PR #4062, opened 2017-01-03
Introduce responsive horizontal cards

tobias74:
By using the markup of the normal horizontal card with the new class "responsive-horizontal" the card will display as a horizontal card only on medium an up devices. On small devices the card will be displayed as a regular card since horziontal cards do not make much sense on small devices anyway.

### DIFF
--- a/sass/components/_cards.scss
+++ b/sass/components/_cards.scss
@@ -98,6 +98,45 @@
     }
   }
 
+  // Responsive Horizontal Cards
+  @media #{$medium-and-up} {
+    &.responsive-horizontal {
+      &.small, &.medium, &.large {
+        .card-image {
+          height: 100%;
+          max-height: none;
+          overflow: visible;
+
+          img {
+            height: 100%;
+          }
+        }
+      }
+
+      display: flex;
+
+      .card-image {
+        max-width: 50%;
+        img {
+          border-radius: 2px 0 0 2px;
+          max-width: 100%;
+          width: auto;
+        }
+      }
+
+      .card-stacked {
+        display: flex;
+        flex-direction: column;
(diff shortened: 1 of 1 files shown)


developeranirudhprabhu:

sass/components/_cards.scss
```diff
@@ -98,6 +98,45 @@
...
+  @media #{$medium-and-up} {
+    &.responsive-horizontal {
+      &.small, &.medium, &.large {
+        .card-image {
+          height: 100%;
+          max-height: none;
+          overflow: visible;
+
+          img {
+            height: 100%;
```
darioalbanesi: I would go with

`img {
  max-height: 100%;
  width: auto;
}`

see: https://codepen.io/italopupo/pen/bRNqZw

tomscholz:
flexbos has only limited support in ie. Can you check if it is working there?

alexjose:
Any update on this? Can this be merged?

tobias74:
I will check this in IE this evening

tobias74:
I checked it in Edge 40 and in IE11, works correctly in both browsers. I currently don't have any machine with IE10 or below, so I cannot verify this PR on those browsers...

tobias74:
Concerning the flexbox-property: I basically just copied the original css of the regular card and put it in a media-query. So this "horizontal-responive-card" is just a combination of a regular card and a horizontal card, switched by a media-query. Therefore it should display correcty whenever the original cards are working -- at least that's what I hope :-)

Dogfalo:
We'd rather not add too many new features for now since we're transitioning into 1.0 beta release soon. After 1.0 we'll evaluate it's place in the framework

Issue #4066, opened 2017-01-04
Utility Classes: Left and Right Align need to be only medium and up


manicmarvin:
### Description
The Utility classes `.left-align` and `.right-align` should not apply to `small-and-down` screen sizes, this is because all text on small screens centers itself, and right-aligned, left aligned text defying this is ugly.

**Solution:**
```
@media only screen and (min-width: 601px){
  .left-align {
  text-align: left;
  }

  .right-align {
    text-align: right;
  }
}
```

ben-sykes:
While what you're proposing isn't necessarily incorrect, I believe text alignment should be application specific. What you consider "ugly" is largely due to the default text-alignment not working with your application's overall design. For a different application, with its own design, this may not work. For example, what if you need to consider LTR vs RTL languages? Your change could throw everything off. 

I disagree that this should be part of the library, although your snippet would fit in nicely with other site-specific overrides to the materialize library.

manicmarvin:
I shouldn't have used the word _ugly_, it was lazy of me, I meant that it produced unexpected inconsistencies when you move from larger screens to smaller ones. However, you're absolutely right, I never considered RTL languages. 

While this might make the codebase much bulkier, could we add separate utility classes that specifically don't interfere with small screens (and/or medium screens) - think `.left-align-s-down` and `.left-align-m-down`. Obviously, these are poor names. 

But I fear that this kind of expansion of utility classes with unnecessarily bulk up the library.

Dogfalo:
We want to keep utility classes very simple and in this case making these particular ones not work on only mobile may be confusing.

PR #4067, opened 2017-01-04
Update _global.scss

manicmarvin:
My commit fixes the issue I pose in #4066


tomscholz:
We want to keep utility classes very simple and in this case making these particular ones not work on only mobile may be confusing.

PR #4079, opened 2017-01-08
Fix order of dialog buttons being flipped

### LINKED ISSUE #4078: The order of dialog action buttons are flipped
### Description
Dialog action buttons have a float: right on them which flips their order, which means it shows Agree on the left and Disagree on the right. This goes against material guidelines for [modal actions](https://material.io/guidelines/components/dialogs.html#dialogs-specs), which says that "Dismissive actions are always placed directly to the left of affirmative actions." This issue appears in the materialize documentation as well.

### Repro Steps
Go to http://materializecss.com/modals.html and click on either/both modal.

Or here's a live example http://codepen.io/codypearce/pen/LxVaZP

### Screenshots / Codepen
material-dialog

materialize-modal

materialize-modal2

codypearce:
Fixes #4078 

Removed the float: right on the action buttons and added text-align-right on the modal footer container.


developeranirudhprabhu:

UncleWalt:
Hi,
this would also fix another problem:
If more than one button is used, the float right removes the space between the buttons (caused by a whitespace character between the button tags), which looks very ugly:

http://codepen.io/anon/pen/Kmzjxd
(I used raised buttons for clarity)

tomscholz:
Hey there 👋 
```
/materialize/sass/
/materialize/js/ 
```

Issue #3685, opened 2016-09-23
Callback triggers


larsvansche:
It has happened several times now that I had to search the materialize.js file and add a trigger after a certain event, just to make sure I can perform code after a materializeCSS component was finished loading/running. 

My suggestion is that triggers get added after every event, so we can write callbacks ourselves. 

I'm not sure if this is the right place to post this, please inform me if it isn't.

acburst:
We've added callback function parameters to some of our components. Can you point out specific ones where you need them?

larsvansche:
I ran in to a lack of callbacks twice: once with tabs and once with pushpin. I added triggers in the materialize.js file that I can call with $("element").on("trigger", function() { /* do something */ });. 

Besides those two, I can't say for which components I will need callbacks in the future.

acburst:
There is a callback parameter for tabs. Is that not sufficient? And I can look into adding some for  pushpin

tomscholz:
Closed due inactivity. Feel free to reopen it, if it's still necessary.

PR #4080, opened 2017-01-08
Adds Pushpin callbacks

marcusmalmberg:
Adds options for callbacks when Pushpin enters the different states: `pinned`, `pin-top`, `pin-bottom`. 

I needed a way to trigger callback when the element was pinned, and "unpinned". Seeing I wasn't alone (https://github.com/Dogfalo/materialize/issues/3685) I decided it to make a pullrequest instead of just monkey-patching my local version.


developeranirudhprabhu:

tomscholz:
Hey there 👋 
```
/materialize/sass/
/materialize/js/ 
```

PR #4085, opened 2017-01-10
Added data-external for external source of materialboxed image showed…

ivellios:
… with overlay; Also added example to documentation;

Materialbox does not support external source, so showing thumnails that will display overlayed full-size image once clicked is impossible. 

I found PR with concept of solution for that: https://github.com/Dogfalo/materialize/pull/2798
But the commit contains some bugs (like it does not change src to thumbnail in returnToOriginal) and is based on the older version of materialize repo.

I used that version to create one that actually works. I added `data-external` as the attribute for img with url to the full-size image. As in the PR above, it uses progress bar while loading full-sized image. 

Also I fixed a bug, where `data-caption` was checked against `!== ""` (old line 113) so when it is not set at all (i.e. `undefined`) it was created anyway and moved it's hiding complete event code regarding origin `img` to origin `complete` event function (new lines 265-296 are moved from old 246-273).

Click event was moved to new function `enlargeImage` and now click checks if the image has external source (in which casse loads it in place of thumbnail) or just launches `enlargeImage` and works as it used to previously. 

I added example section called External image source into documentation after Caption section on JavaScript -> Media page. 

Hope someone will find it useful. 

**Known issues (maybe someone can help with these):** 
- preferably image should do some fadeOut/In animation when returnToOriginal (and maybe while enlarging)
- related to the one above - when image is loaded for a short moment `img` element is being resized, but then size is changed to thumbnail and animation starts. Maybe there's a way to load image in the element of contant size. 

I think it is a good start to this feature. If you can help with the issues above, please contribute.

### DIFF
--- a/js/materialbox.js
+++ b/js/materialbox.js
@@ -56,8 +56,15 @@
             progressWrapper.remove();
           }).attr("src",origin.data("external")); // switch src of thumbnail with original
           imageChanged = true; // now the image is switched
-        }else
+        }else{
+          originalWidth = origin.width(); // remember fullsize file size
+          originalHeight = origin.height();
+          placeholder.css({
+            width: placeholder[0].getBoundingClientRect().width,
+            height: placeholder[0].getBoundingClientRect().height,
+          })
           enlargeImage(); // works like before
+        }
       }); // End origin on click
 
 
@@ -85,8 +92,6 @@
 
         // Set positioning for placeholder
         placeholder.css({
-          width: originalSrcWidth,
-          height: originalSrcHeight,
           position: 'relative',
           top: 0,
           left: 0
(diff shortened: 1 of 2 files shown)


js/materialbox.js
```diff
@@ -16,14 +16,42 @@
...
+      var originalSrcHeight = origin.css("height");
+
+
 
       origin.on('click', function(){
+        if (!imageChanged && origin.data('external')){
+          // if thumbnail has original version of image
+          progressWrapper.insertAfter(origin); // show progress while new src loads
+          origin.load(function(){ // will enlarge image once new src is loaded
+            enlargeImage(); // enlarge loaded src as normal
```
ivellios: There should be also added:

```
            originalWidth = origin.width();
            originalHeight = origin.height();
            origin.css({
              width: originalSrcWidth,
              height: originalSrcHeight
            });
```
So the loaded image will be resized right after loading to the thumbnail size and the resized to full size with overlay. Makes animation look little bit more smooth. Still placeholder changes it's size for a moment. I still think about a workaround.

developeranirudhprabhu:

PR #4088, opened 2017-01-11
Color function optimization

TheGrimSilence:
## Changes
- Optimized color() in `_color.scss`
- Removed unneeded quotations across `$colors` map and all calls to `color()`
- Combined color and shade class functions

### DIFF
--- a/sass/_style.scss
+++ b/sass/_style.scss
@@ -25,18 +25,18 @@ h5 > span {
 }
 
 nav {
-  // background-color: color("materialize-red", "lighten-2");
+  // background-color: color(materialize-red, lighten-2);
 
   a {
     -webkit-font-smoothing: antialiased;
   }
   ul li a:hover, ul li.active {
-    background-color: color("materialize-red", "lighten-1");
+    background-color: color(materialize-red, lighten-1);
   }
 }
 
 .header {
-  color: color("materialize-red", "lighten-2");
+  color: color(materialize-red, lighten-2);
   font-weight: 300;
 }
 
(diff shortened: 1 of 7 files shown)


developeranirudhprabhu:
@acburst and @Dogfalo can you please look into it?

TheGrimSilence:
Consider it a small gift from Ebongarde Studios😉

acburst:
You got rid of our warning for an incorrect color input. Have you tested input errors?

TheGrimSilence:
Given that the colors are highly based on Google's color pallet the chance of 'new' and or 'changed' colors are low. But, if you feel cautious it can be added. Warnings are easy to avoid as long as you know what you're doing.

Dogfalo:
It helps people debug if they have a typo in the color function

TheGrimSilence:
Warning added back, code is safe.

acburst:
That part of the code will never be executed? That was why we checked beforehand if the key existed so we could correctly return null with a warning. Right now it will only return null.

TheGrimSilence:
If there's a typo, it'll give a warning. The error is as originally typed. Code runs fine. Same functionality.

acburst:
I've just tested this with 3 cases:
 

```
color(xxx, base);
```
Which results in a compilation error.

```
color(xxx);
```
Which compiles, but does not raise the warning (which it should).
```
color(teal, xxx);
```
Which compiles, but does not raise the warning (which it should).

TheGrimSilence:
Hm. I'll fix it
.

TheGrimSilence:
Warning didn't work because I forgot to add the parent `@if map-has-key($colors, $color)`

developeranirudhprabhu:

tomscholz:
@DanielRuf Your thought on this, too?

PR #4089, opened 2017-01-11
Add select dropdown style for navbar

jnevesjunior:
Adding style for dropdown when placed into the navbar.

I just added the styles, if removing the example css would look like original.
Example:
http://codepen.io/anon/pen/vgLvdZ


developeranirudhprabhu:

Issue #2949, opened 2016-03-23
Uncaught TypeError: $indicator.velocity is not a function


lvarayut:
I installed the plugins using npm and bundled it using webpack. The plugin was required as following:

``` js
export default class App extends React.Component {
   ...
  componentDidMount() {
    require('materialize-css/dist/js/materialize');
  }

  render() {
    return (
        <div className='row'>
          <div className='col s12'>
            <ul className='tabs'>
              <li className='tab col s3'><a href='#test1'>Test 1</a></li>
              <li className='tab col s3'><a className='active' href='#test2'>Test 2</a></li>
              <li className='tab col s3 disabled'><a href='#test3'>Disabled Tab</a></li>
              <li className='tab col s3'><a href='#test4'>Test 4</a></li>
            </ul>
          </div>
          <div id='test1' className='col s12'>Test 1</div>
          <div id='test2' className='col s12'>Test 2</div>
          <div id='test3' className='col s12'>Test 3</div>
          <div id='test4' className='col s12'>Test 4</div>
      </div>
    );
  }
}
```

I always got the `Uncaught TypeError: $indicator.velocity is not a function` error when clicking on Tab . However, if I required the plugin in HTML instead, it worked just fine. I'm not quite sure what's going wrong.

benelliott:
I ran into the same issue. You just need to make sure you attach jQuery to the window when you `require` it:

```javascript
window.jQuery = window.$ = require('jquery');
```

This should be imported before `materialize`.

linuxenko:
Looks like there have to be `velocity-animate` package installed and exported as following:

```js
window.$.velocity = require('velocity-animate/velocity.js')
```

only then i get it work

enmanuelr:
Thanks @benelliott. So many hours of head banging...

Wifsimster:
Adding
```
window.jQuery = window.$ = require('jquery');
window.$.velocity = require('velocity-animate/velocity.js')
```
doesn't fix the velocity bug anymore on v0.97.7 :-1:

CharlieGreenman:
@Wifsimster agreed this is no longer a viable fix.

MIKUScallion:
I use it on v0.97.7
```js
// Check for jQuery.
// TODO: Fix $(dom).velocity is not function bug
if (typeof(jQuery) === 'undefined') {
  // var jQuery;
  // Check if require is a defined function.
  if (typeof(require) === 'function') {
[...]

PR #4092, opened 2017-01-11
Update JQuery version from 2.1.4 to 2.2.4 because webpack problems

saintplay:
jQuery 2.1.4 is a major problem when you install Materialize from npm and use it in webpack bundle:
```js
require('materialize-css')
```

The workaround is deleting manually the node_modules folder.  
I think jQuery dependency needs to be updated

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   },
   "dependencies": {
     "hammerjs": "^2.0.4",
-    "jquery": "^2.1.4",
+    "jquery": "^2.2.4",
     "node-archiver": "^0.3.0"
   },
   "engine": "node >= 0.10",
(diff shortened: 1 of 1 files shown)


Dogfalo:
Any idea why it works on jquery 2.2.4? What was the original error?

saintplay:
Is related to #2949 
Got the same error with jQuery 2.1.4

ZiperRom1:
Only one strange error on CI check, can this be fixed and merged ?

mastrauckas:
I too am having this same problem.

tomscholz:
https://github.com/Dogfalo/materialize/pull/4587

PR #4094, opened 2017-01-12
Making sure that the indicator is always last

BenjaminVadant:
Hi,

In order to be sure that the indicator is always the last element of the ul group, we append it each time.

This is useful for example when you are looking for the next/prev tab using jquery.prev/next

If the indicator is left in the middle of the ul group, using jquery.prev could lead to a selection of the indicator instead of the li tag, thus making the selection more complicated.

Benjamin

### DIFF
--- a/js/tabs.js
+++ b/js/tabs.js
@@ -88,6 +88,8 @@
         $this.append('<div class="indicator"></div>');
       }
       $indicator = $this.find('.indicator');
+      // we make sure that the indicator is at the end of the tabs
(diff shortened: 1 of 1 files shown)


acburst:
Do you have an example of when the indicator would not be last?

BenjaminVadant:
Here is a [jsfiddle](https://jsfiddle.net/benjaminvadant/ypybh8vg/)

acburst:
Looks good to me. Thanks!

PR #4110, opened 2017-01-17
[Fix] - fix label from blocking

ANewHunter:
fix label from blocking the users from accessing the input when clicked on the label inside the input.


acburst:
This is not a valid solution as we want the label to still be selectable. If the label has the correct for attribute it shouldn't block access to the input

PR #4113, opened 2017-01-18
adjust height of extended fixed nav

lsrock1:
problem: when i want to use fixed navbar with Extended Navbar with Tabs,
tab contents are covered by tab buttons. Because .navbar-fixed is just set to $navbar-height or $navbar-height-mobile without tab's height.(content's div tag goes up to bottom of navbar)

code: http://codepen.io/anon/pen/egBqZR

So i make new class, navbar-extended-fixed. Its height is set $ navbar-height + $tabs-height to widen the height and cover all the tab area.
In variables.scss file i add new variable, $tabs-height. Because tabs-height is accessed in navbar.
It is possible that some changes in tab height can break nav's css.
This tag can be used exactly same way like navbar-fixed.

after code: http://codepen.io/anon/pen/mRONQy

### DIFF
--- a/sass/components/_navbar.scss
+++ b/sass/components/_navbar.scss
@@ -195,6 +195,15 @@ nav {
     position: fixed;
   }
 }
+.navbar-extended-fixed {
+  position: relative;
+  height: $navbar-height-mobile + $tabs-height;
+  z-index: 997;
+
+  nav {
+    position: fixed;
+  }
+}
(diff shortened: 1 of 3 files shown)


lsrock1:
With fixed-nav sidenav doesn't work. sidenav-overlay covers everything. But when i change the z-index of fixed-nav 997 to 998, it works.

Issue #4125, opened 2017-01-24
Materialize Website footer tweet count cutoff


patrickdundas:
### Description
Regarding the public Materialize website, found at www.materializecss.com:

The footer section holding the share links for github, google plus, and twitter cuts off the share count for tweets.

### Repro Steps
To reproduce the issue:

1. Visit www.materializecss.com
2. Scroll to the footer section of the page
3. Under the "Connect" section, look at the twitter share button
4. The share count arrow appears, but the actual share count rectangle is not attached to the flyout

### Screenshots / Codepen
![materializeconnectglitch](https://cloud.githubusercontent.com/assets/6844039/22232995/dd73386a-e1ba-11e6-8034-6bebdd5cf56e.PNG)

HazyNightmare:
Just sent in a PR fixing this issue.

acburst:
Merged

PR #4127, opened 2017-01-24
Issue #4125 fix

HazyNightmare:
A small css rule change to allow the twitter follower count box to be fully seen, rather than being cut off


acburst:
thanks!

PR #4134, opened 2017-01-25
Remove horizontal scroll of side-nav

serartmar:
The line "padding: 0 45px 0 15px" caused horizontal scroll of the whole side-nav on desktop and mobile.


Dogfalo:
We've since fixed this.

PR #4162, opened 2017-01-31
Adding class for compact buttons

gewisser:
For the small screen, the button group wants to shrink into a single line. Adding a Class "comapct" to the text button will maximize to do it

Demonstrate  https://jsfiddle.net/ak5yuf00/12/


gdoron:
This really feels like a thing you need to override in your **personal** CSS files and not put it for everyone.

Dogfalo:
Yes I think especially in this case, since this is something that is easily customized for most people, if they want smaller buttons than standard spec they will have little trouble adding it themselves.

PR #4185, opened 2017-02-04
Update _select.scss

j1016h:
Add below css: 
    input.select-dropdown.active {
        visibility: hidden;
    }

This is to fix a form select issue specific to iPhone. A blue focus cursor will show up when you open form select component.


j1016h:
Close PR. It break form select on desktop version

PR #4198, opened 2017-02-07
Keep 'active' class until next scrollSpy:enter

shin-szk-engineering:
If using ScrollSpy on a long long page, the 'active' class does not appear in table-of-contents when entering between sections.
So I fixed 'active' class to keep it until the next scrollSpy:enter called.

### DIFF
--- a/js/scrollspy.js
+++ b/js/scrollspy.js
@@ -179,6 +179,10 @@
 				visible.push($(this));
 			}
 
+			// Remove other 'active's before adding
+			$('.scrollspy').each(function() {
+				$('a[href="#' + $(this).attr('id') + '"]').removeClass('active');
+			});
 
 			$('a[href="#' + visible[0].attr('id') + '"]').addClass('active');
 		});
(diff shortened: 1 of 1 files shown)


Dogfalo:
I am not seeing this problem on our long forms page: http://materializecss.com/forms

shin-szk-engineering:
I always thank you for using such a wonderful framework:)

Thanks for your advice, I found the cause of this problem.
The forms page contained a description in each .scrollspy divs.

In my case, I didn't put anything other than .raw.col div in .scrollspy div.
.raw.col div is styled `float: left;` by _grid.scss, then it makes .scrollspy div's height 0.
cf. http://stackoverflow.com/questions/5369954/why-is-my-divs-height-zero

And you know, scrollspy.js is checking element's height.
It's why I couldn't see 'active' class between .scrollspy sections.

If someone wants to use scrollspy with 0 height div, there are 2 solutions.
One way is to rewrite scrollspy.js like my PR.
Another way is to set `overflow:hidden;` on .scrollspy div like the answer of stackoverflow.

Whatever, do you think this is a problem to be solved?

Issue #4202, opened 2017-02-08
Make scrollFire resettable [bug] [meteor]


NitroBAY:
Hello,
### Description
Calling `scrollfire` again does not reinit the scrollfire function. A normal app doesn't need to recall scrollfire but with Meteor when you click on a link another template is created and the old one is destroyed.
Example I have a homepage template and a login one. The user scrolls and animations are working great (scrollfire triggers `showStaggeredList`). The user clicks on a link that leads to the login page from the homepage. He comes back to the homepage right after that but the problem is that even tough `scrollfire` is called again `scrollFire` doesn't triggers any animation anymore so half of the homepage is invisible.

### Pull request
For a possible implementation (I tested it)  :
 See #4206

acburst:
ScrollFire is going to be removed in v1 due to lack of complexity and out of project scope. There are plenty of libraries that are more specialized to achieve the same thing but much more robustly.

PR #4206, opened 2017-02-08
Make scrollFire resettable (e.g useful for Meteor)

NitroBAY:
### How it works : 
Delete scrollFireEventsHandled that was preventing, on purpose, scrollFire to be called more than one time.

I tested it and it solves the issue described in #4202 on Meteor.


NitroBAY:
@Dogfalo, @acburst what do you think about it ?

tomscholz:
Hey there 👋 
```
/materialize/sass/
/materialize/js/ 
```

acburst:
WIth v1, you can now destroy and reinit

PR #4208, opened 2017-02-09
swipe back when tab is disabled

Xaber20110202:
and remove window.pre_index


tomscholz:
Can you explain a little bit more what you have done?

Xaber20110202:
1. `index = prev_index = 0`  will add window.prev_index, so we should remove it.
2. when tabs carousel to new_index tab, but that new_index tab is disabled, so we should make it back.

PR #4242, opened 2017-02-15
Avoid TextField labels from getting pointer events

romgrk:
Currently, the text fields labels can be selected. It also prevents the user from focusing the input when they click directly on the label, because it receives the pointer events.


acburst:
This is not a valid solution as we want the label to still be selectable. If the label has the correct for attribute it shouldn't block access to the input

PR #4247, opened 2017-02-16
Fixed tap target position with scrollbars

inventti-leonardo:
I have been tested the tap target feature and got this issue.

### DIFF
--- a/js/tapTarget.js
+++ b/js/tapTarget.js
@@ -90,11 +90,13 @@
       var originWidth = origin.outerWidth();
       var originHeight = origin.outerHeight();
       var originTop = isFixed ? origin.offset().top - $(document).scrollTop() : origin.offset().top;
-      var originLeft = origin.offset().left;
+      var originLeft = isFixed ? origin.offset().left - $(document).scrollLeft() : origin.offset().left;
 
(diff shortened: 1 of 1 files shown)


acburst:
Can you describe the issue more or help me reproduce the problem?

inventti-leonardo:
I saved a codepen using files from git (bin/materialize.css, bin/materialize.js and feature-discovery.html):
http://codepen.io/anon/pen/GrLwgW

To reproduce, click on "open tap target" and will show a target misaligned, because of scroll

inventti-leonardo:
I worked to fix both horizontal and vertical scroll when tap target is showed.
Can you review/refactor/merge @acburst ?

acburst:
Thanks! this works well

adick:
It got a lot better with the above [commits](https://github.com/Dogfalo/materialize/pull/4247/commits), but there was still an issue on mobile devices. The height of the visible/invisible address bar causes the windowHeight to be off. 

To fix this, change line 98 to: 

```
// var windowHeight = $(window).height();
var windowHeight = window.innerHeight ? window.innerHeight : $(window).height();
```

PR #4249, opened 2017-02-16
before complete hook for modal

haojia321:
This is a before modal close hook. Sometime we need this to get a better UI interaction experiences.
jQuery dialog plugin also has it.
http://api.jqueryui.com/dialog/

### DIFF
--- a/js/modal.js
+++ b/js/modal.js
@@ -13,6 +13,7 @@
         inDuration: 350,
         outDuration: 250,
         ready: undefined,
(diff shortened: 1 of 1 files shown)


haojia321:
Can we merge this enhancement please? This hook can improve interaction experiences on mobile a LOT!

Dogfalo:
in 1.0, we'll make all applicable animations have before and after hooks

PR #4254, opened 2017-02-17
Added missing files (pulse.scss & taptarget.scss) to packages.js to let meteor work properly again

tomscholz:


tomscholz:
Note: When using the repo as a local meteor package, meteor gives an error because pulse.scss & taptarget.scss are requested from materialize.scss, but aren't imported by Meteor

acburst:
Thanks!

PR #4278, opened 2017-02-22
Changed (when possible) HTTP urls to HTTPS + change google icons url + mobile.html url

tomscholz:
- Did this, because 2017 and stuff
- New Material url: https://material.io/icons
- Fixed mobile.html url because in navbar it was set to static url which sucked hard in Development or when offline


tomscholz:
@Dogfalo @acburst Please merge or tell me why not.

PR #4296, opened 2017-02-27
Show result in autocomplete even if equal to query

bokub:
In the autocomplete plugin, when you type the whole word you want to select, it just disappears form the suggestions, making it impossible to select.

Check out [this jsFiddle](https://jsfiddle.net/gfa496sj/) for a full example with a more detailed explanation.

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -363,8 +363,7 @@
               if (val.length >= options.minLength) {
                 for(var key in data) {
                   if (data.hasOwnProperty(key) &&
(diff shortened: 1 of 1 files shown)


bokub:
Hi, do you plan to merge anytime soon ?
That's just one line of code, but it fixes a really annoying bug

fega:
@bokub personally, I think that a test should be provided.

bokub:
You're right, here is the same [jsFiddle](https://jsfiddle.net/gfa496sj/1/), but now fixed with the new code

PR #4299, opened 2017-02-28
'.fixed' sidenav with closeOnClick true closes on large screens (issue #1426)

abinthomasonline:
fix issue #1426 - fixed sidenav with closeOnClick enabled

'.fixed' sideNav closes when closeOnclick:true in large screens.
http://codepen.io/anon/pen/LWpRdM 

Solution: check `menu.hasClass('fixed')` and screen width before `removeMenu();`


judasane:
The problem is back with the new materialize version. There is not property `closeOnClick` for the navbar

PR #4336, opened 2017-03-10
Added events onOpen/onClose for SideNav

jpbonsigno:
New methods for SideNav elements for catch open and close events.


tomscholz:
Added in https://github.com/Dogfalo/materialize/commit/8a01666817bd6f6f69b7a5b08b1a3f1d1677f1ee

PR #4364, opened 2017-03-20
Waves: Custom selectors in options and sass update

knitevision1:
To keep your markup clean and keep presentational layer in the stylesheets, I thought it makes sense to expand the functionality a bit by adding an option item that expects an array of classnames of elements that you want to have Wave attached.

In addition to that, I've added some placeholders in SASS to access styles more hassle-free.


Dogfalo:
Seems like a nice case that isn't a problem to most people

PR #4391, opened 2017-03-24
RTL Support

arminmon:
### RTL Support as a component
All scss modifications related to RTL support are placed in sass/components/rtl_support.
It works out of the box for almost every component and element by just adding .rtl class to their container parents.

My goal was to add RTL support alongside the vanilla LTR base at the same time with no conflicts, which can be useful for multi-language web designs.

**These components are not supported yet** (might not even need to be!)**:**
* forms/range
* slider
* carousel
* date_picker/default
* date_picker/default.date
* date_picker/default.time

### DIFF
--- a/sass/components/rtl_support/_badges.scss
+++ b/sass/components/rtl_support/_badges.scss
@@ -1,0 +1,17 @@
+// RTL Badges
+.rtl {
+	span.badge {
+		margin-left: initial;
+		margin-right: 14px;
+		float: left;
+
+		&.new:after {
+			content: " جدید";
+		}
+	}
+
+	nav ul a span.badge {
+		margin-left: initial;
+		margin-right: 4px;
+	}
+}
--- a/sass/components/rtl_support/_badges.scss
+++ b/sass/components/rtl_support/_badges.scss
@@ -4,10 +4,6 @@
 		margin-left: initial;
 		margin-right: 14px;
 		float: left;
-
-		&.new:after {
-			content: " جدید";
-		}
 	}
 
 	nav ul a span.badge {
--- a/sass/components/rtl_support/_grid.scss
+++ b/sass/components/rtl_support/_grid.scss
@@ -1,0 +1,93 @@
+// RTL Grid
+.row.rtl, .rtl .row {
+	.col {
+		float: right;
+		$i: 1;
+		@while $i <= $num-cols {
+			$perc: unquote((100 / ($num-cols / $i)) + "%");
+
+			&.s#{$i} {
+				margin-right: auto;
+			}
+			$i: $i + 1;
+		}
+		$i: 1;
+		@while $i <= $num-cols {
+			$perc: unquote((100 / ($num-cols / $i)) + "%");
+
+			&.offset-s#{$i} {
+				margin-right: $perc;
+			}
+
+			&.pull-s#{$i} {
+				left: $perc;
+			}
+
+			&.push-s#{$i} {
+				right: $perc;
+			}
+			$i: $i + 1;
+		}
+		@media #{$medium-and-up} {
+			$i: 1;
+			@while $i <= $num-cols {
+				$perc: unquote((100 / ($num-cols / $i)) + "%");
+
+				&.m#{$i} {
+					margin-right: auto;
+				}
+				$i: $i + 1;
+			}
+			$i: 1;
+			@while $i <= $num-cols {
+				$perc: unquote((100 / ($num-cols / $i)) + "%");
+
+				&.offset-m#{$i} {
+					margin-right: $perc;
+				}
+
+				&.pull-m#{$i} {
+					left: $perc;
+				}
+
+				&.push-m#{$i} {
+					right: $perc;
+				}
+				$i: $i + 1;
+			}
+		}
+		@media #{$large-and-up} {
+			$i: 1;
+			@while $i <= $num-cols {
+				$perc: unquote((100 / ($num-cols / $i)) + "%");
+
+				&.l#{$i} {
+					margin-right: auto;
+				}
+				$i: $i + 1;
+			}
+			$i: 1;
+			@while $i <= $num-cols {
+				$perc: unquote((100 / ($num-cols / $i)) + "%");
+
+				&.offset-l#{$i} {
+					margin-right: $perc;
+				}
+
+				&.pull-l#{$i} {
+					left: $perc;
+				}
+
+				&.push-l#{$i} {
+					right: $perc;
+				}
+				$i: $i + 1;
+			}
+		}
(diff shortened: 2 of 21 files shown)


sass/components/rtl_support/_badges.scss
```diff
@@ -0,0 +1,17 @@
+// RTL Badges
+.rtl {
+	span.badge {
+		margin-left: initial;
+		margin-right: 14px;
+		float: left;
+
+		&.new:after {
+			content: " جدید";
```
NonameSLdev: Why the content: "جدید"?

tomscholz: It meas new in the persian language

tomscholz: @arminmon Is only the persian language rtl?

NonameSLdev: @tomscholz Also in Arabic, and yeah I know, I was meaning to ask why should the arabic text be included in the official materialize library?
@arminmon It should be removed anyways so it isn't in the official library.
P.S: Tom, it's everything rtl (by looking at the code).

arminmon: @tomscholz Persian, Arabic and Hebrew are RTL for sure.
@NonameSLdev I removed that line. Maybe That "new" content for badge shouldn't be included at all. It seems more flexible if it was fetched as a data-attrib in CSS like this: content: attr(data-badge-text);

tomscholz: @arminmon I guess you are right. I will create a PR that fixes that. The idea with the attribute sounds great but there is currently no or limited support for that feature :( http://caniuse.com/#feat=css3-attr

tomscholz: This is actually a huge internationalization problem...

sass/components/rtl_support/_grid.scss
```diff
@@ -0,0 +1,93 @@
...
+				}
+
+				&.push-l#{$i} {
+					right: $perc;
+				}
+				$i: $i + 1;
+			}
+		}
+	}
+	&.reverse > {
```
arminmon: @tomscholz @NonameSLdev 
This class is added in case you want to change the floating behavior of the .cols inside a .row.
eg, You have RTL content inside a grid but you want to pull columns to left (keeping the deeper tags responding to the parent's .rtl class and prevent adding extra .rtl class to nested tags).

Dogfalo:
I'm not sure this is the best way to add RTL to this project. It would be better as a auto-generated file that is  independent of the current code

arminmon:
@Dogfalo The most practical way is to auto-generate RTL classes with a Sass function but there are problems:
* Not every CSS attribute should be swapped like _float: left_ to _float: right_. Doing so, will break the layout for as I've tested it many times already. And it's not only about the _float_, there are _text-align_, _border_, _left_, _right_ & etc which should be modified in the proper way.
* RTL support should be generated within each component, which will modify all files, otherwise **even if** there was a global function, it should've gone through every single component everytime something changes.
Building a global function that could generate something like I've done, would be the best way to go. But it just seems that there are too many variables.

sass/components/rtl_support/_badges.scss
NonameSLdev: @tomscholz @arminmon We could just use javascript for that :)

arminmon: @NonameSLdev I'm not sure if JS is the most optimized solution for this case. We're talking about static content here, therefore it won't worth it to bother client's memory load with JS.
Data-attribute would be the best option as soon as it gets well-supported by all browsers, till then we could either load each data-attribute into the tag with JS or let the developer set their own "phrase" inside a span tag.

NonameSLdev: @arminmon You're probably right, but it's just one automatic call when the page loads and another method triggered manually by developer, so it wouldn't take so much memory.

tomscholz:
Hey! The more I think about this, I realize it is a bad realization of RTL, unfortunately. Like @Dogfalo said, a auto-generated solution would be better, because if we merge this, every contribution to the project have to include the same changes for the rtl version. This would end up in a huge chaos. Maybe you should consider of just making a supported rtl fork of materialize. Have you looked at the implementation of rtl at foundation [here](http://foundation.zurb.com/sites/docs/rtl.html) and [here](http://foundation.zurb.com/sites/docs/rtl.html) ?

arminmon:
@tomscholz I had quick look at what they're doing for RTL support. It seems quite easy.
https://github.com/zurb/foundation-sites/blob/develop/scss/util/_direction.scss
The hard part is that we'll need to implement this into the whole project for once and then we can change the direction at any time by changing a variable.
We could start working on it, if you're interested.

sepehr1014:
Any news about RTL support?

kmmbvnr:
Btw there is the [rtlcss](http://rtlcss.com/learn/usage-guide/value-directives/) tool that looks pretty configurable. It looks that just a few changes required to convert MaterializeCSS.

arminmon:
Here's a live example of a materialized RTL website.
http://filoode.netlify.com/
Everything looks fine too me so far.

JafarAkhondali:
Will we have rtl support soon? I've found some hacky forks in internet but no one contributes to them.
If it needs a hard work, set a leader for roadmap and we'll absolutely help in contributing.

arminmon:
@JafarAkhondali
I've been working on it for a while (see the live example) but due to some issues (real life) I've been held back from development.
The current version works quite well, no bugs.
I'll be updating it to the latest Materialize version (0.99.00) soon.
The main problem is how it should be implemented. I'm working on a global sass function to get around this issue, but it's gonna take a while. Till then, the current approach seems to be working well enough.

JafarAkhondali:
@arminmon I appreciate you suggestion, but as you said you were busy and that's why you shouldn't do the RTL version. There would be changes to code every time and it would be hard for you to match yours with new version. Also with lack of community to test your code many bugs might be hidden withing your fixes. I still think it would be better if they do this themselves. Until then, I'm gonna use your repo.

arminmon:
@JafarAkhondali
You're probably right. The other reason that I've holding back from the development is that Materialize is still in development phase an the stable version is not out yet. Till then it'd take too much work to keep up with development from the outside.
پ.ن.
خدمت زورکی بد دردیه. وقتی نمونده تو زندگیه ما😔

JafarAkhondali:
@arminmon Yep. I hope we see the released version soon :)
درکت میکنم ... منم خدمت دارم‌:(

Dogfalo:
We aren't looking to support RTL in the near future as this is a large undertaking which will require maintenance

PR #4394, opened 2017-03-24
Add interactive playgrounds to documentation

smileytechguy:
Hi,
This is the first of (hopefully) many additions I would like to add.

What this modification does is add a section to the color.html page which allows the user to edit the classes on a sample element through a simple element.  
It resides directly below the usage category.

image

By editing the text in the class input field, a user can change the classes of the box (asides from the card-panel class.

This is done through vanilla JavaScript.

It does not require any specific ordering of classes in the input field.

Demonstration: 

materialize_color_playground

I didn't add any additional tests, as I was not changing any distribution files, only documentation.

I'm considering adding a "open in CodePen" link or similar.  What do you guys think?

### DIFF
--- a/color.html
+++ b/color.html
@@ -148,27 +148,6 @@
         </div>
       </div>
 
-      <div id="sass" class="section scrollspy">
-        <h2 class="header">Sass</h2>
-        <p>
-        For background colors, you can apply the color simply by extending the classes like the example below.
-        </p>
-        <pre><code class="language-scss col s12">
-  .ilike-blue-container {
-    @extend .blue, .lighten-4;
-  }
-        </code></pre>
-
-        <p>
-        For changing text color, you can apply the color simply by extending the classes like the example below.
-        </p>
-        <pre><code class="language-scss col s12">
-  .ilike-blue-container {
-    @extend .blue-text, .text-lighten-4;
-  }
-        </code></pre>
-      </div>
-
       <div id="playground" class="section scrollspy">
         <h2 class="header">Try It</h2>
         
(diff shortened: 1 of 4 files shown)


smileytechguy:
Added interactive playground for CSS grid documentation  
Allows addition and removal of elements in order to test offset, push, pull, etc.

Currently resides at the bottom of the grid page

image

In the future, I may attempt making each of the examples further up the page interactive.

Demonstration:

materialize_grid_playground

smileytechguy:
I modified it some, now you can click on the box directly and edit it
image

smileytechguy:
In addition, I made it so most of the other examples on the grid.html page is editable

tomscholz:
The idea is cool. But I don´t love it. I know this sounds dumb but I´m not sure if this playground is "too much". Maybe there could be another page called playground where you have your little playground, but this editing thing is to compliacted for starters and then whe will have like a tousand issues why this isn´t working. Don´t get me wrong, please. The idea is great and everything, but I guess a codepen integration or something like [this](https://material.angularjs.org/latest/demo/bottomSheet) here would do the job as well. 
Best wishes from Hamburg 👋

smileytechguy:
@tomscholz I understand the concerns, and maybe the inline edit is confusing, as the descriptions give the premade examples context.  However, I think the "Try It" would be a pretty cool feature, and I'm not sure how you would do it all on one page, apart from just rendering HTML as inputted.

fega:
Personally, I like the idea, but it looks quite invasive done in that way. I agree with @tomscholz 

@smileytechguy maybe the exercises could be grouped in one page with a dropdown where you can select the example and edit it.

Also in that way you avoid over-complexity in the documentation.

acburst:
Maybe instead of having a separate try it playground in every example, there could be a dedicated try it page. And we can add special links to this page that populate it with the current code example?

tomscholz:
There are other priorities right now. Maybe in the future when we do a rework of the documentation we could add something like it.

PR #4402, opened 2017-03-26
bug fix sidebar.fixed li a

interaminense:
add an *display: block* for line break

*before:* 
before

*after:* 
after


interaminense:
@Dogfalo You just corrected this problem! Thanks man 👏👏👏

PR #4407, opened 2017-03-26
Removed radial reaction on switches from l/xl screens

NonameSLdev:
Radial reaction on switches (levers) now only on S/M screens, as only they need it according to the material guidelines: https://material.io/guidelines/components/selection-controls.html#selection-controls-switch

### DIFF
--- a/sass/components/forms/_switches.scss
+++ b/sass/components/forms/_switches.scss
@@ -59,14 +59,26 @@
 // Switch active style
 input[type=checkbox]:checked:not(:disabled) ~ .lever:active::after,
 input[type=checkbox]:checked:not(:disabled).tabbed:focus ~ .lever::after {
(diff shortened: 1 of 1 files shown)


Dogfalo:
You are misunderstanding the spec, the radial reaction isn't tied to screen size, but the type of interaction pointer used. According to the spec, it should appear when it is triggered through touch interaction and is not necessary in desktop mouse interaction.

However while this interaction is technically the correct way of implementing it, I don't like complicating things with JavaScript when the deviation from the spec here is not largely detrimental.

NonameSLdev:
@Dogfalo How about including a mobile/tablet browser detector in materialize and using it to remove radial detection on desktop browsers? It's not such a big difference, and it'd be pretty lightweight

Issue #4378, opened 2017-03-22
Tap Target has problems using it on mobile devices.


luchrv:
### Description
On mobile devices **[Tap Target](http://materializecss.com/feature-discovery.html)** does not retain the position of the element to which it refers. Tends to move to the center of the screen.

### Repro Steps
You should only see this feature on a mobile device.

darckyn:
Same here.

Hainminsoe:
yes , same here

tomscholz:
![image](https://cloud.githubusercontent.com/assets/20381749/24307790/2e771500-10ce-11e7-8556-951ef4b2060a.png)

tomscholz:
I think FeatureDiscovery should look like this on mobile:
![img_0348](https://cloud.githubusercontent.com/assets/20381749/24333387/3d71bfa8-1257-11e7-9940-ac0af251d5bf.PNG)

inventti-leonardo:
You can fix this issue with css:

```
@media only screen and (max-width: 600px) {
  .tap-target, .tap-target-wrapper {
    width: 600px;
    height: 600px;
  }
```

Try here http://codepen.io/anon/pen/dvqzjQ

tomscholz:
Still. If this would be full screen on mobile it would be pretty neat

PR #4416, opened 2017-03-27
Fix tap target for small devices

inventti-leonardo:
From issue https://github.com/Dogfalo/materialize/issues/4378


acburst:
Thanks!

luchrv:
Thanks!

PR #4438, opened 2017-03-29
Include SRI attributes for public CDN-hosted URLs

jonwolski:
When referencing code on a CDN, one should provide hashes for "sub-resource integrity" checks. (cf. https://www.w3.org/TR/SRI/). According to http://caniuse.com/#search=sri, nearly 50%–60% of browsers support this (backwards-compatible) check.

(Technically, SRI attributes will always provide value, but that is especially the case when you reference resources on a third-party domain such as a shared CDN.)

I generated these hashes with https://www.srihash.org/


tomscholz:
And we have to regenerate them with each release. Right?

jonwolski:
@tomscholz, right, that's the challenge. It's easy for me to add this to a one-time pull-request, but it needs to be operationalized. i.e. Whatever build process exists for this documentation should generate the SRI hashes instead.

Issue #4400, opened 2017-03-26
Scrollbar appears for tabbed menu


friek108:
### Description

The horizontal scrollbar for the tabs menu keeps appearing when you navigate to the last menu item on a desktop/laptop browser.

Please refer to the recent discussion here from Feb 11: https://github.com/Dogfalo/materialize/issues/2584#issuecomment-279027006

### Repro Steps
When clicking on the last tabbed menu item on a desktop, the horizontal scrollbar sometimes appears. There are 2 scenarios in which this happens:

1. **Sometimes** when navigating to the last tab, the horiztonal scrollbar appears. You can't scroll this horizontal tab menu, but the scrollbar appears. Have a look at this comment for a screenshot https://github.com/Dogfalo/materialize/issues/2584#issuecomment-279027006

1. If navigating **from** a tab that a) isn't the last tab and b) has a lot of content that requires the vertical page scrollbar to be present, and then you click on the last tab that doesn't require a vertical page scrollbar.  What happens is 1) The tabs width gets set, 2) the vertical page scrollbar is removed by the browser 3) a horizontal tabs scrollbar appears that allows you to scroll the width of the page scrollbar. Refer to this comment https://github.com/Dogfalo/materialize/issues/2584#issuecomment-287696155

acburst:
Hmm, I added some rounding to the indicator positions in 50c6718

Let me know if this fixes the issue

bui:
50c6718 does not fix the issue on my end. For me, the horizontal scrollbar appears when switching from a tab that causes the document to be scrollable due to its length, to a tab that causes the document scrollbar to disappear (ie. less content), if that makes sense.

acburst:
Do you think you could recreate this in a codepen? I'm finding it hard to reproduce

bui:
https://codepen.io/anon/pen/EWeVeN

tomscholz:
@acburst It is easier to reproduce on Windows

friek108:
I've tested the codepen on windows 10 in Chrome, Edge and Firefox on windows 10 and can confirm it is reproducing the issue.

NonameSLdev:
Confirmed in firefox.
The problem is that tabs are [automatically scrollable](http://materializecss.com/tabs.html) scrollable by default, which is correct according to the material guidelines (Because the two tab types are scrollable and fixed).
[...]

PR #4441, opened 2017-03-29
Removed scrollbar from fixed width tabs

NonameSLdev:
Removed scrollbar from fixed width tabs - as it should be. Fixed width tabs shouldn't be scrollable anyways, _and_ this solves https://github.com/Dogfalo/materialize/issues/4400 .


asia653:
What's the status of this?

PR #4480, opened 2017-04-02
Fix issue in pulse when using non-floating buttons

Pierstoval:
When using `.pulse` with non-`.btn-floating` buttons, there was a big glitch on the screen.

This is fixed in this PR, with no impact on current buttons:

Before:

screen recording 2017-04-02 at 08 52 50 pm

After:
screen recording 2017-04-02 at 08 52 10 pm

I also updated the examples so we can see that it works for every button now

### DIFF
--- a/jade/page-contents/pulse_content.html
+++ b/jade/page-contents/pulse_content.html
@@ -7,14 +7,14 @@
 
         <div style="display: flex; justify-content: space-around;">
           <a class="btn btn-floating pulse"><i class="material-icons">menu</i></a>
-          <a class="btn btn-floating btn-large pulse"><i class="material-icons">cloud</i></a>
+          <a class="btn btn-large pulse"><i class="material-icons">cloud</i></a>
           <a class="btn btn-floating btn-large cyan pulse"><i class="material-icons">edit</i></a>
         </div>
 
         <h4>Pulse HTML Structure</h4>
         <pre><code class="language-markup">
 &lt;a class="btn btn-floating pulse">&lt;i class="material-icons">menu&lt;/i>&lt;/a>
-&lt;a class="btn btn-floating btn-large pulse">&lt;i class="material-icons">cloud&lt;/i>&lt;/a>
+&lt;a class="btn btn-large pulse">&lt;i class="material-icons">cloud&lt;/i>&lt;/a>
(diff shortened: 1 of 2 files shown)


jade/page-contents/pulse_content.html
```diff
@@ -7,14 +7,14 @@
 
         <div style="display: flex; justify-content: space-around;">
           <a class="btn btn-floating pulse"><i class="material-icons">menu</i></a>
-          <a class="btn btn-floating btn-large pulse"><i class="material-icons">cloud</i></a>
+          <a class="btn btn-large pulse"><i class="material-icons">cloud</i></a>
```
DanielRuf: do not change it here, provide a codepen with a demo of the fix

tomscholz:
Pulses are not designed for rectangular buttons. I think we should prevent that class from working on rectangular buttons.

DanielRuf:
Pulse is only for round buttons afaik (according to the Google specs)

Pierstoval:
Here's a JSFiddle example of the bug and its fix in the `CSS` panel:

https://jsfiddle.net/x37yLb9o/4/

Even though it's not in Google's specs, I still think this should be implemented in Materialize to avoid people having huge glitches when using pulse with non-`.btn-floating`

acburst:
While I agree its not used in the spec, I think its better to leave the use up to the user. I'll work on merging this but I don't think I'll be changing the docs because I don't want to promote use on non floating buttons.

tomscholz:
But but... It looks like shit. Like I said before, we should just prevent that class from working on rectangular buttons

Pierstoval:
Then if you all agree, I can change my PR and only add restriction for non-floating buttons, but I more like @acburst's proposal to allow the use of pulse for non-floating buttons and remove the documentation change I made

acburst:
I'm merging right now. And as a framework, I think we should strive to be as flexible as possible. Some people may want to use it in a different way and they can add custom css to make it look better.

tomscholz:
OK, but @Pierstoval can you remove that part than from the docs?

acburst:
Merged, don't worry I took care of the docs

Pierstoval:
Thanks @acburst

DanielRuf:
That's a good solution. =)

Issue #4318, opened 2017-03-04
sideNav with fixed desktop nav doesn't work correctly


evanhdc:
### Description
I'd like to make a fixed Navbar for desktop view, as well as navbar for mobile view. This is what I did
https://fiddle.jshell.net/evanhdc/v1d7fk9h/

### Repro Steps
In the above fiddler, resize browser window so the mobile navbar shows up.
Click menu on top left so the sideNav pops up.

Problem is that in mobile view, when I clicked the menu button, sideNav shows up with a sidenav-overlay that covers the entire page including the sideNav so I can't click any option.

The fiddler works fine with materializecss ver 0.97.7 but the issue starts from v0.97.8, and is still happening in v0.98.0.

In order to make sidenav-overlay to cover only the correct portion of the page, I have to remove the outer div with class "navbar-fixed" but this way, the desktop navbar is not fixed anymore. 

### Screenshots / Codepen
Add supplemental screenshots or code examples. Look for a codepen template in our Contributing Guidelines.

b-z:
I have the same problem, while using a fixed navbar.
Maybe it is caused by the z-index.

My solution is move this <ul> out of the navbar-fixed:
`<ul class="side-nav" id="slide-out">`

tomscholz:
These issues are all duplicates:
https://github.com/Dogfalo/materialize/issues/1664
https://github.com/Dogfalo/materialize/issues/3918
https://github.com/Dogfalo/materialize/issues/4046
https://github.com/Dogfalo/materialize/issues/4231
https://github.com/Dogfalo/materialize/issues/437
https://github.com/Dogfalo/materialize/issues/4318
https://github.com/Dogfalo/materialize/issues/3982
https://github.com/Dogfalo/materialize/issues/2879
https://github.com/Dogfalo/materialize/issues/4187

For the future discussion, please use this issue instead.
https://github.com/Dogfalo/materialize/issues/3844

Issue #4231, opened 2017-02-13
Using fixed navbar causes issue in mobile view


ptmaroct:
### Description
To make the navbar fixed, you have to add an outer wrapping div with the class navbar-fixed. This will offset your other content while making your nav fixed.
This works fine for desktop browsers but when mobile view is working on small screen devices and navbar becomes a sidebar, opening the sidebar causes a blackout of whole screen.
TLDR; navbar-fixed isn't compatible with mobile view.
### Repro Steps
I just wrapped the navbar from example given in docs, with a .navbar-fixed div
Here is a pen to demonstrate the issue.
http://codepen.io/ptmaroct/pen/vgvoqG

tomscholz:
Duplicate: https://github.com/Dogfalo/materialize/issues/3844

tomscholz:
These issues are all duplicates:
https://github.com/Dogfalo/materialize/issues/1664
https://github.com/Dogfalo/materialize/issues/3918
https://github.com/Dogfalo/materialize/issues/4046
https://github.com/Dogfalo/materialize/issues/4231
https://github.com/Dogfalo/materialize/issues/437
https://github.com/Dogfalo/materialize/issues/4318
https://github.com/Dogfalo/materialize/issues/3982
https://github.com/Dogfalo/materialize/issues/2879
https://github.com/Dogfalo/materialize/issues/4187

For the future discussion, please use this issue instead.
https://github.com/Dogfalo/materialize/issues/3844

Issue #4187, opened 2017-02-05
0.97.8 and 0.98 break mobile side-nav, 0.96 works fine


jpwynn:
For my Rails 5 app, the nav layout below (in HAML) works fine with 0.97.6 of the materialize (downloaded manually and added to assets folders), specifically the **mobile side-nav** works fine that is displayed when you click the hamburger icon when you make the browser window narrow.

However the exact same code displays a broken mobile side-nav when I upgraded the js and scss to 0.97.8 or 0.98. Downgrading the back to 0.97.6 made it work again.

Was there a breaking change to how side navs get rendered after 0.97.6? (Looking at the nav section of the current materializecss.com website I don't see a difference.)

Symptoms: side nav slides from the left as expected, but the side-nav background is gray (not white) and all of the links on the sidenav seem disabled... no highlighting on hover, nothing happens when click the link. (If I had to guess, it looks like the overlay that 'dims' the rest of the page when the mobile side-nav slides in is actually 'on top' of the side-nav too, dimming it and preventing hover and click events. Interestingly, the side-nav is fine if you add the 'fixed' class, though of course it's always-on in that case. Point being, there seems to be something related to the overlay that dims the rest of the page that's wonky.)

I went back and forth between versions a couple of times with zero other changes to the code case, it is definitely the  0.97.8 or 0.98 that breaks it.

**That said, if you do _not_ make the menu fixed, eg, if you remove %div.navbar-fixed THEN the mobile side-nav works for both 0.98 and 0.97.6.**

Behaves the same whether using Chrome and Safari under Mac OSX Sierra

    %body
        %header

          %div.navbar-fixed

            %nav
              %div.container.city_bg  # adds a bg image to top nav bar
                %div.nav-wrapper

                  %div.brand-logo
                    // Replace this with an image
                    = image_tag "jlogo1b.png", style: "height: 60px;"

                  %a{href: "#", class: "button-collapse", data: {activates: "mobile_side_menu"}}
                    %i.material-icons menu

                  // Normal NavBar that goes across the top
                  %ul#nav-mobile.right.hide-on-med-and-down

                    %li
                      =link_to "Link 1"
                    %li
                      = link_to "Link 2"
                    %li
                      = link_to "Link 3"

                    // Drop Down Menu
                    %li
                      %a{href: "#", class: "dropdown-button", data: {activates: "myDropdown"}}
                        = "Drop Down Menu"
                        %i.material-icons.right 
                          arrow_drop_down
[...]

Issue #4046, opened 2016-12-31
SideNav overlay cover .side-nav when in .navbar-fixed


MacGyer:
### Description
When you render the source for your side nav (.side-nav) inside the "normal" navbar, especially a fixed navbar (.navbar-fixed), the #sidenav-overlay covers the whole page including the actual sidenav. This makes the side nav not usable.

The issue occurs, because the .side-nav is embedded in .navbar-fixed. So the .navbar-fixed and #sidebar-overlay have the same z-index (997).

Any chance to fix the issue? Or is it required to place the sidebar outside of (any) .navbar?

### Repro Steps
View the pen below.

### Screenshots / Codepen
http://codepen.io/anon/pen/egOdPK

MacGyer:
Update: seems to work on 0.97.7. Just set the codepen JS and CSS file to 0.97.7 to confirm.

TheColorRed:
I am seeing this issue on the latest version.

shmshd12:
@MacGyer You have already given almost fix. 
Just add this to your custom styles.css file :
`
.navbar-fixed {
    z-index: 9999;
}`
I check this in your given Codepen link and this worked fine.

MacGyer:
@shmshd12 I just checked the pen. It is configured for 0.97.8 and your suggestion is integrated.
Does not work.

shmshd12:
@MacGyer check this out : http://codepen.io/anon/pen/GrRrVP
This codepen is copy paste of your with addition of `.navbar-fixed { z-index: 9999; }`.

Check if your styles.css is below `materialize.min.css`. If it is at top` styles.css` will not overwrite `materialize.min.css` file.

MacGyer:
@shmshd12 sorry, my bad. Your solution indeed works. (had the CSS comment not closed).

Thanks for providing. Nevertheless I think, this issue should be solved by the Materialize CSS itself.

c0u1d:
I got the same issue with my github pages,this should be solved by materialize itself not overwrite it!

shmshd12:
@c0u1d Above given fix is a temporary one. Yes, this issue will not be closed until Materialize fix the issue.

risendev:
If you adjust the z-index in .sidenav-overlay on /sass/components/_sideNav.scss line 208 from 997 to 996 then it will permanently resolve this issue.

TheColorRed:
[...]

PR #4482, opened 2017-04-02
Added missing fixed navbar documentation

### LINKED ISSUE #1664: navbar-fixed AND mobile collapsable menue.
Hi!
This is a promising framework, I like it!
I wonder though, it seems like there is no solution for having both the `navbar-fixed` and also use the `nav-mobile` `button-collapse`. As of now, the z-indexes is disturbing each other. 
##### The error:

When the collapse mobile nav is pulled out, the fixed-navbar (at the top) is "above" the faded area. 
##### My code

Here is my code that causes the problem.

```
<div class="navbar-fixed">
  <nav class="light-blue lighten-1" role="navigation">
    <div class="nav-wrapper container"><a id="logo-container" href="#" class="brand-logo">Logo</a>
      <ul class="right hide-on-med-and-down">
        <li><a href="#">Navbar Link</a></li>
      </ul>

      <ul id="nav-mobile" class="side-nav" style="left: 0px;">
        <li><a href="#">Navbar Link</a></li>
      </ul>
      <a href="#" data-activates="nav-mobile" class="button-collapse"><i class="material-icons">menu</i></a>
    </div>
  </nav>
</div>
```

I hope you guys will take a look at this soon because I think that the combination of `navbar-fixed` and the mobile slide out menu will be very common. 
Or perhaps am I just doing it wrong?

And again, sweet framework!

tomscholz:
resolve #1664, #3918, #4046, #4231, #437, #4318, #3982, #2879, #4187, #3844

### DIFF
--- a/jade/page-contents/navbar_content.html
+++ b/jade/page-contents/navbar_content.html
@@ -211,6 +211,28 @@
     &lt;/nav>
   &lt;/div>
         </code></pre>
+
+        <br>
+
+        <h5>Fixed Navbar in combination with SideNav component</h5>
+        <pre><code class="language-markup">
+  &lt;div class="navbar-fixed">
+    &lt;nav>
+      &lt;div class="nav-wrapper">
+        &lt;a href="#!" class="brand-logo">Logo&lt;/a>
+        &lt;a href="#" data-activates="slide-out" class="button-collapse">&lt;i class="material-icons">menu&lt;/i>&lt;/a>
+        &lt;ul class="right hide-on-med-and-down">
+          &lt;li>&lt;a href="sass.html">Sass&lt;/a>&lt;/li>
+          &lt;li>&lt;a href="badges.html">Components&lt;/a>&lt;/li>
+        &lt;/ul>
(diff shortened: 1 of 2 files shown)


joshmorel:
Hi,

I also think you should add a note in the documentation on the option closeOnClick: true.

This is option enabled to true is the only thing that makes this workable for same-page anchors IMHO. JavaScript:

```
$(document).ready(function(){
    $(".button-collapse").sideNav({closeOnClick: true});
});
```

acburst:
Is this necessary? most of the code is duplicated. Can we add a note instead about the placement of the side-nav?

tomscholz:
@acburst Interesting question. I get the thing with the duplicate code, but most people don't even read the docs properly and just copy paste. I would be okay with a note, but if the issue still comes up, we could put the full code example in the docs. Ok?

ScottBeeson:
In case anyone finds this before it is published, like I did, just move your second `ul` outside of the `.navbar-fixed` element.

Issue #4410, opened 2017-03-26
Offset is applied for all larger screens


TetianaP:
### Description
Column offset is applied for specified screen and all larger ones.

### Repro Steps
Create simple row with 2 columns:
```
<div class="row">
	<div class="col s12 m6 offset-m3 l8"></div>
	<div class="col s12 m6 offset-m3 l4"></div>
</div>
```

### Expected
For **large** screen content is placed in 2 columns: l8 and l4.
`Offset-m3` is applied only for **medium** screen.

### Result
For **large** screen content is placed in 1 column with left-column-offset: `offset-m3`.
Is there some way to clear offset or specify something like: offset-l0?

Y0lk:
Noticed the same thing, but it was fine in 0.98.0

Y0lk:
Here's the PR that fixes it : https://github.com/Dogfalo/materialize/pull/4483

acburst:
Hmm, I tried to do some refactoring which caused this bug. Sorry about that. Merged in a PR that fixes this

TetianaP:
@Y0lk, @acburst, thank  you!

PR #4483, opened 2017-04-03
Fixed grid offset bug

Y0lk:
See #4410
This appeared when xl breakpoints were added. margin-left, left, and right properties were removed from the media queries which is a problem because now offset-m (for example) would also be applied to large and xlarge.

This PR just reverts the code back the way it was before + adds the properties for xl breakpoints


acburst:
Thanks!

Issue #4487, opened 2017-04-04
FeatureDiscovery doesn't align to the bottom on mobile devices (v0.98.1)


CodeSurfing145:
### Description
The FeatureDiscovery function does algin on Desktop/Laptop but not on mobile devices.

### Repro Steps
Open http://materializecss.com/feature-discovery.html and click on OPEN TAP TARGET
Now try again on Android phone (Not sure about iPhones bahh)

### Screenshots / Codepen
PC: 

![screenshot from 2017-04-04 08-37-53](https://cloud.githubusercontent.com/assets/6624832/24644086/0467d8de-1912-11e7-8622-5b1c35b05310.png)

Android:

![screenshot_20170404-083841](https://cloud.githubusercontent.com/assets/6624832/24644231/ac37ceb6-1912-11e7-8d60-5bc5f3011a85.png)

CodeSurfing145:
![screenshot_20170404-100202](https://cloud.githubusercontent.com/assets/6624832/24646911/e9ae86b2-191d-11e7-9bbc-60e246edefad.png)
Landscape mode seems to align it correctly except for the title and the text.

DanielRuf:
related, fixed width and height for all views under 600px
![image](https://cloud.githubusercontent.com/assets/827205/24655334/1feaf9ac-193e-11e7-9214-d59aaad1cf43.png)

inventti-leonardo:
@DanielRuf @CodeSurfing145 @fega  

I think was fixed here https://github.com/Dogfalo/materialize/issues/4378

DanielRuf:
This does not fix it, still the same since the last build https://github.com/Dogfalo/materialize/commit/0407ee5352271de822cade662353319f6eebe5c8

Also the 600px value seems probleamtic.

DanielRuf:
Open it in desktop and resize to a lower screen width
![image](https://cloud.githubusercontent.com/assets/827205/24658598/6ed0d256-194a-11e7-96f8-ab62e6566a33.png)

DanielRuf:
Directly opened in small view
![image](https://cloud.githubusercontent.com/assets/827205/24658658/a7b8387a-194a-11e7-8929-7ecd684893e1.png)

DanielRuf:
Different behaviour if you start opened or closed.

acburst:
It looks like you are viewing the current site right? This issue has been fixed but is not in the current release. It won't show up in the actual site until we release a new version.

[...]

PR #4488, opened 2017-04-04
Fix tap target when rotate the device

inventti-leonardo:
Issue was found here
https://github.com/Dogfalo/materialize/issues/4487#issuecomment-291526012

And fixed/test here
http://codepen.io/anon/pen/oZVrEQ

### DIFF
--- a/js/tapTarget.js
+++ b/js/tapTarget.js
@@ -59,6 +59,10 @@
           closeTapTarget();
           $(document).off('click.tapTarget');
         });
+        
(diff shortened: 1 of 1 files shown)


DanielRuf:
> Issue was found here
> #4487 (comment)
> 
> And fixed/test here
> http://codepen.io/anon/pen/oZVrEQ

Approved, the extra issue regarding the viewport change seems to be fixed, could not reproduce it anymore. Great work 👍

acburst:
Thanks!

PR #4489, opened 2017-04-05
Correct event names (chip.* not chips.*)

aularon:


tomscholz:
Ref.: https://github.com/Dogfalo/materialize/blob/master/js/chips.js#L280

Issue #4492, opened 2017-04-05
Variable $button-font-size only applies to icons inside .btn


Gorzas:
### Description
If you override on _variables.scss the value of $button-font-size it only applies to the icons inside a button but it doesn't applies the change on the text inside.

### Repro Steps
You can see an example on this [Codepen](http://codepen.io/Gorzas/pen/wJZpPd). The code on the SCSS it's a representation of the file [_buttons.scss](https://github.com/Dogfalo/materialize/blob/f4693a9917df7c40e68f380142110f5c4952cef2/sass/components/_buttons.scss#L42-L53) of the library.

### Screenshots / Codepen
An example of the view when you change the value of $button-font-size to .5rem. The icon gets smaller but the text not.
![screenshot from 2017-04-05 12-53-53](https://cloud.githubusercontent.com/assets/1456562/24702456/16765c06-19ff-11e7-8816-fae994c1f507.png)

DanielRuf:
Should be probably

```CSS
.btn,
.btn-floating,
.btn-large,
.btn-flat {

  outline: 0;
  font-size: $button-font-size;
  line-height: inherit;
  }
}
```
or similar.

DanielRuf:
But it seems this was like this for a very long time, see https://github.com/Dogfalo/materialize/commit/add7cffc6b0ec657a7e8a831f1676a7dcaf56708

DanielRuf:
I am not sure but I think the variable name `button-font-size` is misleading, wrong and the font size is probably inherited from parent elements.

Gorzas:
That's what I thought at first but it feels uncomfortable to have the icons inside a button with a different font-size than the text.

Gorzas:
I think that the best solution for this is the following:
- change the current variable name from `$button-font-size` to ` $button-icon-font-size`.
- add a new variable called `$button-font-size` with a value of 1rem.

What do you think?

acburst:
Merged

PR #4494, opened 2017-04-05
Fixed missleading with variable $button-font-size

Gorzas:
See #4492
There are two changes:
- It changes the name of the variable $button-font-size that is missleading
to a better name like $button-icon-font-size.
- It adds a new variable called $button-font-size with value of 1rem.


acburst:
thanks!

PR #3197, opened 2016-05-23
Added side-nav events (opened, opening, closed, closing).

BenDol:
Need a way to accurately hook to closing, closed, opening & opened actions.

Let me know if there is a more elegant way to do this or if anything is required to change (not sure about the code conventions for this).

Thanks!

PR #4497, opened 2017-04-05
Adding new onClosed and onOpened event to collapsibles

MYDIH:
Adding some events to being able to track collapsible end size (ie: opened or closed size). Personally, I'm using those events to trigger an update of my javascript scrollbars, but there should be some other use cases though. 

Thanks to take the time to consider this PR :)

[EDIT] :

Basically, I added 2 new callbacks called onOpened and onClosed. onOpened is triggered when the opening of the collapsible is completed, after the jquery slideDown ends. On the other hand, onClosed is called after the collapsible is closed (when slideUp ends). This means there is now an onOpen callback called when a user asks for the opening of a collapsible, an onOpened callback called when the opening is finished (all transitions terminated), an onClose callback called when the user asks for a collapsible close, and a an onClosed callback called after the collapsible closing procedure is finished.

* collapsible closed state -> opening asked (by click or programatically) -> callback onOpen is fired -> collapsible is opening and animates it's collapsible body -> collapsible is fully opened -> callback onOpened is fired

* collapsible opened state -> closed asked (by click or programatically) -> callback onClose is fired -> collapsible is closing and animates it's collapsible body -> collapsible is fully closed -> callback onclosed is fired

This way, you can react to collapsible height changes far more easily. In the case I mentionned above it's a real issue, because in my DOM there is only the collapsible elements that are causing document height changes and because I'm using javascripted scrollbars, being able to track those changes enables me to update my scrollbars in an event driven maner instead of getting a perfromance cost by using a timeout.

Hope it's clear,


acburst:
Can you explain more what these are and how these differ from the existing callbacks?

MYDIH:
I edited my first post,

This PR is similar to this one : [https://github.com/Dogfalo/materialize/pull/3197](#3197) but in the context of a collapsible instead of a side-nav.

Thank you for your time

acburst:
It seems you have some merge issues. I would suggest cleaning up your commits and reopening.

Issue #4508, opened 2017-04-07
In Autocomplete, the first result should be focused by default


praveenkumar-outlook:
### Description
In autocomplete, during suggestion of results, the result values are not focused initially. In that case, if enter is done, neither input value nor suggested values are entered. Hence, during suggestion, the first result should be focused by default.

### Repro Steps
1. Type 'a' in the input filed, 'Apple' will be suggested by autocomplete.
2. Press "Enter". Neither input nor result value will be entered.

### CodePen
http://codepen.io/mrpravin94/pen/jBooyK

### Screenshots
![autocomplete](https://cloud.githubusercontent.com/assets/26969693/24786295/026e0594-1b7e-11e7-9b68-c7b37b350a9b.jpg)

acburst:
This is how it used to work, but after some issues and some examples from other frameworks, we decided to do it this way as it was the most consistent with existing autocompletes.

PR #4509, opened 2017-04-07
Feature : Autocomplete will have first result value as autofocused

praveenkumar-outlook:
- See #4508


acburst:
I'm sorry but this is how we decided the autocomplete should behave.

PR #4527, opened 2017-04-12
Adding how to get the value for Date picker

GregBorrelly:
Some confusion on how to do this online.


Dogfalo:
I think this is pretty standard as it is no different from any other input.

PR #4528, opened 2017-04-12
Added variables for footer copyright color

hudadiaz:
Added `$footer-copyright-color` and `$footer-copyright-bg-color` to change footer copyright color.


tomscholz:
Wouldn't it make sense to be able to change the text-color of the footer via a variable? @fega Your thoughts?

fega:
I believe that it's okay in the actual implementation, it's something that could be changed with a 1 or two lines of normal css

PR #4551, opened 2017-04-18
Update _buttons.scss

medinskiy:
If you do not specify a width, it will be dynamic


acburst:
Please put more explanation to what this is fixing. Do you have an issue reference? if not, some reproducible code preferably in a codepen or something.

Issue #4040, opened 2016-12-29
Feature Request : Add SSL to site


vishnu1991:
### Description
Add SSL(https) Support to materializecss.com.
You can use openssl [Lets Encrypt](https://letsencrypt.org/) (Its Open Source).

tomscholz:
Materialize is using github hosting in combination with a (I guess) DNS provider. Github hosting offers https but due to the custom domain the certificate is not valid. Maybe Materialize could switch to cloudflare, where a ssl certificate is free and included?

Dogfalo:
There isn't any sensitive information that should be sent from either side on our website, so I don't think this is all that necessary.

tomscholz:
@Dogfalo @acburst Please, look into it https://github.com/Dogfalo/materialize/pull/4278

DanielRuf:
https should be used by default. It works with Cloudflare and GitHub, see https://websectools.com and https://github.com/DanielRuf/WebSecTools

Modern browsers are pushing more and more to a https-only internet and many apps and search results often redirect to the https version which often causes issues if there is no valid TLS certificate ;-)

tomscholz:
> There isn't any sensitive information that should be sent from either side on our website, so I don't think this is all that necessary.

To quote from an article of at [Googles Web Fundamentals](https://developers.google.com/web/fundamentals/security/encrypt-in-transit/why-https)
> You should always protect all of your websites with HTTPS, even if they don’t handle sensitive communications.

Maybe (in the future) we should consider to move to netlify
See: https://macwright.org/2017/05/08/https.html & https://netlify.com

fega:
I agree with @tomscholz this isn't a priority right now.

PR #4554, opened 2017-04-19
Url changes

tomscholz:
* New material design icons url
* Changed materializecss.com/mobile.html to mobile.html
* Changed some urls to https instead of http. This was also a problem when using sample codes in codepen. [Codepen is going HTTPS](https://blog.codepen.io/2017/03/31/codepen-going-https/). In the future requests that come in over HTTP will be redirected to HTTPS. That means all the images that are required over http, will show a warning. That's annoying. This fixes it.

Also we are getting just a little step closer to https://github.com/Dogfalo/materialize/issues/4040

### DIFF
--- a/jade/_footer.html
+++ b/jade/_footer.html
@@ -20,7 +20,7 @@
           </div>
           <div class="col l4 s12" style="overflow: hidden;">
             <h5 class="white-text">Connect</h5>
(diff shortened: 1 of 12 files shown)


tomscholz:
Mistake, sorry

tomscholz:
Thanks!

PR #4556, opened 2017-04-19
Move Color Utility Classes to global

sirlancelot:
I'm using the Sass version of Materialize in my own project. The way I've set it up is I have a `materialize-vars.scss` file which roughly looks like this:

```scss
@import "~materialize-css/sass/components/_color.scss";
// ==========================================================================
// Materialize variables
// ==========================================================================
// ...
```

The idea behind this file is that I can include it in multiple files which output separate compiled css files. At runtime, the separated css files are included on a single page.

It was unexpected to me that including `_color.scss` outputs classes. The result of that behavior is that multiple files would have these utility color classes.

Moving those utility classes in to the Utility section of `_global.scss` fixes the issue since that file is included further down in my pipeline.

I can create a basic repo of my usage. The general idea I had is that `@import`ing `_variables.scss` should be free in that it doesn't output any selectors by default. However, it has a dependency to `_colors.scss` which outputs selectors.


tomscholz:
I am not sure, if these are related :) Can you check?
https://github.com/Dogfalo/materialize/pull/3760
https://github.com/Dogfalo/materialize/pull/4088

sirlancelot:
Related to #3760. NOT related to #4088 though.

I kind of like the implementation in #3760. Only reason I didn't go with that in my own is that I wanted to avoid creating a new file. It appears to be a better solution overall though and would allow users to avoid including utility color classes entirely, a plus in my book.

sirlancelot:
I updated this PR to be a little closer to #3760 but with the key difference being that with this implementation, it is now only required to single "Variables" file which defines all variables while not outputting any selectors on its own.

I kept the color utility classes inside `_global.scss` because it already contains other "utility" selectors within it. Those can moved out in another PR if anyone wishes to do so.

Thanks for creating this @Dogfalo, I'm finding it very useful! I think the community will appreciate the change proposed here.

sass/materialize.scss
```diff
@@ -1,17 +1,16 @@
...
+// 1. Variables
 @import "components/variables";
 
-// Reset
+// 2. Reset
 @import "components/normalize";
 
-// components
+// 3. Mixins
+@import "components/prefixer";
```
tomscholz: The _prefixer.scss was commented out for a reason... :) Honestly I forgot why, but there was a reason!

sirlancelot: Haha I thought you might catch that! If you take a look that file though, you'll see that nearly everything inside it is commented out except for a single mixin which isn't used anywhere. I figured it would be harmless to uncomment it for now so save a future headache where someone puts things in that file and wonders why they're not working ;)

tomscholz: Actually we use autoprefixer for prefixing. I don't see the point in including it. @DanielRuf @Dogfalo

DanielRuf: @tomscholz is right https://github.com/Dogfalo/materialize/blob/e99b963a1e3bc8b50476cbc48de57d73194fdf4f/Gruntfile.js#L77-L103

sirlancelot: Understood, I left the line commented out in the latest version.

tomscholz:
@DanielRuf Your thoughts?

DanielRuf:
The mixins Sass file contains no actual code and should be removed.

Not sure if we should really rename the colors file to palette. This would also be a small breaking change in existing projects which upgrade to a new version with this change in the future.

The docs have to be checked too if we have references to colors which should be changed.

DanielRuf:
Nevertheless, good PR and makes much sense.

Pierstoval:
I didn't see that `colors` was renamed to `palette` (damn), I think this is not really useful. Keeping `colors` is nice.

Issue #4266, opened 2017-02-19
Memory Leak: `material_select` never removes global `click` handlers


jvilk:
### Description

`$.material_select` installs multiple *global* `click` handlers on `window`, but never cleans them up:

https://github.com/Dogfalo/materialize/blob/28af85de90bad5e7347d7882116fdfca098659f2/js/forms.js#L641

This results in a memory leak in single-page applications that call this function when rendering a view. Every time the function is called, new global click handlers are added. Since they are never removed, the browser cannot garbage collect the handler or any data within its closure (which includes DOM `NodeList`s within jQuery objects).

Is there a way to either avoid a global click handler, to only install a single global click handler, or to clean up global click handlers at some point?

### Repro Steps

N/A

### Screenshots / Codepen

N/A

jvilk:
Note: I discovered this issue while debugging memory leaks in [Piwik](https://piwik.org/), which uses this library.

r3wt:
@jvilk 

SPA's are beyond the scope of jQuery plugins. one approach you might take is to override jQuery's `init()` method and use it to maintain a list of selectors with bound events, and use the well known hack to remove all events.

override `$.fn.init`:
```js
var watchedSelectors = [];
$.fn.init = function(selector,context)
{
    //you could maybe add some logic to see if you want to watch it or not.
    var jQinstance = new jQselector(selector,context||window.document,$);
    
    watchedSelectors.push(selector);

    return jQinstance;
};
```

remove events when needed:
```js
for(var i=0;i<watchedSelectors.length;i++){
        $(watchedSelectors[i]).find("*").addBack().off();
}
watchedSelectors.length = 0;
```

you could probably take it a bit further and remove from specific selectors, but i'll leave that exercise up to you.

jvilk:
Thanks for the info. I'm not immediately familiar with expected jQuery plugin use cases, but if you anticipate web applications calling `material_select` a variable amount of times in any scenario, then this is still a valid issue.

A hackfix for a SPA using materialize would be to call `$(window).off('click')` when the "page" redraws, assuming no other part of the page registers a click handler on `window`.

A possible solution on your end is to expose a cleanup method in some manner, letting folks who are affected by this issue call the method to remove the handlers (`material_select_cleanup`?). I do not know if there is a convenient / expected way for jQuery plugins to expose this functionality.

[...]

PR #4566, opened 2017-04-21
Fix select memory leak (remove window event listeners)

UncleWalt:
This adds removal of the window event listeners for the select component and is a fix for #4266.

There may still be further steps necessary to properly cleanup all event listeners for the select component (i.e. select calls .dropdown() internally).


Dogfalo:
Made some more in-depth changes here: b6fb319e5265721a5c0ffe0e2674588169f2b3f0

PR #4603, opened 2017-04-25
Collapsible: Do nothing when trying to open an open panel

WiktorStarczewski:
When programatically firing 'open' on an already open collapsible, it is being closed instead. This patch keeps it open.


acburst:
In collapsible rewrite, this is now the case with the open and close methods

PR #4607, opened 2017-04-26
fix input[type=search] padding-left not in navbar 

tompipen:


tomscholz:
Please be more specific, what changes you made

tompipen:
I changed the .input-field input [type = search] padding-left to 0.
Added nav .nav-wrapper .input-field input [type = search] padding-left to 4rem.

tomscholz:
But why? What is the result. What are you fixing or adding?

tompipen:
Because i was not using the inside of the navbar inpupt [type = search] and found it padding-left = 4rem.

This will be very unexpected on the screen, I would like to input [type = search] should still be used in the place is not navbar, so I modified it.

But I found some effect I did not modify to, for example, input [type = search]: focus no bottom line.

So the issue I will first close, so I completely deal with the entire effect and then send pull request. thank you for your patience.

PR #4612, opened 2017-04-27
fix SELECT border for valid/invalid state

MaximBalaganskiy:
At the moment adding `valid`/`invalid` classes to the SELECT's input produces a greyish border.


MaximBalaganskiy:
The demo of the issue
http://codepen.io/anon/pen/ZKLRxa

acburst:
Thanks!

PR #4617, opened 2017-04-27
non-breaking text

Rendemedum:
This way non-breaking text in the title, content, and comments has no adverse effects on layout or functionality

### DIFF
--- a/sass/components/_typography.scss
+++ b/sass/components/_typography.scss
@@ -5,6 +5,7 @@ a {
 
 html{
   line-height: 1.5;
+  word-wrap: break-word;
 
   @media only screen and (min-width: 0) {
     font-size: 14px;


DanielRuf:
Why should the default `word-wrap: normal;` be changed?

Your change has unexpected sie effects.

Rendemedum:
non-breaking-text

Can you elucidate the unexpected side effects?
Would be nice to solve it within the Materialize Framework. Do you have another suggestion for it? 
I changed this in my stylesheet and it works.

DanielRuf:
> I changed this in my stylesheet and it works.

This should be the only solution imho.

Rendemedum:
That's a pity. I will dive into it. Keep up the good work.

Issue #4605, opened 2017-04-25
Request feature: Change from px to rem for Button


haojia321:
Please consider changing unit from px to rem for Button and any other components.
It makes UI looking better in cross device applications.

DanielRuf:
Definitely, especially retina and when users zoom in / out.

fega:
@tomscholz it looks more like an enhancement request

tomscholz:
@fega ;) You are right

haojia321:
Hi Guys,
Whats the default @root font-size in this framework? I think I will create a PR to change Button from px to rem.

DanielRuf:
@haojia321 depends

![image](https://cloud.githubusercontent.com/assets/827205/25486880/cb28f8f6-2b62-11e7-94bf-035753a5fdf1.png)

This is not how it should be. Normally just 16p / the default and rem.

DanielRuf:
http://materialdesignblog.com/material-design-typeface-in-css/

DanielRuf:
https://github.com/material-components/material-components-web/blob/master/packages/mdc-typography/_variables.scss#L89-L97

haojia321:
So I think what we need to do is just change component from px to rem. We let user to set root font-size on their own?

haojia321:
Hi Guys,
This is the PR for button size changes. I will change other components later but before I do that. I want to know if you guys are fine with this change or not

Thanks
Hao

Hoshinokoe:
It's very nice feature to have REMs instead of pixels, meaning you don't have to state an explicit height, width, padding, etc, for every device. Simply set font-size on root level and have the whole component (and its nested elements) proportionally changed. Right now, form elements (like checkboxes, radios, etc) have fixed width and height.

This is great for making site mobile friendly.

PR #4626, opened 2017-04-29
change button size from px to rem

haojia321:


DanielRuf:
Keep in mind that it is 14 (not 16) pixels according to the specs and the css
https://github.com/Dogfalo/materialize/issues/4605#issuecomment-297725100

also https://github.com/Dogfalo/materialize/issues/4605#issuecomment-297722124

haojia321:
@DanielRuf My calculation for converting px to rem is based on root font-size:16px; Are you saying it should based on root font-size:14px ?

DanielRuf:
Yes, because 14 is what we are currently using (and also 14.5 and 15, see the comments) and what is used by MDL and others.

haojia321:
I have changed to px-rem based on html font-size:15px. This is based on Large Desktop Devices size. So button size will look smaller in other screen sizes which is what we want.

DanielRuf:
Hm, it should be based on 14px as MDL and other use the same. Currently this is not the right solution as this is dependent from the current viewport size.

In this case you have to clone the media queries and set the rem value for each button class.

haojia321:
I dont think we should give different rem number to same component based on different screen size. The intention of REM is to give a unique number to different screen sizes. Eventually browser will render component in different size based on the root font size because rem is only relative to the html (root) font-size.

DanielRuf:
We have to as 14, 14.5 and 15px are used for html, see 2 days ago

We are running in circles here. https://github.com/Dogfalo/materialize/pull/4626#issuecomment-298182693
Please change the PR and add the media queries for this and use the right rem values for each.

Can you provide exact examples where we use rem and where the size changes accordingly? Because 1rem is 14, 14.5 or 15px depending on the viewport.

Because your change would change the calculated size in at least two viewports (the viewports with html font-size 14 and 14.5px). You have to check how the other values are calculated (if they were based on 14, 14.5 or 15 when these rules were created).

Currently the whole frameworks is not consistent as we should remove the 14.5 and 15 font-size rules and recalculate all values on 14px.

haojia321:
Who said we will "remove the 14.5 and 15 font-size rules and recalculate all values on 14px."? With all due respect that's not right. Give same component a different rem number for same component is really violate the intention of REM.

I understand we have  14, 14.5 and 15px for font-size in html node.
Lets say current setting is:
Small device: font-size=14px. Medium device: font-size=14.5px and Large device: font-size=15px  

So if I give button height=2.5rem; The actual button height will be: Small device button height=35px. Medium device button height=36.25px. Large device button height=37.5px. 
Button displays different size in different screen size is common and as expected. You dont want to see a button display in mobile as big as in desktop. 
See this file https://github.com/Dogfalo/materialize/blob/master/sass/components/_variables.scss  If you check Collapsible or Input component in here, they are doing the same(only one rem).

This change has been in our own application for long time. We have tested and we dont see any issue with it.

DanielRuf:
> Who said we will "remove the 14.5 and 15 font-size rules and recalculate all values on 14px."? With all due respect that's not right.

We should in the future. Just 14 px is right and used by Google (MDL and so on) and this is right according to the specs ;-) Just my opinion that Materialize is not very consistent and right in some places / notfollowing the specs.

DanielRuf:
> So if I give button height=2.5rem; The actual button height will be: Small device button height=35px. Medium device button height=36.25px. Large device button height=37.5px.
> Button displays different size in different screen size is common and as expected. You dont want to see a button display in mobile as big as in desktop.

Well but this is a change from the fixed and same height from before ;-)

haojia321:
> We should in the future. Just 14 px is right and used by Google (MDL and so on) and this is right according to the specs ;-) Just my opinion that Materialize is not very consistent and right in some places / notfollowing the specs.

Right, MDL is more comprehensive. It involves sp/dp/dpi etc. It really more focus on mobile phone in my opinion. We need to change a lot in order to align with MDL. That could come with a major release change. Enhancement should be small.

haojia321:
> Well but this is a change from the fixed and same height from before so it might introduce small changes in the UI and the size of buttons ;-)


Agree there will be small look and feel changes in UI. And will there be code changes needed? there will be if user has customized the UI around the button. But most likely there should be no issue.

haojia321:
Thanks @DanielRuf . I really appreciate it your suggestions and comments. Im looking forward to work with you guys together to make this framework better.

haojia321:
@tomscholz Any plan about this?

PR #4634, opened 2017-05-02
Updated dependencies

tomscholz:
* Removed the _SpecRunner.html file since it’s automatically generated when running the jasmine tests.
* I also talked to Jason, who owns the grunt-rename-util(prev. grunt-contrib-rename), to update the package to be compatible with grunt 1.0.0
* Updated the node engine, because since 2016-12-31
 the support has ended


DanielRuf:
I guess we need another setTimeout + done() callback there? https://github.com/Dogfalo/materialize/blob/master/tests/spec/tabs/tabsSpec.js#L76

tomscholz:
@DanielRuf Oh man. I just saw this now. 

Like this?
```
    it("should horizontally scroll when too many tabs", function () {
      var tabsScrollWidth = 0;
      normalTabs.parent().css('width', '400px');
      normalTabs.find('.tab').each(function() {
        tabsScrollWidth += $(this).width();
      });

      setTimeout(function() {
        expect(tabsScrollWidth).toBeGreaterThan(normalTabs.width(), 'Scroll width should exceed tabs width');
        done();
      }, 400);
    });

  });
});
```

DanielRuf:
Try this

```
it("should horizontally scroll when too many tabs", function (done) {
      var tabsScrollWidth = 0;
      normalTabs.parent().css('width', '400px');
      normalTabs.find('.tab').each(function() {
        tabsScrollWidth += $(this).width();
      });

      setTimeout(function() {
        expect(tabsScrollWidth).toBeGreaterThan(normalTabs.width(), 'Scroll width should exceed tabs width');
        done();
      }, 400);
    });

  });
});
```

tomscholz:
🤔 Well... that doesn't work. @acburst you wrote that test. I don't get why "Scroll width should exceed tabs width" is a thing. Tabs have the same width when they are scrollable or not.

DanielRuf:
There beeds to be a timeout until the styles are applied. Did you test locally using Grunt? The test should not fail with the aforementioned fix. At least this workaround helped to fix other tests.

DanielRuf:
`grunt travis --filter=tabs` should help with local tests. It might be that another timeout is needed around the

```
normalTabs.find('.tab').each(function() {
        tabsScrollWidth += $(this).width();
      });
```

after the `.css()` code. I will test this here with your branch in about an hour.

tomscholz:
@DanielRuf I know. I was running the test locally, before I pushed them!

DanielRuf:
Ok. So on your system the test in the branch did not fail? Will test it here on Windows. Might be that the PhantomJS driver and the headless webkit on Travis under Linux is a bit slower (at least the free plan has less power than the paid ones)?

tomscholz:
@DanielRuf Well, I may sound like an idiot now, but no, the test did fail on my local machine.

DanielRuf:
Then the change did nothing as it should havr fixed the failing test ;-) I will check this for you.

tomscholz:
@DanielRuf Looks good to me. What do you think?

DanielRuf:
Not sure why we need the changes in the package file as some tools still use them.
So far the other things look good to me.

http://stackoverflow.com/questions/32037150/style-field-in-package-json
https://www.npmjs.com/package/npm-sass

DanielRuf:
It should be easy to revert the last commit as this one would break many implementations / projects ;-)

DanielRuf:
Looks good =)

PR #4642, opened 2017-05-03
feature/load-config-grunt

### LINKED ISSUE #4641: Separate modules / scripts from Grunt jobs
The problem is the management of modules and support of _Gruntfile_ files, which end up generating conflicts over time when added new features to it.

For this it would be interesting to use a manager to dynamically load the configuration scripts of the grunt modules, leaving in _Gruntfile.js_ only the specific tasks. With this would give to have:

```
+- Gruntfile.js
+- grunt-tasks/
|    +- clean.js
|    +- connect.js 
|    +- sass.js
|    +- ....
```
 
When there was change in any of these configurations, would be isolated.
An NPM module that does this is: ```load-grunt-config```.  What do you think?

gitarno:
This pull request for close #4641 , where I separated the grunt settings into JS files


DanielRuf:
@gitarno there is a trailing comma missing in package.json at line 43.

tomscholz:
@Dogfalo please close

PR #4649, opened 2017-05-05
Allow carousel reinitialization

aschoenebeck:
Previously it was not possible to reinitialize a carousel if items have changed by adding/removing .carousel-item nodes.

This patch allows reinitialization by calling .carousel() again. All events and classes are first removed and then the normal setup process continues.

I have also included in this patch that if a .carousel-item has the .active class, the carousel jumps to this item after (re-)initialization. The .active class can either be manually set or is set by the carousel itself when the user is scrolling to it.

This patch was necessary, because I pull updated items from a server and dynamically add/remove carousel items using Knockout.js. Now I only need to call $('#myCarousel').carousel() again.

I have tested this with the default carousel options and with { fullWidth: true }.

### DIFF
--- a/js/carousel.js
+++ b/js/carousel.js
@@ -21,7 +21,7 @@
         var uniqueNamespace = namespace+i;
         var images, item_width, item_height, offset, center, pressed, dim, count,
             reference, referenceY, amplitude, target, velocity, scrolling,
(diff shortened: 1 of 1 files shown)


acburst:
A destroy method was added to carousel recently. You can use this to tear down and reinit

Issue #2913, opened 2016-03-15
Select is broken


Writtscher:
Hi,

im using the select component but I have to say that the component is broken - at least the one that contains optgroups.

Using optgroup together with "multiple" does not render any checkboxes neither does it select the correct options.. The label contains totaly different options than I've selected.

Another thing is: I would like to have the option to choose whether the option should scroll to the top as soon as it is selected. It's quite annoying.

haidaraM:
Do you init the the select component in your jquery document ready ?

```javascript
// Plugin initialization
$('select').not('.disabled').material_select();
```

Writtscher:
Sure. Does not change anything.

Btw: using current version.

vincentcox:
Can you look in the "console" for errors? 
To access console you can right click on your page, select "inspect". Now go to the tab "Console" and reload the page. Look for errors there and post them here so we can take a look.

Writtscher:
I know how to access the console - no offence. There is no error. Nohting.

Is it working for you? I mean selects with optgroups and checkboxes? 

Another point is maybe that I'm using browserify to resolve dependencies. Weird. I'll try it without. Maybe this helps.

Writtscher:
Not working. Even without browserify: http://codepen.io/anon/pen/XdRyEe

erdostom:
👍 On this, spent hours trying to figure out what's wrong with my code, but it's the plugin thats' broken. It handles regular multiple selects fine, but can't deal with optgroups.

- no checkboxes
- submits incorrect selection
- displays incorrect selection when fed list of current selections

TyrionGraphiste:
Hi, long time ago I rewrote the entirely component, taking account the select multiple with optgroups.
You can get it at this address : https://github.com/TyrionGraphiste/materialize/blob/master/js/forms.js

There is an example : https://jsfiddle.net/TyrionGraphiste/5deezjj0/1/

Birssan:
Hi,

Same issue here, when using select with `multiple` attribut and `optgroup`, checkbox aren't rendered and default selected values are not correct... Only got 3 "checked" out of 19 and yet in the markup the 19 options are selected, there just and issue with the generated list

storm1er:
Just found this issue on 0.98.0
No PR for this ? Since 2016 ? Arf ... i'm on it :/

aschoenebeck:
[...]

PR #4656, opened 2017-05-06
Fix select with optgroups

aschoenebeck:
This fixes setting checked on select options within optgroups.

The problem with the original each() is, that the index is relative to the containing optgroup. By iterating over every option, checking for selected and taking the index within all options the corresponding li element of the generated optgroup-options can be properly addressed.

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -661,14 +661,14 @@
 
       // Add initial multiple selections.
       if (multiple) {
-        $select.find("option:selected:not(:disabled)").each(function () {
-          var index = $(this).index();
-
-          toggleEntryFromArray(valuesSelected, index, $select);
-          options.find("li").eq(index).find(":checkbox").prop("checked", true);
+          $select.find("option").each(function (index) {
+              if ($(this).is('option:selected:not(:disabled)')) {
+                  toggleEntryFromArray(valuesSelected, index, $select);
+                  options.find("li:not(.optgroup)").eq(index).find(":checkbox").prop("checked", true);
+              }
         });
       }
-
+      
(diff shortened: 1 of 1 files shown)


aschoenebeck:
Fixes issue https://github.com/Dogfalo/materialize/issues/2913

tomscholz:
Does this fix https://github.com/Dogfalo/materialize/issues/2274 aswell?

acburst:
Hmm the multiple select with opt groups seems to work here on the latest version: http://materializecss.com/test/html/forms.html 

Can you provide an updated codepen reproducing this issue?

mortenson:
This fixed multiple errors my team was encountering with multiple select fields and optgroups.

I couldn't get a Codepen working at all for just rendering a simple select group (I'm probably doing something wrong, testing as https://codepen.io/mortenson/pen/JJgLpP), but you can see part of the problem at the http://materializecss.com/test/html/forms.html page posted above. If you look at the "Materialize Multiple Select with optgroups" field, "Lamborgini" is displayed as the selection but it shouldn't be as it's not checked by default. The field should be empty until something is selected.

kmmbvnr:
Just checked, The patch really solve the issue. I've added sample here -
 https://github.com/Dogfalo/materialize/issues/2913#issuecomment-318621604

```diff
   if ($(this).is('option:selected:not(:disabled)')) {
```

This line required to prevent `Choose you option` value been marked with checkbox and came into the textinput

mortenson:
@aschoenebeck Could you merge master into your pull request branch? Thanks!

PR #4666, opened 2017-05-09
Parallax.js: Responsive Image Sources

florian-senn:
After these changes, one could use `<picture>,` as well as `` tags to define a parallax picture source.

Especially `<picture>` with rules for landscape and portrait are very useful, regarding art direction and in combination with imgix and lazysizes.


Dogfalo:
Fixed in v1 dev here: 15c729caa694f256054ce5a857df70db55bbba22

PR #4675, opened 2017-05-10
Created three events for input fields that are fired after validation.

jmahc:
Created two events that are fired after input field validation/invalidation.

- 'materialize:valid-input' - event is fired when the input field receives the 'valid' CSS class.
- 'materialize:invalid-input' - event is fired when the input field receives the 'invalid' CSS class.
- 'materialize:reset-input' - event is fired when the input field has the 'valid' or 'invalid' CSS classes removed.

These events can be handled with the jQuery `on` method.

Valid:
```
$('#inputField').on('materialize:valid-input', function() {
  // handle a valid input field
});
```

Invalid:
```
$('#inputField').on('materialize:invalid-input', function() {
  // handle an invalid input field
});
```

Reset:
```
$('#inputField').on('materialize:reset-input', function() {
  // handle a reset of the input field
});
```

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -93,11 +93,11 @@
             object.addClass('valid');
           }
           else {
-            object.removeClass('valid');
-            object.addClass('invalid');
             if (!object.hasClass('invalid')) {
               object.trigger($.Event('materialize:invalid', {}));
             }
+            object.removeClass('valid');
+            object.addClass('invalid');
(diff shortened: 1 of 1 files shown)


Dogfalo:
Don't think these are that useful as you can do something very similar add a keyup handler and check the input itself or the class we've added

PR #4683, opened 2017-05-14
Added flat navigation bar option

penguoir:
By adding the class `.flat` to a navigation bar, the navigation bar looses its box shadow and becomes "flat".

### DIFF
--- a/jade/page-contents/navbar_content.html
+++ b/jade/page-contents/navbar_content.html
@@ -122,6 +122,74 @@
         </code></pre>
       </div>
 
+      <div id="flat" class="section scrollspy">
+
+        <h2 class="header">Flat</h2>
+        <p>
+          Add flat class to nav tag to make nav not stand out.
+        </p>
+
+        <nav class="flat">
+          <div class="nav-wrapper">
+            <div class="col s12">
+              <a href="#!" class="brand-logo">Logo</a>
+              <ul class="right hide-on-med-and-down">
+                <li><a href="sass.html">Sass</a></li>
+                <li><a href="badges.html">Components</a></li>
+                <li><a href="collapsible.html">JavaScript</a></li>
+              </ul>
+            </div>
+          </div>
+        </nav><br>
+
+        <pre><code class="language-markup">
+  &lt;nav class="flat">
+    &lt;div class="nav-wrapper">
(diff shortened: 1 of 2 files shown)


DanielRuf:
What are good reasons for this? Normally a navbar is elevated and not on the same level as the content.

I will check the material design specs if this makes sense and check the other components.

Also please add this change to the docs in the jade directory so this would be mentioned in the docs if it would be merged.

tomscholz:
Spec: https://material.io/guidelines/layout/structure.html#structure-app-bar

penguoir:
This feature can be used when using a transparent navigation bar. Nobody wants a shadow from a transparent element.

tomscholz:
I would change it to transparent. Flat is misleading since there is no official spec for flat.

penguoir:
Sorry, I don't understand. Change the name to "transparent", remove shadow from the transparent navigation bar or something else?

DanielRuf:
So soecifically it is not flat but transparent according to the specs?

Should we use another class for this or `.flat` to be more precise?

penguoir:
Perhaps a `.flat` class which just removes shadow and a `.transparent` class which turns the navigation transparent as well as removing shadow?

fega:
What about z-depth-0?

tomscholz:
@fega Well, you are a genius

DanielRuf:
The current change is just the z depth class http://materializecss.com/shadow.html
https://github.com/Dogfalo/materialize/pull/4683/files

We should use transparent and set the background color to transparent (maybe).

DanielRuf:
And the background?

penguoir:
`.transparent` already has a selector in colours.scss, no need to repeat the same selector.

DanielRuf:
> .transparent already has a selector in colours.scss, no need to repeat the same selector.

Ah, right. Forgot to check that there :D

acburst:
I don't see the point of a flat class if you can add a z-depth-0 class?

DanielRuf:
> I don't see the point of a flat class if you can add a z-depth-0 class?

Right. And combine this with the transparent class.

I guess it is better to just add an example with these classes to the docs.

tomscholz:
We are currently not adding this into our core framework. It is easy enough to customize yourself.

Issue #4635, opened 2017-05-02
Fixed navbar jumps while showing sidebar, modals, date picker


ManuZiD:
### Description
The content of a fixed navbar moves to the right in browsers on Windows (Firefox, Chrome) when the browser scroll bar is hidden due to sliding in sidebar, modal dialogs or date picker. 

### Repro Steps
1. Use Firefox or Chrome on a Windows desktop
2. Go to https://codepen.io/anon/pen/GmEmgQ
3. Click on "Menu"

### Additional Information
This does not happen if the navbar is not fixed (i.e. no position:fixed).

DanielRuf:
Looks right as the scrollbar is hidden and so the available width is greater (the 70% width)

ManuZiD:
For me as a user it feels buggy because e.g. the sidebar is an overlay and as such not pushing content away. So I expect virtually no jumps of unaffected content (like this: https://contributor.google.com/v/signupdisabled).

DanielRuf:
Well, take a closer look where the scrollbar starts ;-)

![image](https://cloud.githubusercontent.com/assets/827205/25618953/506f8964-2f49-11e7-86de-421b6ae122ef.png)

ManuZiD:
That's true for the header but the body content here doesn't jump because the scroll bar isn't hiding. Could this be a solution? To not hide the scroll bar (maybe optionally) if the side bar or modals are displayed?

huazai-688:
can you speaks chinese？

ManuZiD:
here is a possible solution [4686](https://github.com/Dogfalo/materialize/pull/4686)

Nikitozz13:
Have issue when fixed navbar jumps under the status bar when I open camera to take a picture or choose it from photo library on iOS.
![img_8054](https://user-images.githubusercontent.com/2047718/30061864-c1fb1b46-9251-11e7-89fe-a0c85742d2d7.PNG)
![img_8055](https://user-images.githubusercontent.com/2047718/30061868-c4dc9664-9251-11e7-9e7b-bae56ea4e4c5.PNG)
![img_8056](https://user-images.githubusercontent.com/2047718/30061873-c75d8d12-9251-11e7-8894-67eff8ff1d74.PNG)

Dogfalo:
Added an option in sidenav and modal in a9a87f4f9cec5e7412a097a132c64cb8ff671663

PR #4686, opened 2017-05-15
Feature/fixed navbar jumps

ManuZiD:
Reference to the following issue: [4635](https://github.com/Dogfalo/materialize/issues/4635)

Now it's possible to prevent this issue with the option `hideScrollBar`. The default value of this option is true. To prevent these jumps you have to set the value to false.

Default behaviour: https://codepen.io/anon/pen/GmEmgQ

Behaviour without jumps: https://codepen.io/anon/pen/XRYZMp


tomscholz:
Are you running on windows?

ManuZiD:
On work. Why do you ask?

tomscholz:
Cuz I wasn't able to reproduce it on a mac in chrome,safari or firefox. Was just confused

ManuZiD:
Yes, that's a windows exclusive bug :)

Dogfalo:
Added an option in sidenav and modal in a9a87f4f9cec5e7412a097a132c64cb8ff671663

Issue #4716, opened 2017-05-22
New List Feature


penguoir:
Hi, everyone!

I'm working on a [new feature](https://github.com/penguoir/materialize-list) and I'm almost done. Wondering if I should finish off (organise sass and such) and submit a pull request or if I've missed anything.

Thanks for the help!

tomscholz:
@DanielRuf this looks good to me on mobild, but can you have a look at this and give some feedback too?

penguoir:
I just realised that collections is really similar to this.

DanielRuf:
@penguoir but lists as you implemented are still missing, see https://github.com/Dogfalo/materialize/issues/4553

tomscholz:
@penguoirI think collections are a little bit out of scope from the material design spec. Maybe you could rewrite the collections with your lists. I really think this could fit into this framework

penguoir:
Cool. Is there a specific way to add a component to the framework?

DanielRuf:
It's recommended to fork Msterialize, create a new feature branch, add the files in sass/ and js/ and docs in /jade, add the files to the Gruntfile, test locally and push it to GitHub and open a new pull request for this.

penguoir:
I've made the [pull request](https://github.com/Dogfalo/materialize/pull/4727)

penguoir:
I think it would be a good idea to integrate lists as I implemented them and slowly deprecate the collections component.

djensen47:
> I think it would be a good idea to integrate lists as I implemented them and slowly deprecate the collections component.

Agreed. It's a little confusing that it is currently called collections when the Material Design spec refers to lists. It's also a good idea to deprecate so that existing sites or libraries that depend on Materialize have time to adjust.

tomscholz:
The discussion has moves to the pr

djensen47:
Um, but the issue is not resolved. There isn't a way to lock a conversation without closing?

tomscholz:
;)

tomscholz:
Changed my mind. Closing this one and keeping [this one ](https://github.com/Dogfalo/materialize/issues/2837)open.

PR #4727, opened 2017-05-26
Added list component to docs

penguoir:
The list component lets users show a number of similar items in an easy-to-read way.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -358,6 +358,7 @@ module.exports = function(grunt) {
           "forms.html": "jade/forms.jade",
           "buttons.html": "jade/buttons.jade",
           "navbar.html": "jade/navbar.jade",
+          "lists.html": "jade/lists.jade",
           "cards.html": "jade/cards.jade",
           "preloader.html": "jade/preloader.jade",
           "collections.html": "jade/collections.jade",
--- a/jade/_navbar.jade
+++ b/jade/_navbar.jade
@@ -71,6 +71,8 @@ header
                 a(href='forms.html') Forms
               li(class=(page == "Icons" ? "active" : ""))
                 a(href='icons.html') Icons
+                li(class=(page == "Lists" ? "active" : ""))
+                  a(href='lists.html') Lists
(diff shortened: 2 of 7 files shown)


penguoir:
See #4716 for more details

penguoir:
What do you guys think? Any improvements? Extra features?

ManuZiD:
looks great. I started to working on this too but you was quicker :)

👍

acburst:
So is this supposed to be a replacement to collections? If this is basically a rewrite it would make more sense instead of adding a whole new component to just restyle the collections component? This just feels weird for backwords compatibility

djensen47:
In the material design specification there is no "collection" component. There is, however, a list component and I think this pull request is great.

I would recommend accepting this PR as is and deprecate the collections component to be completely removed for version 1.0. Yes, there would be two similar components for a time but you would ensure backwards compatibility until the old component is removed.

djensen47:
@penguoir I reviewed the PR, I think it looks good (but I'm not a committer on this project). One thing, it looks like the documentation page `page-contents/lists_content.html` is missing?

djensen47:
So, I tried using this and ran into a problem. The `.primary` class clashes with a class name that I already have defined elsewhere. Actually, it appears to be used by aurelia-ui-toolkits/aurelia-materialize-bridge to set the color scheme.

Collections actually uses `.secondary-content` and I think that would be great here as well. Just prefix `-content` to `.primary` `.secondary` and `.tertiary`

ManuZiD:
We're not able to co-work on this, right?

I have a template of the list example site here.

djensen47:
Since a pull request is just a branch, you can branch a branch and submit it as a new PR.

tomscholz:
> Collections actually uses .secondary-content and I think that would be great here as well. Just prefix -content to .primary .secondary and .tertiary

@djensen47 Did you mean suffixing? `.primary-content` `.secondary-content` & `.tertiary-content`?

djensen47:
Yes. 😀

devansh1110:
The demo is no longer working fyi.

Issue #634, opened 2015-02-06
Bug when used with AMD (require.js)


max1011:
the company i am working with likes material design, so i told them about materialize, and they said ok use it , no problem , but currently i am stuck with this so please help !

when i try to require materialize , no errors are shown, but when i actually use it with jquery, it tells me that the method that i am trying to use on that element is undefined
here is my code (js):

``` javascript
requirejs.config({
  paths: {
    'materialize': '../../bower_components/materialize/dist/js/materialize',
    'jquery': '../../bower_components/jquery/dist/jquery'
  },
  shim: {
    'materialize': {
      deps: ['jquery']
    },
    'jquery': {
      exports: '$'
    }
  }
});

require(['jquery', 'materialize'], function($) {
  $('select').material_select();
});
```

Screen Shot: 
![image](https://cloud.githubusercontent.com/assets/6626011/6079567/92fbe448-ae0f-11e4-94e6-222736a11ee0.png)

Dogfalo:
Currently there are some problems with using materialize + requirejs. We will look into this when we can.

max1011:
i solved this on my own, there is an issue with the buttons, but other than that it works fine, i will post the fix soon

Dogfalo:
Great, looking forward to seeing it.

max1011:
@Dogfalo , I just fixed it all up it works 100% Love It, i want to ask you if i post this solution will i be one of the contributors on this project ??

Dogfalo:
Yes if you make a pull request and we accept it.

max1011:
OK I've got my code on a separate file, i will add it to the project itself and i'll make a pull request, so all what's left is to modify the source code, and hopefully you'll like it.

max1011:
To fix this here is what your file should look like, @Dogfalo please add this to the documentation so that if there is anyone that uses require knows what to do and you certainely don't want to lose them
```javascript
requirejs.config({
  paths: {
    'materialize': '../../bower_components/materialize/dist/js/materialize',
    'jquery': '../../bower_components/jquery/dist/jquery',
    'velocity': '../../bower_components/materialize/js/velocity.min',
    'hammerjs': '../../bower_components/materialize/js/hammer.min'
  },
  shim: {
    'materialize': {
      deps: ['jquery', 'hammerjs', 'velocity']
    }
  }
});


// Require both jquery and materialize (materialize will load other stuff that is needed)
require(['jquery', 'materialize'], function($) {

  $(function() {
[...]

PR #4751, opened 2017-06-03
 The hammer.js bug only happens if the value of eventType is 8.

tawon2137:
I am a developer studying the materialize css source. I noticed that during the creation of sidenav, Chrome experienced a drag error on newer browsers, and when the eventType value is 8, an odd value comes in. Share the source you have resolved as above.

### DIFF
--- a/js/sideNav.js
+++ b/js/sideNav.js
@@ -160,18 +160,13 @@
           $dragTarget.hammer({
             prevent_default: false
           }).bind('pan', function(e) {
-
-            if (e.gesture.pointerType == "touch") {
+            // Vertical scroll bugfix eventType Value === 8 is hammer.js bug
+            if (e.gesture.pointerType == "touch" && e.gesture.eventType !== 8) {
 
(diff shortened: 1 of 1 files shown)


tomscholz:
Which hammer.js bug do you mean?

https://github.com/Dogfalo/materialize/issues/634, https://github.com/Dogfalo/materialize/issues/1422

DanielRuf:
I think he generally means the removed line in the linked commit which has unexpected results and does not reliably filter the vertical scroll event.

tomscholz:
@tawon2137 Can you please provide some kind of demo/codepen when this issue appears?

tawon2137:
[Https://github.com/Dogfalo/materialize/issues/4082](Https://github.com/Dogfalo/materialize/issues/4082)
It was a pull request for the above issue,
Currently, the problem is solved, but I thought it would be better to check the value of eventType Value in hammer.js, rather than checking the coordinates of x and y.

PR #4753, opened 2017-06-03
'%' is redundant

Rendemedum:
Unit of measure


DanielRuf:
Beware, calc is not supported by older browsers, especially the integrates browser on Android, see http://caniuse.com/#feat=calc

So this is not really redundant but some sort of fallback.

Rendemedum:
Redundant was for the padding: 0% in _carousel.scss. I shall be more careful next time and make the pull request seperate. It was late :-)
But I understand why the width is double mentioned. Thanks for explaining.

PR #4775, opened 2017-06-06
Update _buttons.scss see FAB to Toolbar

lil5:
fix ul was not above .fab-backdrop
see FAB to Toolbar <fab-toolbar-demo.html>


tomscholz:
Can you post a codepen, that shows the problem in depth?

lil5:
If someone wanted to add anything else into the toolbar it would be invisible.
<https://codepen.io/lil5/pen/NgGweL>

PR #4778, opened 2017-06-07
SideNav: Run 'onClose' and 'onOpen' hooks when opened/closed via swipe

konradjurk:
Both hooks were only called in case the sidenav was opened via a button click and closed with a click on the backdrop.

Let me know if there is a better way to do this. This is the only one I could find on the first glance.


DanielRuf:
Looks good!

tomscholz:
Damnit wrong button ;)

PR #4782, opened 2017-06-08
Buttons: Change z-index of .fixed-action-btn from 998 to 997

konradjurk:
This way it is behind the sidenav backdrop and not in front of it. I noticed the issue when using an FAB.


DanielRuf:
I guess 996 would also work as the overlay has 997. Can you test this with a (fixed) navbar and a FAB?

Issue #2033, opened 2015-09-14
Extras folder is not included in the zips


acburst:

tomscholz:
Extras are still not a part of the zips!

fega:
OK!!! sorry! :open_mouth:

tomscholz:
@acburst Why do they even have to be a part of "the zips"? I mean people can easily download them or obtain them via a git cdn like rawgit

tomscholz:
https://github.com/Dogfalo/materialize/pull/4799

PR #4799, opened 2017-06-12
Added extras to zips

tomscholz:
Solves Dogfalo/materialize#2033


tomscholz:
@acburst Reopen if needed

Issue #4764, opened 2017-06-05
Provide browser-default for all components (or avoid global styling)


djensen47:
In order to avoid collisions with other CSS libraries, either provide browser-default for all elements affected or please avoid styling html components a class specific to Materialize is used. For instance in `sass/compoents/forms/_forms.scss`

```
select:focus {
  outline: $select-focus;
}

button:focus {
  outline: none;
  background-color: $button-background-focus;
}

label {
  font-size: $label-font-size;
  color: $input-border-color;
}
```

This literally makes it nearly impossible to use other CSS libraries. It actually looks like all of the forms components act this way.

tomscholz:
Duplicate: https://github.com/Dogfalo/materialize/issues/4661

djensen47:
It's not a duplicate. Using browser-default does not solve this problem/

djensen47:
Wow, on top of that browser-default only works for one component, select lists.

djensen47:
I updated the title to more accurately reflect the issue. 

It looks like browser-default only handles select and date-picker. It also does not reset the :hover behavior.

tomscholz:
https://github.com/Dogfalo/materialize/pull/4041

PR #4041, opened 2016-12-29
Display default inputs

developeranirudhprabhu:
Added pseudo classes  to the input selectors so that "browser-defaults" can be applied to other input fields (https://github.com/Dogfalo/materialize/issues/3975)

Issue #3975, opened 2016-12-06
Display default inputs


levani:
As I see only the select element supports the `browser-default` class to display the default select element. Is there a way to use this class on other form elements as well? I want to use the default text inputs and checkboxes in some parts of the app but can't find an 'official' way to do so...

developeranirudhprabhu:
There is official thing for this requirement at present. Will add this and commit. For now you can add ":not(.browser-default)" to all the input types you wish to target at line 5700 in materialize.css. Present selector listed are

input:not([type]),
input[type=text],
input[type=password],
input[type=email],
input[type=url],
input[type=time],
input[type=date],
input[type=datetime],
input[type=datetime-local],
input[type=tel],
input[type=number],
input[type=search],
textarea.materialize-textarea

developeranirudhprabhu:
@acburst Please merge the PR

developeranirudhprabhu:
@Dogfalo Please merge my PR so that we can resolve this ticket

mckenzieja:
Been struggling with the same issues. We need browser defaults for forms for increased flexibility.

PR #4802, opened 2017-06-12
Display browser-default for inputs

tomscholz:
Added pseudo classes  to the input selectors so that "browser-defaults" can be applied to other input fields. This is a replacement PR for #4041 

closes Dogfalo/materialize#4041
closes Dogfalo/materialize#3975
closes Dogfalo/materialize#4764

### DIFF
--- a/jade/page-contents/helpers_content.html
+++ b/jade/page-contents/helpers_content.html
@@ -166,6 +166,10 @@
               <td>SELECT</td>
               <td>Browser default select element</td>
             </tr>
+            <tr>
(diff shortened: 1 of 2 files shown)


nateq314:
What is the status on this? I'm using a version downloaded just 2+ weeks ago and just tried setting the class of a text input element to "browser-default", but not seeing any change?

samarjit:
Browser-default class still does not solve the problem of inter operating well with other libraries. A parent class approach will be much better.

Issue #4814, opened 2017-06-14
Side Navigator not fixed on small devices


alexisdiel:
On small devices, when i open the side navigation and open a collapse menu that overflows the height, I can make small move on X-axis, 

I found an inline-style on `ul.side-nav` element.
`overflow: auto;`

But not found when/where it's applied, i thing that's applied from some external JS.

I resolved the problem changing the inline-property from `overflow: auto` to `overflow-y: auto` 

![image](https://user-images.githubusercontent.com/7047656/27135904-258dc3b4-50f0-11e7-8575-51d7cfb8c68b.png)

DanielRuf:
The inline style is from the sidenav component https://github.com/Dogfalo/materialize/blob/master/js/sideNav.js

DanielRuf:
Seems we need a fix like https://github.com/Dogfalo/materialize/commit/d9d0028e55d5477d66761fe53231e8750c22b5ba.

DanielRuf:
I guess you mean overflow and not override.

alexisdiel:
See my pull request, i've fixed changing the css override on nav-mobile initialize

acburst:
Changing the overflow might just be a bandaid fix. If the layout is done correctly then you shouldn't be able to scroll horizontally. This might mean there is some css error that is causing some element to have extra width giving the sidenav horizontal scrollability.

I think its a good principle to try first to really diagnose the problem before proposing a fix that may just be fixing the symptoms instead of the root cause.

DanielRuf:
@alexisdiel it is just in the docs as far as I can see it.

PR #4815, opened 2017-06-14
SideNav menu was moving on X-axis

alexisdiel:
when menu contents overrides the height

### DIFF
--- a/js/init.js
+++ b/js/init.js
@@ -134,7 +134,7 @@
       }
     }
     if (is_touch_device()) {
-      $('#nav-mobile').css({ overflow: 'auto'});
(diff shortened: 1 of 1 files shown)


DanielRuf:
Fixes https://github.com/Dogfalo/materialize/issues/4814

PS: overflowY without quotes should also work.

DanielRuf:
This is the wrong file, please change the right lines in the sidenav.js file.

acburst:
I agree with @DanielRuf. I believe there is something wrong with the css that is causing this error and changing the overflow is not the way to go fixing this.

tomscholz:
https://github.com/Dogfalo/materialize/commit/a8d472fca1abf72afa638aa5e7e9bf1882d9cd15

PR #4854, opened 2017-06-20
move navbar-fixed z-index to nav element

Snewk:
### Description
The materialize navbar contains code to handle a fixed navigation bar, but the z-index property for this navbar is on the class `.navbar-fixed` which belongs to the parent element instead of on the actual `.navbar-fixed nav` element itself.

### Repro Steps
1: create a `<nav>` element.
2: add `class="navbar-fixed"` to the parent element (`<body>`).
3: scroll down to see elements on top of the fixed navbar instead of underneath it.

### Screenshots / Codepen
https://codepen.io/anon/pen/vZmWEy
currently

### Corrected
https://codepen.io/anon/pen/Kqmypx
fixed

### DIFF
--- a/sass/components/_navbar.scss
+++ b/sass/components/_navbar.scss
@@ -189,10 +189,10 @@ nav {
 .navbar-fixed {
   position: relative;
   height: $navbar-height-mobile;
-  z-index: 997;
 
   nav {
     position: fixed;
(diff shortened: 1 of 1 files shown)


Snewk:
should I move `height: $navbar-height-mobile;` to the nav element too?

PR #4859, opened 2017-06-21
use <body> instead of separate wrapper div for navbar-fixed

Snewk:
### Description
The materialize navbar contains code to handle a fixed navigation bar, but developers must add a separate div to handle the page offset and z-index. This commit moves the z-index property to the actual `.navbar-fixed nav` element itself, and handles offset using `top`on both elements

### Repro Steps
1: create a `<nav>` element.
2: add `class="navbar-fixed"` to the parent element (`<body>`).
3: scroll down to see elements on top of the fixed navbar instead of underneath it.
4: page offset does not function in this case.

### Screenshots / Codepen
https://codepen.io/anon/pen/vZmWEy
currently

### Corrected
https://codepen.io/anon/pen/Kqmypx
fixed

### DIFF
--- a/jade/page-contents/navbar_content.html
+++ b/jade/page-contents/navbar_content.html
@@ -195,11 +195,11 @@
 
         <h2 class="header">Fixed Navbar</h2>
         <p>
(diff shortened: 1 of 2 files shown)


Snewk:
this commit also includes updated documentation in jade/page-contents/

Snewk:
i didn't update the media queries in this pr. closing in favor of https://github.com/Dogfalo/materialize/pull/4915/commits/2bef95f2313aaecf59261da2b7ab9b2a5b4b86b5

PR #4860, opened 2017-06-21
Add portrait/landscape selector classes

rossjrw:
As suggested by myself and *upvoted* by DanielRuf in #4849 

Two new selectors have been added in addition to `.s#`, `.m#`, `.l#` and `.xl#`

These are `.P#` and `.L#` which are active for portrait and landscape screen sizes. Unsure of their interactions with the standard classes; would advise to avoid using both together.

Also added `.hide-on-portrait` and `.hide-on-landscape`, as well as the respective `.show...` classes.

Issue #4849, opened 2017-06-20
Add grid sizing classes for detecting landscape/portrait


rossjrw:
Adjusting the grid based on screen width using `.s6`, `.m12` and such is all well and good and are a good solution for most scenarios.

However, a more useful approach in my opinion would be to detect portrait vs landscape. Take the iPad Pro for example - the resolution is 1024x1366. 1024px falls under the `large` category for screens which is for all intents and purposes a desktop - however, the screen is still portrait and as such to display a desktop-oriented page may look silly.

Adding classes to detect portrait vs landscape would solve this issue and, for sites that only have a mobile version and a desktop version with no in-between, provide a more complete solution to responsiveness.

Detection of portrait vs landscape is easy using the CSS rules:
```css
@media screen and (orientation:portrait) { ... }
@media screen and (orientation:landscape) { ... }
```

My suggestion for the selectors for these classes would be `.P12` and `.L6`, using capitals to distinguish the landscape selector from the large-screen selector.

Thank you for your time.

PR #4863, opened 2017-06-21
Add portrait/landscape selector classes

rossjrw:
Corrected version of #4860 

As suggested by myself in #4849 

Two new selectors have been added in addition to `.s#`, `.m#`, `.l#` and `.xl#`

These are `.P#` and `.L#` which are active for portrait and landscape screen sizes. Unsure of their interactions with the standard classes; would advise to avoid using both together.

Also added `.hide-on-portrait` and `.hide-on-landscape`, as well as the respective `.show...` classes.

Also added `.offset-P#`, `.pull-P#` and `.push-P#` with respective versions for landscape.

### DIFF
--- a/sass/components/_global.scss
+++ b/sass/components/_global.scss
@@ -304,6 +304,16 @@ ul.staggered-list li {
     display: none !important;
   }
 }
+.hide-on-portrait {
+  @media #{$portrait} {
+    display: none !important;
+  }
+}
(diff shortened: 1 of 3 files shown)


bert-w:
not a fan of uppercased css classes though. It's very concise however it doesnt align with the rest of the classes

PR #4868, opened 2017-06-22
Incorrect input type in form example.

labnol:
Input type should be set to text for time picker example.


DanielRuf:
Any reason why? Date is used for datetime.

DanielRuf:
https://github.com/Dogfalo/materialize/pull/4882#issuecomment-310851766

I think text is not the right type, might need some refactoring imho (the datepicker uses date).

tomscholz:
Why not choose `type="time"` over `type="text"` ?

Dogfalo:
As of now text is the correct type, time will add other UI elements which are not easily overridden

Issue #1582, opened 2015-06-24
Spring / Thymeleaf checkbox compatibility issue


michalsukupcak:
Hello,

I'd like to let you know that there's a compatibility issue with Thymeleaf templating engine (used in Spring framework).

The issue in question:
Required structure of checkbox input in Materialize is following (`<p>` can be replaced for any other tag AFAIK):

```
<p>
    <input type="checkbox">
    <label> ... </label>
</p>
```

With CSS code used to target the label being:

```
[type="checkbox"] + label { ... }
```

Meaning "a label right after a checkbox". Problem is that Thymeleaf adds **additional `<input type="hidden">`** element after each checkbox input, **breaking the checkbox + label selector**. Following is actual Thymeleaf code I use in one of my projects. It uses a th:each loop to go over a list of items:

```
<div th:each="anArtwork : ${listOfArtworks}" class="row">
    <div class="col s12">
        <p>
            <input th:field="*{artworks}" th:value="${anArtwork.id}" type="checkbox" class="filled-in">
            <label th:for="${#ids.prev('artworks')}" th:text="${anArtwork.nameSk}"></label>
        </p>
    </div>
</div>
```

Which results into the following HTML code (generated repeatedly for each item in list with small differences in ID-related values):

```
<div class="row">
    <div class="col s12">
        <p>
            <input class="filled-in" type="checkbox" value="2" id="artworks2" name="artworks" />
            <input type="hidden" name="_artworks" value="on" />
            <label for="artworks2"> ... </label>
        </p>
    </div>
</div>
```

This of course breaks previous CSS selector for checkbox + label elements, as it adds additional element between them. Workaround I'm using adds additional selector to the CSS code in Materialize like this:

```
[type="checkbox"] + label,
[type="checkbox"] + [type="hidden"] + label { ... }
```

Which accounts for the additional hidden input and thus targets the required label. This is a Thymeleaf-specific solution only however and fastest one I could think of for my project.

My suggestion for a full compatibility fix/solution would be adding a mandatory wrapper element (like `<p>` or `<span>`) with a mandatory class (eg.: `.material-checkbox`) to the material checkboxes and target the inside checkbox and label elements using this class (`.material-checkbox [type="checkbox"]` and `.material-checkbox label`) instead of relying on the fact that the two elements will always be next to each other, which in Thymeleaf is not the case:

```
<p class="materialize-checkbox">
    <input type="checkbox">
    <label> ... </label>
</p>

[...]

PR #4874, opened 2017-06-23
Added Thymeleaf compatibility for checkboxes

andhen:
Thymeleaf adds hidden fields between the input and label of checkboxes, which is documented in issue #1582. This means styling of checkboxes is broken when using Thymeleaf.

This change fixes that by altering the rules for checkboxes to use the general sibling selector instead of adjacent sibling selector.

Performance of this solution is probably a little bit worse, but the difference should be negligeble since you probably won't have many siblings of a checkbox. In the same way, this breaks styling of any additional labels which are accidental siblings of the checkbox, but this should also be rare in practice.

### DIFF
--- a/sass/components/forms/_checkboxes.scss
+++ b/sass/components/forms/_checkboxes.scss
@@ -22,7 +22,7 @@ form p:last-child {
 // Checkbox Styles
 [type="checkbox"] {
   // Text Label Style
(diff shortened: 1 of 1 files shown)


Dogfalo:
We have a different change in mind that will fix this

Issue #4877, opened 2017-06-23
There not contain js file, CSS File


fumiya-kume:
materialize/templates/starter-template/

not contain materialize.js in js folder
not contain materialize.css in css folder

DanielRuf:
See my comment in your PR. The files are dynamically copied with each `grunt release`.

PR #4878, opened 2017-06-23
Add js, css file in starter template #4877 

fumiya-kume:


DanielRuf:
This is not needed because the files are copied by the `grunt release` task, see

https://github.com/Dogfalo/materialize/blob/master/Gruntfile.js#L297-L309

https://github.com/Dogfalo/materialize/blob/master/Gruntfile.js#L586-L605

Especially the process tasks.

DanielRuf:
@Dogfalo @tomscholz we can close this too.

PR #4888, opened 2017-06-26
Remove 'data-tooltip-id' reference from the origin on remove

jeroenqui:
Do not only remove the tooltip content element, but also the 'data-tooltip-id' reference to it from the origin.


acburst:
This is not how you remove a data attribute. Also make sure all checks are passing

jeroenqui:
Excuse me, that was sloppy and done too quickly.
I have made a new pull request with a proper and tested way of doing this.

PR #4893, opened 2017-06-27
Corrects the search input on navbar without affect the normal input search

alexisdiel:
I've create an example to demonstrate the problem

https://codepen.io/alexisdiel/pen/RgLqWX

### DIFF
--- a/sass/components/forms/_input-fields.scss
+++ b/sass/components/forms/_input-fields.scss
@@ -210,8 +210,8 @@ textarea.materialize-textarea {
 .input-field input[type=search] {
   display: block;  
   line-height: inherit;
-
-  .nav-wrapper  & {
+  
+  .nav-wrapper & {
     height: inherit;
     padding-left: 4rem;
     width: calc(100% - 4rem);
--- a/sass/components/forms/_input-fields.scss
+++ b/sass/components/forms/_input-fields.scss
@@ -215,6 +215,8 @@ textarea.materialize-textarea {
     height: inherit;
     padding-left: 4rem;
     width: calc(100% - 4rem);
+    border: 0;
+    box-shadow: none;
   }
 
   &:focus {
--- a/sass/components/forms/_input-fields.scss
+++ b/sass/components/forms/_input-fields.scss
@@ -207,12 +207,15 @@ textarea.materialize-textarea {
 
 /* Search Field */
 
-.input-field input[type=search] {
-  display: block;
+.nav-wrapper .input-field input[type=search] {
   height: inherit;
-  line-height: inherit;
   padding-left: 4rem;
   width: calc(100% - 4rem);
+}
+
+.input-field input[type=search] {
+  display: block;  
+  line-height: inherit;
 
   &:focus {
     background-color: $input-background;
--- a/sass/components/forms/_input-fields.scss
+++ b/sass/components/forms/_input-fields.scss
@@ -207,15 +207,15 @@ textarea.materialize-textarea {
 
 /* Search Field */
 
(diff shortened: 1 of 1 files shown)


DanielRuf:
I think

```Scss
.input-field input[type=search] {
  display: block;		
    line-height: inherit;		
    padding-left: 4rem;
    width: calc(100% - 4rem);
    .nav-wrapper  & {
      height: inherit;
    }
}
```

is much better

alexisdiel:
@Dogfalo ... much better with `.nav-wrapper &`

thnks

DanielRuf:
@alexisdiel please make sure to exactly copy my suggested code / change, your code is a bit different and has probaly some extra space before & too.

alexisdiel:
@DanielRuf ... I've include the  `padding-left: 4rem; width: calc(100% - 4rem);` on .nav-wrapper to fix the cursor position when the search input is not in the navbar, so it get the same input padding of all others inputs.

but if u thing that it's not relevant, I fix only the height.

alexisdiel:
I've adjusted the codepen to show you these modifications, click on the last search input to see.

DanielRuf:
@alexisdiel you remove the padding from the normal search input which is not wrapped in a nav-wrapper.

Try the following code on https://www.sassmeister.com/
```SCSS
.input-field input[type=search] {
  display: block;		
    line-height: inherit;		
    padding-left: 4rem;
    width: calc(100% - 4rem);
    .nav-wrapper  & {
      height: inherit;
    }
}

/* separator */

.input-field input[type=search] {
  display: block;  
  line-height: inherit;
  
  .nav-wrapper & {
    height: inherit;
    padding-left: 4rem;
    width: calc(100% - 4rem);
  }
}
```

alexisdiel:
@DanielRuf right, but the modification was to put the same padding of the other inputs (that not in nav-wrapper), which in my point of view was the initial idea set for navbar and not for the other forms, look at codepen.

DanielRuf:
@alexisdiel I think you do not fully understand what I mean. Your code removed these styles from the general search input which is not good. `.input-field input[type=search]` does not have the rules that it should have. My code fixes this.

@tomscholz your opinion about the changes in the PR?

alexisdiel:
@DanielRuf I understand that you are referring about the `padding-left: 4rem; width: calc(100% - 4rem);`, but I put it together with `height: inherit;` to get the same padding of all others inputs fields and only use the padding on nav-wrapper.

DanielRuf:
> but I put it together with height: inherit; to get the same default padding of all others inputs fields and only use the padding on .nav-wrapper.

Sorry but this is not the case. Please take a look at the compiled code at https://sassmeister.com

The padding is just on element in nav-wrapper in your code.

alexisdiel:
It generates exactly what I want.
Take a look on https://codepen.io/alexisdiel/pen/RgLqWX
You will see these modifications, and see how actual padding goes wrong when input search it's not in `.nav-wrapper`

```
.input-field input[type=search] {
  display: block;
  line-height: inherit;
}
.nav-wrapper .input-field input[type=search] {
  height: inherit;
  padding-left: 4rem;
  width: calc(100% - 4rem);
  border: 0;
  box-shadow: none;
}
```

DanielRuf:
But the width and padding are missing here:

```
.input-field input[type=search] {
  display: block;
  line-height: inherit;
}
```

cc @tomscholz

alexisdiel:
@DanielRuf did u saw the codepen?

acburst:
This makes sense to me. Adding the parent selector removes the necessity of having those padding and width by default.

Issue #3450, opened 2016-08-02
Background page is scrolling in Modal popup


ragu-nathan:
After opening the Modal page, the background scroll is working.. this is very much annoying in mobile only. It is working fine in Web page. Could you please help me to resolve this. Please refer this: https://jsfiddle.net/ragu_nathan/w9tmwv2z/2/
Thanks

jdkealy:
👍  same issue here
works the same on the modal demo, background scrolls on mobile
http://materializecss.com/modals.html#!

rubendinho:
any update/suggested fixes on this?  thanks

haojia321:
Do this: Set your body width: 100%; and position: fixed; when modal is displayed. 
Only set those 2 styles when modal is showed.

I will try to make a PR for this issue. but for now you can try my above solution.

acburst:
I can't seem to reproduce this issue on the latest version. Do you have a codepen?

haojia321:
@acburst This is mobile specific, I tested on my iphone it is reproducible. 
just open modal from http://materializecss.com/modals.html# from iphone will see the issue. 
Im not sure about android though.

tomscholz:
@acburst After opening a modal, try to scroll on the overlay...

Greta:
This is still a bug.  I believe it can be resolved if you add the `overflow-y: hidden` style to the `<html>` tag as well as the `<body>` tag.

mangakid:
+1

acburst:
Modal was been rewritten in alpha v1.0.0, if this issue still exists please open a new issue

vishwasvogga:
body{
  overflow-y: scroll !important;
}

This will solve the issue..

PR #4894, opened 2017-06-27
#3450 Modal mobile scrolling fix

Greta:
Scrolling fix for mobile browsers.
Also Minor cleanup for var consistency.

### DIFF
--- a/js/modal.js
+++ b/js/modal.js
@@ -32,7 +32,7 @@
           $modal.removeClass('open');
 
           // Enable scrolling
-          $('body').css({
+          $('body, html').css({
             overflow: '',
             width: ''
           });
@@ -72,10 +72,10 @@
         };
 
         var openModal = function($trigger) {
-          var $body = $('body');
+          var $body = $('body'),
+              $html = $('html');
           var oldWidth = $body.innerWidth();
-          $body.css('overflow', 'hidden');
-          $body.width(oldWidth);
+          $body.width(oldWidth).add($html).css('overflow', 'hidden');
 
           if ($modal.hasClass('open')) {
             return;
@@ -90,7 +90,7 @@
           $modal.data('overlay-id', overlayID).css('z-index', 1000 + lStack * 2 + 1);
           $modal.addClass('open');
 
-          $("body").append($overlay);
+          $body.append($overlay);
 
           if (options.dismissible) {
             $overlay.click(function() {


js/modal.js
```diff
@@ -76,6 +79,7 @@
           var oldWidth = $body.innerWidth();
           $body.css('overflow', 'hidden');
           $body.width(oldWidth);
+          $('html').css('overflow', 'hidden');
```
DanielRuf: I guess we can add `$('html')` to https://github.com/Greta/materialize/blob/71fe188a14b091c1a415e35cb9a34480b905f58c/js/modal.js#L78 to cache the lookup.

Then you can merge this with line 80.

js/modal.js
```diff
@@ -36,6 +36,9 @@
             overflow: '',
             width: ''
           });
+          $('html').css({
+            overflow: ''
+          });
```
DanielRuf: Can we merge this with the rule above this one?

js/modal.js
```diff
@@ -90,7 +94,7 @@
           $modal.data('overlay-id', overlayID).css('z-index', 1000 + lStack * 2 + 1);
           $modal.addClass('open');
 
-          $("body").append($overlay);
+          $body.append($overlay);
```
DanielRuf: Good catch =)

js/modal.js
Greta: I'll clean that up.  I could instead, if you'd rather, make these changes by adding a custom class to `<html>` instead of injecting inline css.  I believe that would be a less intrusive way to add the styling.

acburst:
What exactly is this fixing? Those who approved this change: What browsers / phones did this fix work on? I've tested this on an iPhone 6 and this did not fix the bug. And this bug is not reproducible on an android phone

tomscholz:
Tested on Moto G running LineageOS Jelly Browser. See https://goo.gl/photos/HmNY3kY1bXwsvm3z5. Oddly enough, the fix does not work anymore in the latest version of jelly. Jelly btw uses Webview.

tomscholz:
@acburst I was able to reproduce it on an iPhone in Safari running IOS 10.3.2 . Here is a link https://goo.gl/photos/9Yc15mduvDGmbLtn7

acburst:
@tomscholz So this fix does not work for iPhones?

tomscholz:
@acburst Nope. Just had the opportunity to test it now :(

PR #4899, opened 2017-06-28
Make sure form autocomplete handler is bound only once on each input

ClemArt:
# Context
I was building a *React* app. As such my components are often updated but rarely destroyed.

In my component I call $.fn.autocomplete() on input fields, attaching autocomplete handler to them.

This is done each time my component gets new data, to update the available options in the autocomplete.

# Issue
The more I call the $.fn.autocomplete(), the more the page slows until it ultimately crashes.

I noticed there was more and more calls to the onAutocomplete handler, growing with each autocomplete init.

# Cause
It seems that the event handler passed as onAutocomplete option is attached on each $.fn.autocomplete() call. But is never detached.

So the page is accumulating handlers for the same input, all of them executing at the same time.

This is a codepen demonstrating 1000 bindings with a simple loop. https://codepen.io/ClemArt/pen/KqZpML . You'll see 1000 results in the console, slowing drastically the page...

# Proposition
I propose the following fix, just following what you did for all the other events : unbind with OFF and bind the new event with ON.

Simple enough, and it did the job in my case. My app works flawlessly now.

------------
I'm open to any discussion if needed,
All the best !

PS : Thanks for this awesome Framework !


ClemArt:
I assume the history is really not clean enough to directly merge, but the proposed fix is just 2 lines, so you can eventually copy / paste it in your own branch.

DanielRuf:
No, this is not clean. Please do not change the dependency or add the unused prefixer file back and just push the needed change. These are also separate things.

PR #4919, opened 2017-06-30
Ability to use autocomplete with "StartsWith" instead of "Contains"

shaylevi2:
Changes:
* Ability to use autocomplete with "StartsWith" instead of "Contains" (also in chips module and any other auto-complete module)
* Ability to include full-match in the autocomplete results (also in chips module and any other auto-complete module)

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -395,8 +395,9 @@
               if (val.length >= options.minLength) {
                 for(var key in data) {
                   if (data.hasOwnProperty(key) &&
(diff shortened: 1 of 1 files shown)


acburst:
If you want to add an option you should make sure it has a default value as well. And add it to the docs.

shaylevi2:
The default value is when it's off.

Regarding adding to the docs, I'd prefer if a native English speaker would
do it.

PR #4921, opened 2017-06-30
Using flex to correct the grid cols bug using different heights

inventti-leonardo:
## Proposed changes
Using flex to correct the grid cols bug using different heights

https://codepen.io/anon/pen/wemvrb

## Types of changes
What types of changes does your code introduce? Put an `x` in all the boxes that apply:

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
Go over all the following points, and put an `x` in all the boxes that apply. If you're unsure about any of these, don't hesitate to ask. We're here to help!

### DIFF
--- a/sass/components/_grid.scss
+++ b/sass/components/_grid.scss
@@ -57,6 +57,8 @@
   margin-left: auto;
   margin-right: auto;
   margin-bottom: 20px;
+  display: flex;
+  flex-wrap: wrap;
 
   // Clear floating children
   &:after {


DanielRuf:
Not really a bug. Close the row after 12 columns (4 x 3). Equal height has to be done manually if it is needed. People can then use any of the available solutions like flexbox.

Dogfalo:
Adding display: flex is not a good solution, and can break a lot of existing projects.

iMerica:
@inventti-leonardo Thank you 🙏 . This was a lifesaver

PR #4946, opened 2017-07-08
Fix a bug with proper initialization of select tag

VeeK727:
## Proposed changes
$select.data() was being used to add a data-attribute which was not actually appearing on the select tag. This was fine for normal usage but it leaves trash when updating or destroying. That was because the check for initialized select was looking for the said data-attribute.

I've now used $select.attr() instead and everything is working properly now.

## Types of changes

## Checklist:

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -509,7 +509,7 @@
       }
 
       var uniqueID = Materialize.guid();
-      $select.data('select-id', uniqueID);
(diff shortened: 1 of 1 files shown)


Dogfalo:
You missed one of them so I just made the change myself here: f4b31a29cbecad3aaa48ba56f6944b0a7210a9f2

VeeK727:
Great. Thanks

Issue #4212, opened 2017-02-09
Update velocity, 1.2.2 > 1.2.3 || 1.4.2


smeijer:
Although 1.4.2 would be better, even the 1.2.3 release has a useful new method, called `finishAll`.

According to https://github.com/Dogfalo/materialize/pull/2252 , 1.2.3 should already be there, but I guess it has been rolled back some day.

smeijer:
Refering to this line:

https://github.com/julianshapiro/velocity/blob/1.2.3/velocity.js#L2077

tomscholz:
Pull requests are always welcome

Dogfalo:
Updated to 1.4.2 in v1-dev

PR #4947, opened 2017-07-08
Updated velocity from 1.2.3 to version 1.5.0

tomscholz:
- Closes Dogfalo/materialize#4212

Non breaking change in my opinion


Dogfalo:
https://github.com/julianshapiro/velocity/issues/781

Seems like there is a major bug present since 1.4.3. Did you test out the components to see if they are still working?

Issue #2228, opened 2015-10-31
Allow using material-icons also on <span>, not only <i>


radek-anuszewski:
Hi!

Now, `material-icons` class can be used on `<i>` tag, when I tried to use it on `<span>` tag I found it not working as expected: [Plunker demo with i and span tags inside button](http://plnkr.co/edit/iOACScvQyBxlCsFbKfQK?p=preview). Many times I saw that `span` tag is used as inline wrapper for something, [Font Awesome examples](https://fortawesome.github.io/Font-Awesome/examples/) works on `<i>` tag but `<span>` tag also can be used and it works well - of course it is not giant issue, but maybe using `material-icons` on `<span>` tag also should be possible?

Thanks for your attension!

tomscholz:
Why?

radek-anuszewski:
@tomscholz Because, as one guy answered on my [StackExchange question](http://webmasters.stackexchange.com/questions/83446/seo-aria-label-vs-sr-only), `<i>` tag has `alternate voice or mood` meaning.

DanielRuf:
Bootstrap 3 uses
```HTML
<span class="glyphicon glyphicon-ok form-control-feedback" aria-hidden="true"></span>
```

The `aria-hidden="true"` is important for screen reading soltutions and they switchted to `span` in the official docs.

Dogfalo:
`span` has no meaning alone. `i` signifies that the section of text has a different nature from that of the normal prose that surrounds it. While it is very true that using `i`  for icons isn't entirely ideal, it seems more fitting than span as the text inside icons.

PR #4949, opened 2017-07-09
Allow using material design icons in combination with `span` tags

tomscholz:
## Proposed changes

These changes allows users to use the `span` tag in combination with the material design icons.
Solves https://github.com/Dogfalo/materialize/issues/2228

## Types of changes

## Checklist:


Dogfalo:
keeping it as `<i>` is better as it allows the user to use different icon sets more easily

PR #4963, opened 2017-07-14
Fix for issue 4962

camro:
List style is not applied if browser default is in another ul

screenshot


DanielRuf:
Please redo the commit with the correct line endings and indentation.

PR #4980, opened 2017-07-18
Added container option to timepicker, issue #4900

### LINKED ISSUE #4900: Timepicker container option
### Description
Materialize timepicker option container now working.

### Example
```
$('.timepicker').pickatime({
 container: '#div'
  });
```
### Solution
Is there any solution e.g. .appedTo('body') ?

Y0lk:
Makes sense to have the container option, fixes #4900

### DIFF
--- a/js/date_picker/picker.time.js
+++ b/js/date_picker/picker.time.js
@@ -347,8 +347,7 @@
 		ampmclickable: true,  // set am/pm button on itself
 		darktheme: false,			 // set to dark theme
 		twelvehour: true,      // change to 12 hour AM/PM clock from 24 hour
-		vibrate: true,          // vibrate the device when dragging clock hand
-    container: 'body'   //container
+		vibrate: true          // vibrate the device when dragging clock hand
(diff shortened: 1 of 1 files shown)


Dogfalo:
appending after the input needs to be the default

Y0lk:
Wouldn't it make sense for it to have the same behavior as the datepicker by default though? The datepicker gets added to the body

Y0lk:
Nevermind, I'm wrong. Will update the pull request right away

Y0lk:
done

Dogfalo:
Won't  `if(this.options.hasOwnProperty('container'))` always return true here?

Y0lk:
It won't because the option isn't defined in the defaults, so if you don't specify the container, the property doesn't exist and it'll default to adding it after input. I did it this way simply because that's exactly the way it's implemented in materialize clockpicker.

We could change this to check for an empty string if you prefer (and have the default container option as empty string)

PR #5022, opened 2017-07-24
A Simple Solution For Swipeable Tabs Fixed Height

atemiz:
When tabs swipeable option is setted to true, height is fixed 200px.
To solve this, setting carousel-item height to auto which is inside of
tabs-content class can be useful.


acburst:
This doesn't work because the height is set on the wrapping carousel.

atemiz:
Oh I see. Maybe we can move this to after carousel height.

PR #5042, opened 2017-07-29
add keyboard access to cards

neko2016:
## Proposed changes

Previously the cards is opened only via mouse click. I add keyboard access so that when navigating the website with keyboards, user can use the ENTER key to open and close the card component. Keyboard access is an important feature in web accessibility. jQuery has built-in accessibility feature.

I only change two lines of codes as follows
`$(document).on('keypress click', '.card', function (e) {
      if (e.which === 13 || e.type === 'click') { ...... } 
})`

instead of 
` $(document).on('click.card', '.card', function (e) {......})`

## Types of changes

## Checklist:

### DIFF
--- a/js/cards.js
+++ b/js/cards.js
@@ -1,36 +1,36 @@
 (function ($) {
   $(document).ready(function() {
-
-    $(document).on('click.card', '.card', function (e) {
-      if ($(this).find('> .card-reveal').length) {
-        var $card = $(e.target).closest('.card');
-        if ($card.data('initialOverflow') === undefined) {
-          $card.data(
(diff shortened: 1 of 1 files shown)


DanielRuf:
Isn't `click.card` an event by Materialize so we need it instead of `click`?

neko2016:
Shall we change it to the following instead? Then we keep the `.card` namespace for the both the `click` and `keypress` events, and it's not breaking the original codes. 

``` 
$(document).on('keypress.card click.card', '.card', function (e) { 
    if (e.which === 13 || e.type === 'click') { ...... } 
})
```

PR #5061, opened 2017-08-02
Enable Checkbox and RadioButtons to be clickable when are inside input-field div

alexisdiel:
I created the following CodePen to show the problem and solution.
https://codepen.io/alexisdiel/pen/gxwPWj

This fix makes it possible to use checkboxes and radio-buttons inside the input-field div so that they inherit the CSS (margin, padding...).

## Types of changes

## Checklist:

### DIFF
--- a/sass/components/forms/_checkboxes.scss
+++ b/sass/components/forms/_checkboxes.scss
@@ -31,6 +31,7 @@ form p:last-child {
     line-height: 25px;
     font-size: 1rem;
     user-select: none;
+    pointer-events: auto;    
   }
 
   /* checkbox aspect */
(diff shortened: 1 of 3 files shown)


alexisdiel:
Fix issue: https://github.com/Dogfalo/materialize/issues/5062

MathieuDerelle:
I've tested your changes by adding this

```
[type="checkbox"] + label,
[type="radio"]:not(:checked) + label,
[type="radio"]:checked + label {
  pointer-events: auto;
}
```

but the specificity of `.input-field label` is winning over `[type="checkbox"] + label` so nothing was fixed (the extra class `.fixed`  in your codepen example makes it ok)

So, this did the trick : 
```
.input-field {
  [type="checkbox"] + label,
  [type="radio"]:not(:checked) + label,
  [type="radio"]:checked + label {
    pointer-events: auto;
  }
}
```

Dogfalo:
See comment at #5062

Issue #5105, opened 2017-08-14
window.jQuery is ignored, if page has require function (part of babel module)


VadimZhiltsov:
## Summary
There is an issue with jQuery avialability check.

## Expected Behavior
Materialize should use existing jquery instead of including another one

## Current Behavior
Materialize.js uses own version of jquery, included by `require` function, instead of using jQuery from global scope.

## How to reproduce
https://jsfiddle.net/puz2m9hd/1/

## Context
It appears when page has included jquery and page has require function. As example I've included jquery from CDN by <script> tag and bundled my javascript by Babel, so there is require function in global scope.

Dogfalo:
Jsfiddle seems to be doing something odd as jQuery is not defined at all. It works as intended here
https://codepen.io/Dogfalo/pen/XaeyOX

VadimZhiltsov:
@Dogfalo yeah, you are right. 
Looks like hoisting works differenly in global scope rather than function scope. And jsfiddle wraps everything in self-called function. So we can reach the same result on codepen
https://codepen.io/anon/pen/brYGjm

The same does Babel, when I include materialize by import as npm module. So it force additional jquery include despite I have jquery on the page included from CDN

Dogfalo:
In your example, you wrap the entire code in an IIFE which is causing the problem, but this isn't reflected in what materialize.js is doing. So I don't believe this is the issue.

VadimZhiltsov:
@Dogfalo materialize distributed as npm module and the most part of bundling tools wraps modules in IIFE, so we should change the materialize.css to work properly with these bundlers.
We must do changes in this code, if we want materialize.css be working with babel.

Looking on the bootstrap.css, it just takes jQuery from the global namespace. Probably we should do the same with materialize.css.

Dogfalo:
Fixed in v0.100.2

PR #3782, opened 2016-10-16
update checking for jquery strategy

forehalo:
note that users always define or require `jQuery` and `$` in  global `window` variable, but in some context, 'jQuery' not representing `window.jQuery`. This PR make sure materialize and other library use the same instance of jquery but not init materialize's own which may cuz some conflict.

PR #5106, opened 2017-08-14
fix issue #5105

VadimZhiltsov:
fix issue #5105 (see description inside)


DanielRuf:
Sorry but this is not the right fix.

tomscholz:
#3782 #3552

Issue #4413, opened 2017-03-27
Blinking cursor in Select on IE11


stsje:
Focusing a Select in IE11 will add a blinking cursor that is visible through the dropdown.

Bug can be reproduced using the selects in the documentation http://materializecss.com/forms.html

Any CSS fix for this?

DanielRuf:
Try the following CSS code and let me know if this fixes your problem.
```CSS
input.select-dropdown:focus {
    text-indent: -9999999em;
}
```

stsje:
It seems to be working.
**But!** there seems to be a visible transition effect.

Is it necessary to have transition on all properties? What properties need to have transition?
`transition: all 0.3s;`

DanielRuf:
That is a good question. @tomscholz do you have some more information?

tomscholz:
These issues are all duplicates:
https://github.com/Dogfalo/materialize/issues/4413
https://github.com/Dogfalo/materialize/issues/4315
https://github.com/Dogfalo/materialize/issues/4306
https://github.com/Dogfalo/materialize/issues/3715
https://github.com/Dogfalo/materialize/issues/3171
https://github.com/Dogfalo/materialize/issues/630

For the future discussion, please use this issue instead.
https://github.com/Dogfalo/materialize/issues/3790

madhmeht:
The css fix causes another problem. After selecting an option from keyboard, the field stays blank. After clicking somehwere else, the value appears.

DanielRuf:
@madhmeht please provide a coden which shows this issue.

madhmeht:
@DanielRuf here is the codepen link https://codepen.io/anon/pen/KyRRJy

kaszubski:
> @DanielRuf here is the codepen link https://codepen.io/anon/pen/KyRRJy

Thanks 🙏🙏🙏

Issue #3715, opened 2016-10-02
IE11 appear into input select a cursor


gomezv9:
Hi

The input select in IE11 hava a bug, this show me like a "Cursor" ( if I was going to write something ) when I open de input select.

It only appear in IE11

Any help with this bug..

Thanks

tomscholz:
These issues are all duplicates:
https://github.com/Dogfalo/materialize/issues/4413
https://github.com/Dogfalo/materialize/issues/4315
https://github.com/Dogfalo/materialize/issues/4306
https://github.com/Dogfalo/materialize/issues/3715
https://github.com/Dogfalo/materialize/issues/3171
https://github.com/Dogfalo/materialize/issues/630

For the future discussion, please use this issue instead.
https://github.com/Dogfalo/materialize/issues/3790

PR #5108, opened 2017-08-14
Blinking cursor in Select on IE11

pharindoko:
On open the select shows a blinking cursor in IE 11 (Closed Issues #4413, #3715)
See in demos: http://materializecss.com/forms.html

As this issue hasn`t been fixed for IE 10 / 11 I thought I propose this fix which has been mentioned in closed issue #4413.

I added for transition: none but it takes some time.
Maybe you have a better solution for this @Dogfalo

I have tested the fix in Chrome,FF and IE 11 and it works equal for all 3 browsers

### DIFF
--- a/sass/components/forms/_select.scss
+++ b/sass/components/forms/_select.scss
@@ -118,6 +118,11 @@ select:disabled {
   user-select: none;
 }
 
+.select-wrapper input.select-dropdown:focus {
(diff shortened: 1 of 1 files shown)


Dogfalo:
Have you tried adding the attribute `readonly` to the input instead? That might be a cleaner solution.

PR #5175, opened 2017-09-07
Restructure sass imports

tomscholz:
- Moved variables and normalize outside the components folder, since they just are not components.


flexbox:
The actual [version of Normalize.css](https://necolas.github.io/normalize.css/) is 7.0.0

@tomscholz Do you think we can bump it on the same PR?

DanielRuf:
Imo this should be integrated with npm and an import rule, not in our sass folder.

flexbox:
Totally agreed!

But Maybe there are some custom rulez like in this file `js/date_picker/picker.date.js`

DanielRuf:
Which rules do you mean?

flexbox:
Just [watched the history](https://github.com/Dogfalo/materialize/commits/master/sass/components/_normalize.scss), And I think we can bump the version directly from the npm registry :rocket:

tomscholz:
normalize has been updated to 7.0 in the v1 https://github.com/Dogfalo/materialize/commit/505ff791ad607b7b757945f68374f5a9db714081

PR #5182, opened 2017-09-10
Added remote ajax call to autocomplete

AdamSGit:
Hi. First of all, thanks for this nice framework. This is the first time I contribute to this repository. I added this feature because I need it in some project. 

## Proposed changes
This PR add the possibility to the autocomplete framework to fetch remote server to get results. I did this with the care of not breaking anything existing in mind. So, basically, if you replace the "data" parameter by an "url" parameter, the plugin will change its behavior to fetch results on the url. If you dont, no changes will be made at how the plugin work.

About how the Ajax part works, please note the following : 

- Request is send x ms after the user stop typing. x can be set with the "debounceTime" options, default is 500.
- If a request start but a previous one is still running, it cancel the old one to avoid unpredictable behavior
- The request is sent by GET, according to HTTP protocol specification. The string parameter name is "val" 
- The server output must be an objects array in XML or JSON. The needed key is "value" (the text), and an optional "img" key can be used to pass image url. Any other key -> val present in the objects will be available in the onAutocomplete callback (see example code for this in the updated doc).

I have updated the doc but since i'm not a native english speaker, things can be not so well said, you can check it and make corrections if it need some.

If you want others changes about this, feel free to ask.

Screenshot would be the same as the actual autocomplete, and I can't provide a pen since it need a server call to give back results. You will however find some code in the doc page I updated.

## Types of changes

## Checklist:

### DIFF
--- a/jade/page-contents/forms_content.html
+++ b/jade/page-contents/forms_content.html
@@ -911,6 +911,29 @@
     minLength: 1, // The minimum length of the input for the autocomplete to start. Default: 1.
   });
         </code></pre>
+
+        <h4>Remote ajax call</h4>
+        <p>To fetch a remote server, you need to replace the "data" parameter with the "url" parameter containing the remote content url.</p>
+        <p>Few things to note :</p>
+        <ul>
+          <li>The request is send by GET according to HTML protocol spec for fetching content</li>
+          <li>The get variable name containing the search is "val" </li>
+          <li>Your server must respond in XML or JSON</li>
+          <li>The output should be an object's array. Each object must have at least a "value" key containing the text. It can also have an "img" key containing the optional picture url. <b>Any other key present in the response objects can be retrieve in the onAutocomplete callback. See below :</b></li>
+        </ul>
+        <pre><code class="language-javascript">
+  $('input.autocomplete').autocomplete({
+    url: "/my/api/users/search"
+    limit: 20, // The max amount of results that can be shown at once. Default: Infinity.
+    onAutocomplete: function(val, data) {
+      // Callback function when value is autcompleted.
+      // If your response objects has and "id" key, you can access it like that
+      $('#my-hidden-user-field').val(data.id);
+    },
+    minLength: 1, // The minimum length of the input for the autocomplete to start. Default: 1.
+    debounceTime: 500 // Time after you stop typing for the request to be sent. Default: 500
+  });
+        </code></pre>
       </div>
 
 
--- a/js/forms.js
+++ b/js/forms.js
@@ -310,10 +310,28 @@
         data: {},
         limit: Infinity,
         onAutocomplete: null,
-        minLength: 1
+        minLength: 1,
+        url: null,
+        debounceTime: 500
       };
 
       options = $.extend(defaults, options);
+
+      // Debounce function
+      function debounce(func, wait, immediate) {
+        var timeout;
+        return function() {
+          var context = this, args = arguments;
+          var later = function() {
+            timeout = null;
+            if (!immediate) func.apply(context, args);
+          };
+          var callNow = immediate && !timeout;
+          clearTimeout(timeout);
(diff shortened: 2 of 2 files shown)


jade/page-contents/forms_content.html
```diff
@@ -911,6 +911,29 @@ <h2 class="header">Autocomplete</h2>
...
         </code></pre>
+
+        <h4>Remote ajax call</h4>
+        <p>To fetch a remote server, you need to replace the "data" parameter with the "url" parameter containing the remote content url.</p>
+        <p>Few things to note :</p>
+        <ul>
+          <li>The request is send by GET according to HTML protocol spec for fetching content</li>
+          <li>The get variable name containing the search is "val" </li>
+          <li>Your server must respond in XML or JSON</li>
+          <li>The output should be an object's array. Each object must have at least a "value" key containing the text. It can also have an "img" key containing the optional picture url. <b>Any other key present in the response objects can be retrieve in the onAutocomplete callback. See below :</b></li>
```
storm1er: I think we should display an example of the result you expect instead of explanation.

AdamSGit: I totally agree, but for this, it takes to have an api endpoint to fetch data to autocomplete.

js/forms.js
```diff
@@ -310,76 +310,185 @@
...
+          if(xhr) xhr.abort();
+
+          xhr = $
+            .ajax({
+              type: 'GET',
+              url: options.url,
+              data: { val: val }
+            })
+            .always(function() { xhr = null })
+            .fail(function(a,b,c) { console.log(a,b,c) })
```
storm1er: Displaying a toast instead of silence death ?

jade/page-contents/forms_content.html
storm1er: I was talking about writing an exemple in this text like this : 

> Here's an example of what we expect of `http://your-awesome-api.com?val=google` : 
> 
> ```json
> {"result" : [
>   {
>      "value":"google-map",
>      "img":"url-to-google-map.png"
>   },
>   {
>      "value":"google-doc",
>      "img":"url-to-google-doc.png"
>   },
>   ...
> ]}
> ```

AdamSGit: You're right. I will update the doc this evening.

js/forms.js
AdamSGit: Well, the user doesn't have to know that the Ajax call failed. The console is here for the dev during the dev process. After that, if the request fail, the behavior is the same than when there's no result since before my PR: nothing happend.

Dogfalo:
As a heads up you might have seen on the homepage, we are rewriting our plugins in 1.0. So unless it's unlikely we'll be able to merge this. At most we would be able to take concepts from it.

AdamSGit:
No problem. So v1 can be a good occasion for you guys to add this feature. If you need a hand I'll be your man ;)

storm1er:
@Dogfalo @adamansky 
As some people are already running with 99.* version, some people will still want/need this feature with old version. This could be usefull to display something like "*if you want this feature with older version, you can use this fork ...*". Do you agree ?

Dogfalo:
I don't think so, it isn't good to link to an old version with an unsupported feature

gg4u:
Hi thank you @AdamSGit 
I also looking at flexibility of custom templates.

One would like to add more text and image to results of autocomplete.
Custom templates would be perfect (In my use case, I'd like to use cards and have more space to images and other description fields, for each found element).

The other aspect, is UX and intervene on the CSS
The behaviour I like most for an autocomplete is to have a scrollable list that goes _under_ the input field, like Google Maps to say an example.

I tried `position : fixed  ` on the div containing the input, but the list elements (`<ul class="drop-down">`) are generated in a way that inherit fixed position.

I wish you find useful feedback for rewriting the plugins.

Issue #5023, opened 2017-07-24
Add new and missing tests


tomscholz:
We want to provide new and more tests for our components. If you want to help, check out our [jasmine testing guide][1]. If you have any questions or feedback, don't hesitate to use the comment function ;) 

Theses components are lacking any tests:
- [ ] Carousel
- [ ] Slider
- [ ] Breadcrumb
- [ ] Pagination
- [ ] SideNav
- [ ] FeatureDiscovery
- [ ] Pushpin
- [ ] Forms
  - [ ] DatePicker
  - [ ] TimePicker
  - [ ] Checkboxes
  - [ ] Switches
  - [ ] Input fields

[...]

[1]: https://github.com/Dogfalo/materialize/blob/master/CONTRIBUTING.md#jasmine-testing-guide

storm1er:
@tomscholz I'll do some, but since this is from July, is it up to date ?
I'm checking sources & PR . . .

PR #5202, opened 2017-09-15
Added tests for the classic input fields.

deather:
## Proposed changes
I propose multiple tests about input fields #5023. I give you the list of field test :
- text
- password
- email
- url
- tel
- number
- search
- textarea

Tests added should cover all interactions that use could do.
The pull request may throw an error due to a bug about reset action on input. When the user click on reset button the label shouldn't be active.

## Types of changes

## Checklist:

### DIFF
--- a/tests/spec/forms/inputfields/inputfieldsFixture.html
+++ b/tests/spec/forms/inputfields/inputfieldsFixture.html
@@ -1,0 +1,60 @@
+
+<form>
+  <div class="row">
+    <div class="col s12 input-field">
+      <input id="input_text" type="text"/>
+      <label for="input_text">Materialize CSS Text</label>
+    </div>
+  </div>
+
+  <div class="row">
+    <div class="col s12 input-field">
+      <input id="input_password" type="password"/>
+      <label for="input_password">Materialize CSS Password</label>
+    </div>
+  </div>
+
+  <div class="row">
+    <div class="col s12 input-field">
+      <input id="input_email" type="email"/>
+      <label for="input_email">Materialize CSS Email</label>
+    </div>
+  </div>
+
+  <div class="row">
+    <div class="col s12 input-field">
+      <input id="input_url" type="url"/>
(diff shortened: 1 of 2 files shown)


storm1er:
image

deather:
@storm1er I'm using node in version 7.5.0 and npm in version 5.4.1. Like I said in my previous comment, this test shouldn't work due to a bug.
The use case test is when a user fill an input and click on reset button. The input is clear but the label keep the class CSS **active**, but the input have no value so the label shouldn't keep the class CSS **active**.

storm1er:
I'm using your code, aaannnnd it works well xD this is a missing feature ...  
When i read "reset" I though "when empty by user" // keyup as materialize do it, not with a reset button

deather:
@storm1er : No problem. My explanations were not really clear :s.

storm1er:
@deather please check this also =) https://github.com/deather/materialize/pull/1

deather:
@storm1er : it is good for me, I merge your work :)

storm1er:
I realize, I there somewhere on the web, an `input[type=reset]` ?
It doesn't even render correctly with materializecss ...

deather:
It's not deprecated https://www.w3.org/TR/html5/forms.html#reset-button-state-(type=reset)

This input is just here to trigger the reset event, I think it's not really important to have a good render

storm1er:
The real question was : Is this test-case necessary ?
I don't see any PR/Issue about this + MaterializeCss didn't stylized this button ... 

If it is, we should also correct `input[type="reset"]`'s style.

deather:
I write the test because in the forms.js there were these lines :
https://github.com/Dogfalo/materialize/blob/9bc43a1199ad5dfb78d58ba47726ab039218a939/js/forms.js#L35

storm1er:
Well, I realized that `input[type="button"]` and `input[type="submit"]` wasn't stylized anyway : we're using `.btn` class instead.

-> working on fixing this reset bug =)

PR #5212, opened 2017-09-18
Adding test on form switches. #5023

deather:
## Proposed changes
This changes proposed to add test on switches. #5023 

## Types of changes

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "materialize-css",
-  "version": "0.100.1",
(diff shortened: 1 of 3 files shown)


Dogfalo:
I don't think it's a good idea to add in the color checking, as that is liable to change at any time.

deather:
I'm agree with you, it is not the best method. Do you have a better strategy to test this part ?
A second question comes after some reflexion : Are switches testable ?

Issue #5222, opened 2017-09-23
Margins removed for .card-content p


apancutt:
Is there a reason for removing the margins on `<p>` elements within `.card-content`?

https://github.com/Dogfalo/materialize/blob/9bc43a1199ad5dfb78d58ba47726ab039218a939/sass/components/_cards.scss#L146

I tend to restore the margins locally in my projects but was wondering if it could be 'fixed' upstream.

Apologies if this has been documented/discussed before, but I couldn't find anything.

DanielRuf:
I guess because of the bottom margin.

apancutt:
Would you be open to removing top margin for `p:first-child` and bottom margin for `p:last-child` only? Will make a PR if you like, but I guess the focus is on the 1.0 release?

DanielRuf:
The current focus is on the 1.0 release but you can still make a PR for `> p:first-child` and so on.

IAMtheIAM:
Yes, good fix, please merge for upcoming release.

IAMtheIAM:
@apancutt are you going to make a PR or should I?

apancutt:
Already created at #5224

IAMtheIAM:
Oh yea that right, I saw that already. Thanks.

PR #5224, opened 2017-09-25
Restore margins for <p> within .card-content

apancutt:
Restores margins for paragraphs within `.card-content`.

Fixes issue #5222


IAMtheIAM:
Thank you, I was just about to ask why the margins were gone.

PR #5233, opened 2017-09-27
Allows element with .waves-effect to be centered

IAMtheIAM:
This fixes elements with `.waves-effect` class added to this. Currently they cannot be centered without overriding the `display: inline-block` style that was set here. I do not see a reason for setting that property as it appears to do nothing. Removing it allows for easier centerting of buttons with `.waves-effect` class added to them

## Proposed changes

Remove `display: inline-block` from `.waves-effect` class.

## Types of changes

## Checklist:

- [ ] I have read the **[CONTRIBUTING document](https://github.com/Dogfalo/materialize/blob/master/CONTRIBUTING.md)**.

### DIFF
--- a/sass/components/_waves.scss
+++ b/sass/components/_waves.scss
@@ -12,7 +12,6 @@
 .waves-effect {
   position: relative;
   cursor: pointer;
-  display: inline-block;
   overflow: hidden;
   user-select: none;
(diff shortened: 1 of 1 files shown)


IAMtheIAM:
Nevermind, I see why it was done. Buttons default to inline block anyway.

PR #5295, opened 2017-10-23
Change hiddenDiv.height() to .outerHeight()

jekuer:
The autosize feature for textareas does not work that well if one changes the style of the textarea to make use of padding and border-box.
I would suggest to simply change .height to .outerHeight for the hiddenDiv. It does not change anything to the current behavior, but one now can simply add the style also to the hiddenDiv, which then gets applied to the autosize automatically.

Another option would be to consider the textareas other css (beside the existing checks) and check for padding as well as box-sizing, and more. Also possible, but needs way more testing. The first approach would be a easier and faster fix.

Chose your favorite solution. One of them would be awesome!!

## Types of changes

## Checklist:

### DIFF
--- a/js/forms.js
+++ b/js/forms.js
@@ -159,8 +159,8 @@
        * Resize if the new height is greater than the
        * original height of the textarea
        */
-      if ($textarea.data('original-height') <= hiddenDiv.height()) {
-        $textarea.css('height', hiddenDiv.height());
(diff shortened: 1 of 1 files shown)


Dogfalo:
This should be fixed in v1-dev as the text area now uses border-box. Please give it a try and I can reopen this if it is still not working.

PR #5315, opened 2017-10-30
Bullet lists: allow to customize style type and padding

jouvin:
## Proposed changes

Currently, list style type is set to `none` for all bullet lists and this is not necessarily the expected behaviour. This PR allows to define the list style type and left padding with 2 new variables:

* `$ul-padding-left`: defaults to 0 (previous behaviour)
* `$ul-list-style-type` : defaults to `none` (previous behaviour)

Also the modification takes core of not affecting the list characteristics for navbar and footer which remain with no padding and a style type defined to `none`. The `ul` lists in these contexts remain uncustomizable.

## Types of changes

## Checklist:

- [x] My change requires a change to the documentation (to describe the new features).


jouvin:
As said in the description, this PR itends to add a backward-compatible new feature but this is not completely the case because the definition of `ul` attributes in `_global.scss` have been moved under `main`. Other parts of Materialize require the page content (except the navbar and the footer) to be enclosed in `<main></main>` tags, e.g. sticky footer but I don't think this is an explicit requirement. The problem that led to doing this is the present of `&.not(.browser-defaults)` which cause the definition in `_global.scss` to override the one for navbar or footer. May be there is a better of doing this...

May be there is a better way to achieve this without requiring the `<main>` section.... (I don't know what `.browser-default` is doing...)

PR #5319, opened 2017-10-30
removed font-weight declaration from flow text

TheAvinashSingh:
The user should determine the font-weight in their own stylesheet, not the framework. Mainly because a user may want to use flow-text where h1-h6 tags occur, which effectively removes the bold from all h1-h6 tags. This forces setting font weight on them using !important which is not a good practice, or override it by explicitly setting

### DIFF
--- a/sass/components/_typography.scss
+++ b/sass/components/_typography.scss
@@ -45,7 +45,6 @@ small { font-size: 75%; }
 
 
 .flow-text{
-  font-weight: 300;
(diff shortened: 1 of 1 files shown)


tomscholz:
Closes: #5319

Dogfalo:
Fixed here: b34f1dff8fa6a966b7fc7a26af417cc8845d9141

PR #5347, opened 2017-11-12
Fixed inline input field in documentation

nkalupahana:
## Proposed changes

In the documentation (in Forms > Text Inputs), after clicking on the full-column "Email" field, clicking on the inline "Email" field re-opens the full-column email input field instead of the inline field, because they have the same id. I changed the inline input field's id from email to email_inline, and changed the code snippet to reflect that change.

## Types of changes

## Checklist:


Dogfalo:
Thanks!

PR #5354, opened 2017-11-14
Added missing closing backtick on data-target

jacobq:
Trivial typo correction


acburst:
This was fixed in b17de31

PR #5375, opened 2017-11-17
Allow touch events to be disabled in carousel

rvirji-va:
## Proposed changes
Sometimes, touching is a no-no. This PR adds a flag to Materialize called `enableTouch` which allows the user to optionally disable the touch event binding.

## Types of changes

## Checklist:

### DIFF
--- a/jade/page-contents/carousel_content.html
+++ b/jade/page-contents/carousel_content.html
@@ -98,6 +98,12 @@
             <td>Don't wrap around and cycle through items.</td>
           </tr>
           <tr>
+            <td>enableTouch</td>
+            <td>Boolean</td>
+            <td>true</td>
+            <td>Enable scriolling by touch.</td>
+          </tr>
+          <tr>
             <td>onCycleTo</td>
             <td>Function</td>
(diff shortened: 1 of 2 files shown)


GitFlip:
Question about this feature. Correct me if I am wrong, but it seems like this flag is being used for disabling the touch events when the carousel is initialized. 

Are there any plans to toggle the touch events on an off? The toggle feature is seen elsewhere, and I was surprised to see that this was not working for the carousel.

I'm currently getting around this by removing all of the touch events when I'm in a "read only  mode" and then reintializing the carousel when this mode is over. Is there a more appropriate way to handle this?

rvirji-va:
@GitFlip As far as I know, there are no plans to make this toggle, but I wouldn't mind revisiting this PR to implement a toggle feature. As a workaround to this, I had to do something similar to you where I manually removed the touch events after initializing the carousel.

SakuraLove:
Cloud we call the _removeEventHandlers? I tried it and it did disable the touch event and it can still be changed by calling the functions such as set(), next() and so on.

PR #5401, opened 2017-11-23
Fix for tabs breaking when there's no associated content to a tab

abradley2:
## Proposed changes

Tabs currently break if there's not a matching content div found by [this line](https://github.com/Dogfalo/materialize/compare/v1-dev...abradley2:fix/tabs-without-content?expand=1#diff-3dd5119e1daf5898ff764a06ac0aca24R164) because the check here is looking against `undefined`, whereas an empty jquery-like wrapper is returned when it can't find anything.

I propose changing the check so it no longer fails with a "cannot read property `style` of undefined" error by properly checking if the selector is empty or now.

This is useful when you want to use the tabs component but aren't relying on the tabs themselves to manage showing/hiding the content- which I'm doing in another app that wraps Materialize in Preact.

## Types of changes

## Checklist:

### DIFF
--- a/js/tabs.js
+++ b/js/tabs.js
@@ -179,7 +179,7 @@
           });
         }
       } else {
-        if (this.$content !== undefined) {
+        if (this.$content[0]) {
           this.$content[0].style.display = 'block';
           this.$content.addClass('active');
           if (typeof(this.options.onShow) === 'function') {


acburst:
Fixed in 66869208.

Thanks for catching this! I used .length instead as a personal preference and also added a test for this case.

PR #5481, opened 2017-12-17
Change of SCSS file importing

Bricklou:
This change allows fourseven:scss of Meteor.JS to import the files SCSS of this package.

Now we can do `@import '{materialize:materialize}/sass/components/_variables.scss';` in Meteor when this change will be apply.

## Types of changes

### DIFF
--- a/package.js
+++ b/package.js
@@ -61,7 +61,7 @@ Package.onUse(function (api) {
     'sass/materialize.scss'
   ];
 
-  api.addFiles(scssFiles, 'client');
+  api.addFiles(scssFiles, 'client', { isImport: true });
 
 
   api.export('Materialize', 'client');


Dogfalo:
Can you explain a bit more about what this will do? Will it break functionality for some users in the meteor package?

Bricklou:
The change allows access to SCSS files of your package. I have do this because before imports like `@import '{materialize:materialize}/sass/components/_variables.scss';` was not possible.

It will not break functionalities because it just a permissions change of access of files.

Dogfalo:
Thanks!

Bricklou:
Currently, I'm looking for a Color Picker:
Image of color picker
This image is an old screenshot.

Issue #5471, opened 2017-12-14
modal position and size


ray007:
The defaults for `M.Modal` are nice for most cases, but I'm having a really hard time with some things:

- have the modal show up in the fixed position. Especially `top` is always overridden with the opening animation. So I can get this to work by setting `startingTop` and `endingTop` to the desired value.
- have the dialog take its size from the contents. For non-IE browsers I can make this work by forcing `width: fit-content;`, but the only way I've found so far for IE is settings `display: table;`, and doing this in the `ready` callback is a bit late and looks ugly.
- setting x position is also off due to `margin: auto;` on modals.

Is this a missing feature, or do I just miss an example to copy?

Dogfalo:
See my comment on #5493

PR #5493, opened 2017-12-27
avoid wrong ending position due to animation if modal is displayed at…

ray007:
… fixed x-position

## Proposed changes

This fixes part of my problems mentioned in #5471:
A `Modal` with a fixed x-Position ends up in the wrong place after the opening animation is finished.
Avoiding an animated `scaleX` in the animation fixes this.

For this the fixed position to work correctly, there also needs to be set `margin:0;` on the modal element.

## Types of changes

## Checklist:

### DIFF
--- a/js/modal.js
+++ b/js/modal.js
@@ -58,6 +58,7 @@
       this.id = this.$el.attr('id');
       this._openingTrigger = undefined;
       this.$overlay = $('<div class="modal-overlay"></div>');
+      this.fixedX = !!(el.style.left || el.style.right);
 
       Modal._increment++;
       Modal._count++;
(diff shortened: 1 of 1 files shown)


Dogfalo:
This seems like a niche case particular to your project. I don't think its that important to include this in the framework

ray007:
Absolute positioning of a modal is a niche case?
I'd have said it's one of the fundamentals of any ui framework.

Dogfalo:
It's not a common request and the way it's implemented in the PR is very hacky as it disables animations

ray007:
No disagreements that the current implementation is hacky. Though it doesn't disable the animation, it only modifies the parameters.
Without any indication about interest in this, there was no reason to look for a better solution. Maybe even setting a better `transform-origin` would help, haven't yet tested other options since...

PR #5514, opened 2018-01-02
Pass correct element reference to Autocomplete.init

aularon:
## Proposed changes
[Chips](http://next.materializecss.com/chips.html) feature is broken in v1-dev (Throwing `TypeError:  .Autocomplete.init(...) is null @ chips.js:340:26`). This simple fix makes it work (passes the element itself to the .init call instead of the container jquery-like object.

## Types of changes

## Checklist:

### DIFF
--- a/js/chips.js
+++ b/js/chips.js
@@ -337,7 +337,7 @@
         this.$input[0].focus();
       };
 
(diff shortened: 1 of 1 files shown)


Dogfalo:
Thanks!

PR #5546, opened 2018-01-12
 Correctly set chipsInstance.autocomplete, plus docs!

aularon:
## Proposed changes
chips instance `.autocomplete` was being set to `undefined`, now the underlying autocomplete instance is correctly exposed.

## Types of changes

## Checklist:

- [x] My change requires a change to the documentation: Added it to the bottom of [Chips Properties table](http://next.materializecss.com/chips.html#properties).

### DIFF
--- a/jade/page-contents/chips_content.html
+++ b/jade/page-contents/chips_content.html
@@ -279,6 +279,11 @@
               <td>Boolean</td>
               <td>If the chips has autocomplete enabled.</td>
             </tr>
+            <tr>
+              <td>autocomplete</td>
+              <td>Object</td>
+              <td><a href="autocomplete.html">Autocomplete</a> instance, if any.</td>
+            </tr>
           </tbody>
         </table>
       </div>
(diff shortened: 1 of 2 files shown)


DanielRuf:
Can you provide a new unit test which fails with the old code and succeeds with the new code?

aularon:
Hello @DanielRuf,

I don't really know my way around writing tests for similar cases (client side javascript), but [the only change I made to the code](https://github.com/Dogfalo/materialize/pull/5546/files#diff-db5d39b9a0acd9608c98ea365623bf8dL340) is removing the `[0]` from the return value of `M.Autocomplete.init`, as the return value is the autocomplete instance, and accessing its `[0]` always returns `undefined`, while it is clearly meant that we want to assign to `.autocomplete` so it can be used later.

That line was originally written by @Dogfalo, and I updated it later with #5514. Maybe he can shed a light on this case.

Dogfalo:
We recently changed the way the init funciton works which caused this issue

PR #5551, opened 2018-01-14
Update modal.js to properly call onOpenStart

giovalerio:
## Proposed changes
the initialization of this._openingTrigger happens AFTER the onOpenStart callback; this means the callback receives "undefined" the very first time it's called and in the subsequent calls
it receives the element that triggered the previous modal open
 
by moving the initialization before the onOpenStart call, the callback receives the correct information 

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue).

## Checklist:

- [X] I have read the **[CONTRIBUTING document](https://github.com/Dogfalo/materialize/blob/master/CONTRIBUTING.md)**.

### DIFF
--- a/js/modal.js
+++ b/js/modal.js
@@ -279,6 +279,9 @@
 
       this.isOpen = true;
 
+      // Set opening trigger, undefined indicates modal was opened by javascript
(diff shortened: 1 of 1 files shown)


acburst:
Good catch!

PR #5588, opened 2018-01-23
Added border and label colors for Select

shmshd12:
## Proposed changes

When an option is selected which have a value in it, proposed js will add colors for label and input bottom border of Select. This is just for an idea and this can be improved. This is done for the uniformity of other Form elements, to match with the colors. Select menu is always reverting to grey (default) color after the user leaves ":focus" on it. This will fix the issue. 

## Types of changes

## Checklist:

### DIFF
--- a/js/select.js
+++ b/js/select.js
@@ -384,3 +384,18 @@
     M.initializeJqueryWrapper(Select, 'select', 'M_Select');
   }
 }(cash));
+
+ /**
+  * Set color for label and input bottom border on select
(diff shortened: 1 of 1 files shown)


DanielRuf:
Imo this can and should be in your own custom code.

DanielRuf:
And in general this should be solved in CSS/SCSS.

Dogfalo:
I would agree with @DanielRuf here.

PR #5593, opened 2018-01-24
blur input element on picker open to avoid onscreen keyboard

### LINKED ISSUE #5592: pickers on mobile and keyboard
Activating a date- or timepicker on mobile not only opens the picker, but also shows the phone onscreen keyboard.

## Expected Behavior

Only show the picker, no onscreen keyboard.

## Current Behavior

The picker opens and most times the onscreen keyboard is shown too. (tested with android, waiting for our iOS test env to become available)

## Possible Solution

Call `blur()` on the input tag when the picker is opened.
For datepicker this can be done from the `onOpen` callback, timepicker does not have a callback to do this in the options.
In  Chrome on android we may be able to use the native pickers.

ray007:
## Proposed changes

Blur input element to avoid onscreen keyboard opening with date- or timepicker.
Fixes #5592.

## Types of changes

## Checklist:

### DIFF
--- a/js/datepicker.js
+++ b/js/datepicker.js
@@ -834,6 +834,7 @@
         return;
       }
 
+      this.$el[0].blur(); // avoid onscreen keyboard on mobile
       this.isOpen = true;
       if (typeof this.options.onOpen === 'function') {
         this.options.onOpen.call(this);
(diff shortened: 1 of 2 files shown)


anoadragon453:
This would be very nice to have.

PR #5614, opened 2018-01-29
Increment year

andrey-helldar:
Update year in the footer


DanielRuf:
In my opinion we should use js to keep it uptodate, see http://updateyourfooter.com

Dogfalo:
Merged #5640

PR #5684, opened 2018-02-21
Update Scrollspy documentation example for native javascript

mariogmz:
The documentation shows both native javascript and jQuery way to initialize the Scrollspy plugin, the native javascript example uses document.querySelector, which only returns the first element found matching the selector, resulting in only initializing one element, this needs update  so it uses document.querySelectorAll, which returns the whole NodeList before calling the M.ScrollSpy.init method.

Solves https://github.com/Dogfalo/materialize/issues/5569

## Proposed changes
Just a simple update to the `jade` template for that particular example

image

## Types of changes
- Jade template updated for the documentation.


Dogfalo:
Yes, good point. We'll be changing it in the future for all the plugin docs

PR #5725, opened 2018-03-07
Ref. 4736 Chips autocomplete should have option to prevent user own input 

dorphalsig:
Chips autocomplete should have option to prevent user own input 
----
- added boolean option autoCompleteOnly to chips defaults options
- when pressing enter to add a chip, the chip is only added if M.Chips.options.autocompleteOnly==false (or if hasAutcomplete == false)

### DIFF
--- a/js/chips.js
+++ b/js/chips.js
@@ -6,6 +6,7 @@
     placeholder: '',
     secondaryPlaceholder: '',
     autocompleteOptions: {},
+    autocompleteOnly: false,
     limit: Infinity,
     onChipAdd: null,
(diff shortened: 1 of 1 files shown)


dorphalsig:
@jagansindia  would be even more perfect if it got merged xD

ryanrapini:
Any chance this is getting merged? Would love this option

alphamarket:
Why on earth this hasn't got merge since 2018?

amardeepranu:
Any chance of merging?

Issue #4086, opened 2017-01-11
Improve autocomplete


djensen47:
Autocomplete is problematic for a number of reasons:
- can't limit results
- can't use IDs instead of strings
- difficult to internationalize values (see previous about using IDs)
- drop down pushes other elements down on the page

Here's a library that has solved many of these issues:
https://github.com/icefox0801/materialize-autocomplete

I would like to suggest either integrating this library or all of it's features.

acburst:
Added a first pass at chips autocomplete integration in 83e6ba0.

Would appreciate some testing. You can check it out on the latest master version under the chips docs.

gewisser:
Great! 
Before the new year, he made such a revision, did not manage to propose the inclusion of improvements.

djensen47:
Chips are cool but I was actually more concerned about the other items. I've updated the issue above to specify the features that I'm looking for in a robust autocomplete solution.

djensen47:
Regarding using IDs. This helps immensely with a few things including i18n.

For instance:
`[{value: 'apple', id: '4rew65s1', img: '…'}, {value: 'banana', id: '7r98vv8s', img: '…'}]`
and 
`[{value: 'manzana', id: '4rew65s1', img: '…'}, {value: 'plátano', id: '7r98vv8s', img: '…'}]`
would be the same with the labels localized

With the current autocomplete implementation, I would have to have a lookup table for every single language that I support to get the proper id, which is not scalable.

hfrid:
+1 really need IDs... now I need to first match a string then make another db-call to fetch the ID :(

djensen47:
I ended up building my own autocomplete from scratch to support IDs. It's built using Aurelia, however, so it might not be helpful to you unless you're using Aurelia.

bkrajendra:
Autocomplete Pushes other content down looks like non standard behavior.
I fixed it by adding CSS:
```
 .autocomplete-content {
    position: absolute;
} 
```

bugsoftware:
Hi,

I decided to contribute a bit with my feedback. 2 things are critical to autocomplete that are missing, the first one is the ability to pull data from outside source (JSON/XML) and as djensen47 pointed out, the ability to combine ID's is also critical / [{value: 'apple', id: '4rew65s1', img: '…'} ]

drusepth:
[...]

PR #5847, opened 2018-04-22
Initial stab at future Ajax support

melvyn-sopacua:
## Proposed changes
Add a callback hook just before data is consulted, leveraging
the logic that is already there (irrelevant keys, minlength, etc).

This needs more work to deal with requests that typically take longer
than the time it takes for a user to type multiple keys, but the API
for the callback is very unlikely to change as all that logic can
be encapsulated in the callback itself.

I'm not well-versed in JS unit tests, but will add them if the proposed
implementation is accepted.

Contributes to #4086 .

A (working) AJAX implementation is in [this branch](https://github.com/melvyn-sopacua/materialize/blob/poc-komoot-photon/js/init.js).

## Types of changes

## Checklist:

### DIFF
--- a/jade/page-contents/autocomplete_content.html
+++ b/jade/page-contents/autocomplete_content.html
@@ -89,6 +89,12 @@
               <td>Callback for when autocompleted.</td>
             </tr>
             <tr>
+              <td>onValueChange</td>
+              <td>Function</td>
+              <td>null</td>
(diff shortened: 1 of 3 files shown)


jpic:
The API looks good but I have a question: should we trigger this instantly or should we defer like 100ms to give the user a chance to type another character before firing the event ?

Issue #5912, opened 2018-05-11
rc1 causes cash-dom crash with rollup.js


ginlime:
## Expected Behavior
Works well with rollup.js.

## Current Behavior
Console of Chrome shows:
```
Uncaught ReferenceError: $ is not defined
```

Console of Firefox also shows:
```
ReferenceError: $ is not defined
```

## Possible Solution
Keeping lead UDM statement of cash-dom solves this problem:
``` javascript
;(function (root, factory) {
  if (typeof define === "function" && define.amd) {
    define(factory);
  } else if (typeof exports !== "undefined") {
    module.exports = factory();
  } else {
    root.cash = root.$ = factory();
  }
})(this, function () {
```

## Your Environment
* Version used: rc1
* Browser Name and version: Chrome 66.0.3359.139, Firefox 61.0b3
* Operating System and version (desktop or mobile): Windows 7 SP1

PR #5915, opened 2018-05-14
Reverted UMD as original according to #5912

ginlime:
See #5912


Dogfalo:
We'll probably bind it to M as this has caused problems for other loaders in the past

ginlime:
@Dogfalo 
Oh, I see.
I'm looking forward to see implemented version.

PR #5920, opened 2018-05-15
Workaround race condition in some IE11 versions

jpic:
On IE 11.0.9600.18977 there is a race condition that caused both
previous and current value to be displayed as selected in the selected.

_setValueToInput() indeed gets value from each el that is selected. And
with IE 11 it found two selected options at the time of execution.

This patch ensures that silly browsers who can't maintain a consistent
state for a select field will not fall in the trap.

Please release friend :)


jpic:
Resolving some conflicts ...

PR #5921, opened 2018-05-15
Workaround race condition in some IE11 versions

jpic:
## Proposed changes

On IE 11.0.9600.18977 there is a race condition that caused both
previous and current value to be displayed as selected in a non-multiple
select.

_setValueToInput() indeed gets value from each el that is selected. And
with IE 11 it found two selected options at the time of execution.

This patch ensures that silly browsers who can't maintain a consistent
state for a select field will not fall in the trap.

## Types of changes

## Checklist

- [ ] I have read the **[CONTRIBUTING document](https://github.com/Dogfalo/materialize/blob/master/CONTRIBUTING.md)**.

### DIFF
--- a/js/select.js
+++ b/js/select.js
@@ -358,6 +358,10 @@
         }
       }
 
+      if (!this.isMultiple) {
+        values = values.slice(values.length - 1)
+      }
+
       this.input.value = values.join(', ');
     }
 


angelo510:
The idea looks good.
But the code snap would not work on ie11, even on chrome.

I just corrected it like this. 
if (!this.isMultiple) {
  values = values.slice(values.length - 1)
}

According to my code, not only the values length would be forced to 1 if not multiple, but also the values always would be the last item of array.
for example, 
---------, Soumise  -> Soumise
Rejetee -> Rejetee

I just tested on ie11 and it works good.

jpic:
Hello, can we release this please ? Thank you very much.

jpic:
Hello ! We are still affected, please feedback ;)

Thanks ;)

jpic:
2018-08-22-145408_1198x128_scrot

jpic:
Yeah so if anyone has to deal with a government that is stuck in that shitty IE version, here's our fork: https://www.npmjs.com/package/mrsmaterialize

We've been sponsoring MaterializeCSS all year and yet in 5 months no feedback from such a trivial PR FYI.

jpic:
Hello, this is not in 1.0, which means that MaterializeCSS 1.0, is still not compatible with the browser used inside the government, our customer.

Please release, I beg you.

jpic:
I don't think they will merge this PR.

I hope this government updates their browser one day :joy:

Issue #5959, opened 2018-06-01
Collapsible closes on pressing enter even after being destroyed (desktop/mobile).


CiriousJoker:
## Expected behavior:
- The user shouldn't be able to interact with the collapsible in any way after it's been destroyed

## Actual behavior:
- Pressing enter closes it

## Steps to Reproduce (for bugs)
```
        const elemCollapsible = document.getElementById("some_collapsible");
        const collapsible = new M.Collapsible(elemCollapsible, {});
        collapsible.destroy();
```

## Context
- I have a collapsible search bar (search results are in the body) and whenever the user submits his search query via Enter (desktop/mobile doesn't matter), the collapsible is closed.

## Your Environment
* Version used: 1.0rc1
* Browser Name and version: Chrome 67
* Operating System and version (desktop or mobile): Windows 10 & Android 8.1

PR #5960, opened 2018-06-01
Remove all the event handlers after destroying it.

CiriousJoker:
Potentially fixes [this issue](https://github.com/Dogfalo/materialize/issues/5959).

This is in no way tested and I cannot guarantee that this will work and have no side effects.

This is just meant to locate the problem that caused the bug.


acburst:
looks good, thanks!

PR #6044, opened 2018-07-24
Refine the handling of `container` attributes of some components.

drothmaler:
## Proposed changes
There are multiple components that allow the user to specify a HTML element, that should be uses as a container for the rendering of their dynamically added child HTML components.
Those components are: Date-Picker, Time-Picker and Dropdown.

Unfortunately this `container` attributes are not handled/documented uniformly, so I updated the documentation to improve that. 
I also use those components in a project that doesn't use the full materialize-css style. So I want to be able to contain materialize stuff inside a dedicated container. Unfortunately the Date-Picker, didn't pass it's container configuration on to its child dropdowns - so they where rendered in the document body instead - which is very unfortunate, as it broke the layout of other things in my app. So I fixed the month/year dropdowns to use the same container as the Date-Picker (this might be a breaking change in some rare cases).

## Types of changes

- [x] Documentation fix.

## Checklist:

### DIFF
--- a/jade/page-contents/dropdown_content.html
+++ b/jade/page-contents/dropdown_content.html
@@ -95,9 +95,9 @@
             </tr>
             <tr>
               <td>container</td>
-              <td>Element</td>
+              <td>Element || String</td>
               <td>null</td>
-              <td>Provide an element that will be the bounding container of the dropdown.</td>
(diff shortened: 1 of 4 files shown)


drothmaler:
As this PR seems to be ignored, I decided to split it to submit the documentation and behaviour changes separately. Maybe this we we get at least the doc fix merged quickly.

PR #6056, opened 2018-08-03
add vanilla JS init example for CharacterCounter

blunket:
## Proposed changes

This PR accomplishes 3 things:
1. add a vanilla javascript initialization example for the CharacterCounter plugin.
2. normalize the "initialization" examples everywhere.
3. clarify any help text where necessary, or remove any help text that is redundant.

Open to discuss.

### DIFF
--- a/jade/page-contents/navbar_content.html
+++ b/jade/page-contents/navbar_content.html
@@ -641,9 +641,6 @@ $(".dropdown-trigger").dropdown();
 
         <br>
         <h5>Initialization</h5>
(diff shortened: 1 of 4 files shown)


jade/page-contents/navbar_content.html
```diff
@@ -641,9 +641,6 @@ <h3 class="header">Mobile Collapse Button</h3>
 
         <br>
         <h5>Initialization</h5>
-        <p>After including the sidenav-trigger line into your navbar, all you have to do now initialize the plugin. This example
```
blunket: I do not believe this text is necessary. If it is, then we ought to include similar text in **all** of the examples.

PR #6068, opened 2018-08-13
Remove clickedTimeout var

Wifsimster:
## Proposed changes
`clickedTimeout` is declared but var is never read.
In that case, remove `clickedTimeout` var.


acburst:
No longer applies after 1.0.0 changes

PR #6090, opened 2018-08-29
Fix logic error that prevents tabbing through autocomplete

yavuztor:
## Proposed changes
When I realized that I could not tab through autocomplete component, I started looking into the code. I was able to find where the tabbing went awry. When user tabs into the autocomplete input element, we don't want to open the dropdown, yet. The code was testing for that condition, but it needed a slight change. It should open if the tab is not pressed.

https://github.com/Dogfalo/materialize/blob/ca9beaf1591f9e5fa4536a4f6d8dabd6136406fb/js/autocomplete.js#L204 

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue).

## Checklist:

- [x ] I have read the **[CONTRIBUTING document](https://github.com/Dogfalo/materialize/blob/master/CONTRIBUTING.md)**.
- [ x] All new and existing tests passed.

### DIFF
--- a/js/autocomplete.js
+++ b/js/autocomplete.js
@@ -201,7 +201,7 @@
 
       // Check if the input isn't empty
       // Check if focus triggered by tab
-      if (this.oldVal !== val && (M.tabPressed || e.type !== 'focus')) {
+      if (this.oldVal !== val && (!M.tabPressed || e.type !== 'focus')) {
         this.open();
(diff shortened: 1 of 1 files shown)


yavuztor:
I don't thinks this will get any attention. It might be something that's fixed by other changes.

PR #6091, opened 2018-08-29
useless float: left

DraftProducts:
float: left;
is ignored on a 
display: inline-block;

## Proposed changes
float: left removed because we have a display: inline-block and float is ignored with this display

## Checklist:


acburst:
Not sure where you heard that, but floats definitely work on display: inline-block elements.

PR #6106, opened 2018-09-10
Use custom format function on datepicker init if defined

Mzeus:
## Proposed changes
When you define a custom format function for the datepicker, its not used on the init. The datepicker create the date object from the input value when no default date is defined, and uses the default format function from javascript that only accepts dates on the format "YYYY-MM-DD".

With these proposed change, on the init the datepicker will check if there is a custom format function defined and will use it insted of the default javascript one to load the date from the input value when no default date is defined.

## Types of changes

## Checklist:

- [x ] I have read the **[CONTRIBUTING document](https://github.com/Dogfalo/materialize/blob/master/CONTRIBUTING.md)**.

### DIFF
--- a/js/datepicker.js
+++ b/js/datepicker.js
@@ -141,7 +141,12 @@
       this._setupEventHandlers();
 
       if (!this.options.defaultDate) {
-        this.options.defaultDate = new Date(Date.parse(this.el.value));
+        // Check if there is a custom format function and use it to load the date from input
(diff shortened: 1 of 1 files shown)


eduardoalcantara:
I am still wainting for this to get done.
Maybe we should pass this project to Google or merge ir with Bootstrap since most people have to choose between them two.

PR #6211, opened 2018-11-16
Added extra large css helpers

ghost:
I've just started using Github! I dont' know if I'm right place to make a suggest! Sorry if I'm not!

I've just seen these two css classes missing in the documentation on Helpers page!

.hide-on-extra-large-only
.show-on-extra-large


Chiagozielam:
good stuff

al1Null:
yo dog, that doesn't make sense. It's still not updated in the docs

PR #6315, opened 2019-03-27
Fix collapsible-body height 0

KillTrot:
## Proposed changes

In specific cases the height of a `active` collapsible is not correct after `_animateIn`.
If the collapsible-body height is equal to 0 it uses the outerHeight.
With these changes it should be correct.

## Codepen:

Version 1.0.0: https://codepen.io/anon/pen/NJVoRo
Fixed: https://codepen.io/KillTrot/pen/PLvVJx (fixed with onOpenEnd event and jQuery, in the pull request it's fixed in the code directly)

## Types of changes

## Checklist:

### DIFF
--- a/js/collapsible.js
+++ b/js/collapsible.js
@@ -174,7 +174,7 @@
               overflow: '',
               paddingTop: '',
               paddingBottom: '',
-              height: ''
+              height: $body[0].height == 0 ? finalHeight : ''
             });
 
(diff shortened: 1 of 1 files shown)


acburst:
This bug is caused by a mistake in your grid use. All .col elements must have a direct .row parent. Otherwise it can cause float issues like this.

PR #6357, opened 2019-05-06
Avoid setAttribute for CSP compliance

evs-ch:
## Proposed changes

Under some circumstances (we experienced this issue with a recent rails and appropriately strict CSP policies, see mkhairi/materialize-sass#186), materialize causes CSP violations.

Using `setAttribute('style', style)` causes issues with the style policy. `element.style = ...` is CSP compliant and could be used instead. jQuery offers `.css()`, which uses the aforementioned `.style`.
This PR changes the problematic `setAttribute` calls to using a `setStyle()` function which checks if jQuery is loaded and uses `.css()` if possible or falls back to `setAttribute`.
It might be possible to enhance the function to fall back to looping through the styles and directly applying them with `.style = `, but I didn't look into that yet.
I "played it safe" with regards to the jQuery check, but maybe `.css()` is already implemented by whichever way materialize chose to replace/deprecate jQuery, so it might not even be necessary.

/

## Types of changes

## Checklist:

### DIFF
--- a/js/waves.js
+++ b/js/waves.js
@@ -51,6 +51,15 @@
         return style;
     }
 
+    function setStyle(obj, style) {
+      if (M.jQueryLoaded) {
+        $( obj ).css(style);
+      }
+      else {
+        obj.setAttribute('style', convertStyle(style));
+      }
+    }
+
     var Effect = {
 
         // Effect delay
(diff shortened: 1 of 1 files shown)


acburst:
We no longer use jQuery as a dependency

evs-ch:
That's good to know, but what about the actual intention of my PR? Should I rewrite it without using jQuery, or is the original problem (CSP violations) a thing that you are not interested in getting fixed? And if it should be rewritten: is there something else implementing `.css()` used in materialize? Otherwise, I'd need to rewrite the relevant method with plain JS, which should be possible but might be unneccessary.

PR #6359, opened 2019-05-09
Fix #6306 Tooltip positions shifted in horizontally scrolling table.

### LINKED ISSUE #6306: Tooltip positions shifted in horizontally scrolling table
I have a very wide table that uses tooltips in many cells. The table is wider than can be displayed in a browser window so only about half the columns are visible at a time. When the page is scrolled fully to the left, all visible tooltips function properly. When scrolled to the right to reach the other columns, those tooltips will always be displayed at the far right of the screen no matter where the tooltip trigger is.

## Expected Behavior
When scrolling horizontally, tooltips should continue to function properly being displayed next to their trigger.

## Current Behavior
Tooltips that are initially outside the viewport when a page is loaded will display at the far right side of the viewport regardless of the position of its trigger.

## Possible Solution
Issue #5790 was similar to this, but dealing with vertical placement of the tooltips instead of horizontal. A fix was done here: https://github.com/Dogfalo/materialize/commit/c00c954f04ad58bb5e274e9d410811c38492c808

A similar fix can be applied within the checkWithinContainer function in materialize-v1.0.0.js at line 1264. Just after line 1274, I added:

`var containerRight = container === document.body ? Math.max(containerRect.right, window.innerWidth) : containerRect.right;`

Then I changed the current line 1287 to this:

`if (scrolledX + bounding.width > containerRight - offset || scrolledX + bounding.width > window.innerWidth - offset) {`

With that, everything works as expected.

## Steps to Reproduce (for bugs)
https://codepen.io/akemd/pen/GeOMvj
1. Size window so only half the table columns are seen. Tooltips work as expected for the visible triggers when scrolled to the left.
2. Scroll to the right and hover over those tooltip triggers. The tooltip is displayed to the far right.

## Your Environment

* Version used: 1.0.0
* Browser Name and version: Firefox 65.0.2, Chrome 72.0.3626.121
* Operating System and version (desktop or mobile): Windows 10 desktop

valeriacobian:
## Proposed changes
Issue #6306 
This fix resolves a bug with tooltip position horizontally. 
**The current behavior of tooltips:** When a page is loaded will display at the far right side of the viewport regardless of the position of its trigger and tooltips are initially outside of the viewport. I think is important to accept this fix because this bug affects how is the data display in the view. 

Bug behavior: 
PRMAterialize

Here's a codepen example: https://codepen.io/akemd/pen/GeOMvj

The resolved bug is interacting in this way:
PRMAT

## Types of changes

## Checklist:

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -217,6 +217,10 @@ M.checkWithinContainer = function(container, bounding, offset) {
     container === document.body
       ? Math.max(containerRect.bottom, window.innerHeight)
       : containerRect.bottom;
+  let containerRight = 
+    container === document.body 
+      ? Math.max(containerRect.right, window.innerWidth) 
+      : containerRect.right;
 
   let scrollLeft = container.scrollLeft;
   let scrollTop = container.scrollTop;
@@ -230,7 +234,7 @@ M.checkWithinContainer = function(container, bounding, offset) {
   }
 
   if (
-    scrolledX + bounding.width > containerRect.right - offset ||
+    scrolledX + bounding.width > containerRight - offset ||
     scrolledX + bounding.width > window.innerWidth - offset
   ) {
     edges.right = true;


SubJunk:
Just wanted to say thanks for this since I've experienced that bug before and it caused me to not use this for tooltips

acburst:
Merged into rollup in 1498e0c5, will be merged into next release

PR #6381, opened 2019-06-03
Update CSS to prevent dropdowns from spilling over modal edges

Fauxil-Fox:
## Proposed changes
In applications I am working on, I have found the need to place dropdowns into modals quite common.  However, large dropdowns would always spill over and force the modal to have overflow.  This fixes this by forcing dropdowns to be no larger than 60vh (the modal itself will always be no longer than 70vh).

It may be more appropriate to add this into the JS itself or another CSS file, I just thought I'd share my fix.

You can see the issue before and after my patch:
https://codepen.io/anon/pen/EzrOjE

## Types of changes

## Checklist:

### DIFF
--- a/sass/components/_dropdown.scss
+++ b/sass/components/_dropdown.scss
@@ -83,3 +83,8 @@ body.keyboard-focused {
 .dropdown-trigger {
   cursor: pointer;
 }
+
+// Do not spill over the edge of a modal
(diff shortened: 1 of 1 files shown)


acburst:
Thanks for bringing this issue up, however this seems like more of a bandaid fix. 

I've fixed this issue in b44047c1. I now handle containers of different position and overflow values.

Issue #6403, opened 2019-06-27
[Feature Request] Allow Duplicate chips to be added


JanieHuang:
Hi, 

It would be great if we can decide whether or not to allow duplicate chip entries. What I'm trying to do is have the chips form a sentence, but whenever I type a word, say "and", but it already exists, it just removes itself. 

My suggestion would be to have a boolean check such as "allowDuplicate:" and default set to false so as not to interfere with anything currently, and then have the condition checked in the validation. 

In the validation, have another control where "allowDuplicate" is checked. If set to false, do not allow duplicates and continue with the current behaviour. If set to true, allow duplicates and add the chip.

I do see there are 2 PR's that are still open, and subsequently have not passed the checks,  #6292 which is5 months old (and failed checks) and #3752 which is 2 years old. 

So if no one minds, maybe I'll create a new PR?

PR #6404, opened 2019-06-28
Added function to allow duplicate chips to be added

JanieHuang:
## Proposed changes

Current functionality: When a duplicate chip is added, it immediately disappears as it is meant to check to see if the chip already exists. If it exists, it is not added. 

However, there are some situations where adding duplicate chips would actually be useful such as if you wish to do sentence forming in conjunction with the Auto complete on chips. 

I propose to add a boolean option to allow duplicates. The default is set to false, meaning no duplicates are allowed so this change will not affect any current users of the library. 

Only users that wish to use this will be able to explicitly state they wish to have duplicates.

The check is done in the _isValid(chip) function in `chip.js`. 

In the code, it can be specified like so: 
```
var instances = M.Chips.init(elems, {
  allowDuplicates: true
});
```

 ## Types of changes
 
 ## Checklist:
 
Please let me know if there's anything I would need to do in order for this to be accepted.  
 
 #6403 

EDIT: I've updated the documentation

### DIFF
--- a/jade/page-contents/chips_content.html
+++ b/jade/page-contents/chips_content.html
@@ -176,6 +176,12 @@
               <td>Callback for chip delete.</td>
             </tr>
             <tr>
+              <td>allowDuplicates</td>
+              <td>Boolean</td>
+              <td>false</td>
+              <td>Allow duplicate chips to be added.</td>
+            </tr>
+            <tr>
           </tbody>
         </table>
       </div>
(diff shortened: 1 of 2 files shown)


JanieHuang:
Build failed due to this `Warning: temp/js/materialize_concat.js: Unexpected token (7115:13)`. 

EDIT: Fixed it. Was due to extra bracket when committing code.

NoahvdAa (approved):

PR #6456, opened 2019-10-04
Footer Padding

ibadeeCodes:
## Proposed changes
Added 20px left-right-padding to footer-copyright, so that whenever someone opens the footer section in the documentation, the copyright text will have some gap between the text and its borders.
It looks in a manner for the documentation readers.

**Before:**
footer-before

**After:**
footer-after

## Types of changes

## Checklist:

### DIFF
--- a/jade/page-contents/footer_content.html
+++ b/jade/page-contents/footer_content.html
@@ -25,10 +25,10 @@
                 </ul>
               </div>
             </div>
-          </div>
-          <div class="footer-copyright">
-            © 2014 Copyright Text
-            <a class="grey-text text-lighten-4 right" href="#!">More Links</a>
+            <div class="footer-copyright">
+              © 2014 Copyright Text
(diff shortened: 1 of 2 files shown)


pwcreative:
Surely the issue here is a missing container tag on the example? The footer markup includes a container inside the footer-copyright and this should provide the spacing rather than some arbitrary padding added later? Or is that not how the docs display?

ibadeeCodes:
The footer contains a container div but the footer-copyright doesnt..
Thats why I added padding to footer copyright..
Otherwise I can also move the footer-copyright section inside the
container...
Any of these two ways are suitable..

Thanks for looking into it.

pwcreative:
I think adding the container into the markup seeing as this is how the element is structured.

ibadeeCodes:
I have moved the footer-copyright inside the container to avoid additional padding..

PR #6472, opened 2019-11-07
Update Ignore collapse action with class, Validate Textfields on updateTextFields

JackGlobetrotter:
## Proposed changes
Added a class (collapsible-ignore) check to the onClick handler for the Collapsible Header. This class gives the possibility to add input elements in the header and to click them without triggering the collapse action.
Added an optional argument to the UpdateInputFields method to validate the Fields once updated. Default is false. 

## Types of changes
-Added class check for collapsible-ignore to prevent collapse action from collapsible header
-Added optional argument to UpdateInputFields to validate Fields

### DIFF
--- a/js/collapsible.js
+++ b/js/collapsible.js
@@ -109,7 +109,7 @@
      */
     _handleCollapsibleClick(e) {
       let $header = $(e.target).closest('.collapsible-header');
-      if (e.target && $header.length) {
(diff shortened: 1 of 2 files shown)


JackGlobetrotter:
Update Ignore collapse action with class, Validate Textfields on updateTextFields

PR #6482, opened 2019-11-26
datepicker font size now displays month correctly

gilbertoayala12:
## Proposed changes
Added changes to datepicker specifically when selecting a month because the size of the text was to big and now it shows with a correct size

## Pre-fix image
Screen Shot 2019-11-26 at 12 25 07 PM

## Fixed 
Screen Shot 2019-11-26 at 12 26 37 PM

## Types of changes

## Checklist:

### DIFF
--- a/sass/components/_datepicker.scss
+++ b/sass/components/_datepicker.scss
@@ -29,6 +29,7 @@
       border-bottom: none;
       text-align: center;
       margin: 0;
+      font-size: 1.0rem;
     }
 
     .caret {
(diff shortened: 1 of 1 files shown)


DanielRuf:
Wouldn't this fail with a longer string too?

gilbertoayala12:
> Wouldn't this fail with a longer string too?

Yes. I tested it with `september` since it's the longest string in the calendar and it works correctly.

DanielRuf:
And other languages? ;-)

gilbertoayala12:
You're right and I added a text-overflow property just in case anything like that happens 😅

PR #6521, opened 2020-02-23
datepicker: add yearRangeReverse property sort year in reverse order

k2s:
This add yearRangeReverse property to date picker to sort year range in reverse order

## Types of changes

## Checklist:

### DIFF
--- a/js/datepicker.js
+++ b/js/datepicker.js
@@ -513,9 +513,7 @@
       }
       return (
         `<td data-day="${opts.day}" class="${arr.join(' ')}" aria-selected="${ariaSelected}">` +
-        `<button class="datepicker-day-button" type="button" data-year="${opts.year}" data-month="${
-          opts.month
-        }" data-day="${opts.day}">${opts.day}</button>` +
+        `<button class="datepicker-day-button" type="button" data-year="${opts.year}" data-month="${opts.month}" data-day="${opts.day}">${opts.day}</button>` +
         '</td>'
       );
     }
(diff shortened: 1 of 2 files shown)


js/datepicker.js
```diff
@@ -604,11 +602,15 @@
...
+      let years = [];
       for (arr = []; i < j && i <= opts.maxYear; i++) {
         if (i >= opts.minYear) {
-          arr.push(`<option value="${i}" ${i === year ? 'selected="selected"' : ''}>${i}</option>`);
+          years.push(
+            `<option value="${i}" ${i === year ? 'selected="selected"' : ''}>${i}</option>`
+          );
         }
       }
+      arr = arr.concat(opts.yearRangeReverse ? years.reverse() : years);
```
DanielRuf: Not sure why arr.concat is used here. You can probably leave the old code untouched and use one line with this:

```
if(opts.yearRangeReverse) {
  arr.reverse();
}
```

Which will reverse the array elements in place.

DanielRuf (changes requested):

js/datepicker.js
DanielRuf: You might also want to add it as default at the top.

k2s: @DanielRuf I fixed the overhead.
Pls. what do you mean by `add it as default at the top` ?

DanielRuf: See https://github.com/Dogfalo/materialize/blob/v1-dev/js/datepicker.js#L4, the defaults are readable and visible when you output a component instance on the console.

DanielRuf (approved):

DanielRuf:
Merged at https://github.com/materializecss/materialize/commit/5a4c33c0e8fa001e78522e88cc39c3b9dfe293a6 (includes the original commits at https://github.com/materializecss/materialize/commits/v1-dev).

PR #6523, opened 2020-02-25
Update .row

AbigaelleFerrier:
The behavior of certain elements is not yet perfect, the .offset classes must change to !important and other bug we were noticed during my tests. I will do a push request when all will be corrected to work at 100% 

Update for correctly align col class if they are < at 12

## Types of changes

## Checklist:

https://codepen.io/TairaK2/full/wvagVdj


roiLeo (changes requested):
* flex-wrap: Partial support on ie11

PR #6567, opened 2020-06-05
Update input width in _chips.scss sass component

Git-Harshit:
This is a minor but suitable change which updates the width of input field (for adding new chips) from fixed 120px to fit-content, with same !important weight. Also, the min-width and max-width properties are set to limit its size between 100px and 200px. With this change, the input placeholder (if limited to under 200px) will not get cropped, and it will also have a better fit into the input field, thereby offering it more flexibility to adapt the content.

## Proposed changes

Basically, this change fixes the issue that I had encountered recently with the placeholder or text of the input before pressing enter to convert it to a chip. Please have a look at how this small change affects the chips input (Screenshot from Google Chrome):

Capture-ChangePropose

## Types of changes

- [ x ] Bug fix (non-breaking change which fixes an issue).

## Checklist:

- [ x ] I have read the **[CONTRIBUTING document](https://github.com/Dogfalo/materialize/blob/master/CONTRIBUTING.md)**.

### DIFF
--- a/sass/components/_chips.scss
+++ b/sass/components/_chips.scss
@@ -62,6 +62,7 @@
     outline: 0;
     margin: 0;
     padding: 0 !important;
+    width: 120px !important;
     width: fit-content !important;
     min-width: 100px;
     max-width: 200px;
--- a/sass/components/_chips.scss
+++ b/sass/components/_chips.scss
@@ -62,7 +62,9 @@
     outline: 0;
     margin: 0;
     padding: 0 !important;
-    width: 120px !important;
+    width: fit-content !important;
+    min-width: 100px;
+    max-width: 200px;
   }
 
   .input:focus {


DanielRuf (changes requested):
Attention, this will not work on IE11. We should keep the old rule and add the new rule after it. So browsers who support it will use the second width rule.

https://caniuse.com/#feat=mdn-css_properties_width_fit-content

Git-Harshit:
Yes @DanielRuf sir! Thank you for mentioning! It did actually slip out of my mind when I had made this pull request. I think even the default 'auto' value for width property might not work in the desired way for IE.
Also, can you please tell if the _!important_ weight is really necessary to be used here, as I don't think it is required for the width changes to take effect? 

Now, I guess it's time to update this PR.

DanielRuf:
Regarding the new or old width setting? !important is bad in general.

Git-Harshit:
> Regarding the new or old width setting? !important is bad in general.

Actually, it was about the old width setting, as there was *!important* used with `width:120px` CSS. Completely agreed, it isn't a good practice to use !important weight in general. 

Still, as it was used in the previous code, I didn't remove it for now. Please let me know if in case it has to be discarded.

DanielRuf (approved):
LGTM

Issue #1683, opened 2015-07-06
Dropdown - click stopPropagation


otroboe:
Hey !

Do you think it would be useful to add an option to stop event click propagation when opening the dropdown ?

``` javascript
$('.dropdown-button').dropdown({
    stopPropagation: true // default: false
});
```

In my case, I use the collection component and a dropdown menu in each row. But I want a custom behavior when click on a collection-item. But this custom bind is triggered when click on the dropdown-button :)

For now I check the clicked target DOM element, but it's not very clean !

Thanks :)

rotoxl:
+1

vinikento:
+1

fega:
implemented

BrendanBehr:
It looks like this was undone with 3c76a65589ef536c0068c20e7644056b57de9a4f.  Before submitting a pull request, I just want to check to see if this was intended or an accidental removal.  I currently have a workaround implemented but I would like this feature and I have a potential fix for this.

PR #6613, opened 2020-11-02
Add stopPropagation to the Dropdown class

BrendanBehr:
This commit re-adds the stopPropagation property to a dropdown for the dropdown trigger

## Proposed changes

Adding the stopPropagation property allows an html element that contains a dropdown trigger to have a click event that does not get triggered when selecting the dropdown trigger.

In [this](https://codepen.io/brendanbehr/pen/xxOjvWW) example I have two chips with an ellipsis icon as the dropdown trigger, The left chip fires the chip event when you click on the chip and anything related to the dropdown trigger.  The right chip contains the click event on the text which is not ideal because there is un-clickable space on the chip and if I want to add an icon to the left of the text I would need to wrap the icon and text in a span that has a click event.  

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality).

## Checklist:

- [ x ] I have read the **[CONTRIBUTING document](https://github.com/Dogfalo/materialize/blob/master/CONTRIBUTING.md)**.
- [ x ] My change requires a change to the documentation.
- [ x ] I have updated the documentation accordingly.

### DIFF
--- a/jade/page-contents/dropdown_content.html
+++ b/jade/page-contents/dropdown_content.html
@@ -153,6 +153,12 @@
               <td>null</td>
               <td>Function called when dropdown finishes exiting.</td>
             </tr>
+            <tr>
+              <td>stopPropagation</td>
+              <td>Boolean</td>
+              <td>false</td>
+              <td>Stop the propagation of a click event for the dropdown trigger</td>
+            </tr>
           </tbody>
         </table>
       </div>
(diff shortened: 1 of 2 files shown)


BrendanBehr:
Dropdown - click stopPropagation #1683

This is the closest issue I could find.  I thought I saw comment related to this somewhere but Im having a hard time finding it,

mujeeb100 (approved):

